PR #3, opened 2023-04-06
WIP: Add thoughts on initial API endpoints

dwhenry:
This needs to be agreed by dev and TA for the initial naming and approach to the API endpoints before work is started.

Once agreed the actual API documentation will be done using rSwagger to generate it.


mhunt3142 (approved):

dwhenry:
outdated

PR #7, opened 2023-04-13
Laa claim non standard magistrate fee backend swagger

mhunt3142:
## Description of change
Adder swag and rspec tests for swagger api

## Notes for reviewer
first attempt


dwhenry:
@mhunt3142 so we now have a linter than run on each PR and need to pass before the code can be merged. The linter has an autofix feature so just run before you push:
```
rubocop -A
```

Some issues can't be autofixed, so if you have any trouble with them just give me a shout. :)

PR #12, opened 2023-04-24
ClaimReason WIP

mhunt3142:

### DIFF
--- a/.idea/workspace.xml
+++ b/.idea/workspace.xml
@@ -1,0 +1,328 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="AutoImportSettings">
+    <option name="autoReloadType" value="SELECTIVE" />
+  </component>
+  <component name="ChangeListManager">
+    <list default="true" id="4031468a-023f-4ab0-a007-a6772b0ae26a" name="Changes" comment="">
+      <change beforePath="$PROJECT_DIR$/app/services/decisions/simple_decision_tree.rb" beforeDir="false" afterPath="$PROJECT_DIR$/app/services/decisions/simple_decision_tree.rb" afterDir="false" />
+      <change beforePath="$PROJECT_DIR$/bin/dev" beforeDir="false" />
+      <change beforePath="$PROJECT_DIR$/config/routes.rb" beforeDir="false" afterPath="$PROJECT_DIR$/config/routes.rb" afterDir="false" />
+      <change beforePath="$PROJECT_DIR$/yarn.lock" beforeDir="false" afterPath="$PROJECT_DIR$/yarn.lock" afterDir="false" />
+    </list>
+    <option name="SHOW_DIALOG" value="false" />
+    <option name="HIGHLIGHT_CONFLICTS" value="true" />
+    <option name="HIGHLIGHT_NON_ACTIVE_CHANGELIST" value="false" />
+    <option name="LAST_RESOLUTION" value="IGNORE" />
+  </component>
+  <component name="Git.Settings">
+    <option name="RECENT_GIT_ROOT_PATH" value="$PROJECT_DIR$" />
+  </component>
+  <component name="HighlightingSettingsPerFile">
+    <setting file="file://$PROJECT_DIR$/app/controllers/steps/claim_reason_controller.rb" root0="SKIP_INSPECTION" />
+  </component>
+  <component name="MarkdownSettingsMigration">
+    <option name="stateVersion" value="1" />
+  </component>
+  <component name="ProjectId" id="2Oe7o1gvLl54GpQjhKLBROQ5tdm" />
+  <component name="ProjectViewState">
+    <option name="hideEmptyMiddlePackages" value="true" />
+    <option name="showLibraryContents" value="true" />
+  </component>
+  <component name="PropertiesComponent"><![CDATA[{
+  "keyToString": {
+    "RunOnceActivity.ShowReadmeOnStart": "true",
+    "WebServerToolWindowFactoryState": "false",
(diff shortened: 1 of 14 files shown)


.idea/workspace.xml
```diff
@@ -0,0 +1,328 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dwhenry, line 1: can we add this to the `.gitignore` file

app/controllers/steps/claim_reason_controller.rb
```diff
@@ -0,0 +1,19 @@
+module Steps
+  class ClaimReasonController < Steps::BaseStepController
```
dwhenry, line 2: This is my fault, but think we are better off use @imtiazAhmd naming convestions for the steps (making this one ReasonForClaim instead

app/forms/steps/claim_reason_form.rb
```diff
@@ -0,0 +1,48 @@
+require 'steps/base_form_object.rb'
+
+module Steps
+  class ClaimReasonForm < Steps::BaseFormObject
+    attribute :claim_reason, :value_object, source: ClaimReason
+
+    attribute :rep_order_date, :multiparam_date
+    attribute :cntp_order, :string
+    attribute :cntp_date, :multiparam_date
```
dwhenry, line 9: we can removed the previous fields and validations are they can't be updated via this form

app/forms/steps/claim_reason_form.rb
```diff
@@ -0,0 +1,48 @@
+require 'steps/base_form_object.rb'
+
+module Steps
+  class ClaimReasonForm < Steps::BaseFormObject
+    attribute :claim_reason, :value_object, source: ClaimReason
```
dwhenry, line 5: so the claim reasons are a series of checkbox with no limit on how many can be selected at once. As such I think we will need/want a field for each checkbox (note some checkboxes then have an additional data requirement)

app/services/decisions/simple_decision_tree.rb
```diff
@@ -15,7 +15,7 @@ def after_claim_type
       if form_object.claim_type.supported?
         edit(:firm_details)
       else
-        index(:claims)
+        index(:claim_reason)
```
dwhenry, line 18: This shouldn't be changed

bin/dev
```diff
@@ -1,9 +0,0 @@
-#!/usr/bin/env bash
```
dwhenry, line 1: can we undelete this file... :)

db/migrate/20230423152339_create_claim_reasons.rb
```diff
@@ -0,0 +1,8 @@
+class CreateClaimReasons < ActiveRecord::Migration[7.0]
+  def change
+    create_table :claim_reasons, id: :uuid do |t|
+
+      t.timestamps
+    end
+  end
```
dwhenry, line 7: do we need a claim reasons table? while it could be usefully to allow creating new claim reasons on the fly, this is unlikely to work given that would also need to be replicated in the other apps we will create.

This also goes against the ValueObject concept we are using to hold the list. Happy to discuss further.

yarn.lock
```diff
@@ -4,23 +4,28 @@
 
 "@ministryofjustice/frontend@1.6.4":
   version "1.6.4"
-  resolved "https://registry.yarnpkg.com/@ministryofjustice/frontend/-/frontend-1.6.4.tgz#396ffdc1699253671c8bc93ffa879735b8d4e8da"
+  resolved "https://registry.npmjs.org/@ministryofjustice/frontend/-/frontend-1.6.4.tgz"
```
dwhenry, line 7: This will annoy me, but I honestly have no idea why you are connected to a different MOJ registry... :(

app/views/steps/claim_reason/edit.html.erb
```diff
@@ -0,0 +1,33 @@
...
+<% step_header(path: applications_path) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary(@form_object) %>
+    <%= step_form @form_object do |f| %>
+        <%= f.govuk_radio_buttons_fieldset :claim_reason, legend: { tag: 'h1', size: 'xl' } do %>
+          <% @form_object.options.each do |option| %>
+            <% if option == ClaimReason::REPRESENTATION_ORDER_WITHDRAWN %>
+              <%= f.govuk_check_box :claim_reason, option.value do %>
```
dwhenry, line 11: is value from VALUE_TABLE should match field names from table.

```<%= f.govuk_check_box option.value do %>```

.idea/workspace.xml
mhunt3142: added

app/controllers/steps/claim_reason_controller.rb
mhunt3142: should this be done for the entire app - all references to ClaimReason replaced with ReasonForClaim including filenames ?

app/services/decisions/simple_decision_tree.rb
mhunt3142: reverted

bin/dev
mhunt3142: restored

db/migrate/20230423152339_create_claim_reasons.rb
mhunt3142: removed

app/forms/steps/claim_reason_form.rb
mhunt3142: will this be in the Claim table or a seperate table ? will they be of type bool

PR #30, opened 2023-05-30
Claim details2

mhunt3142:
## Description of change
Claim Details

## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/c/projects/CRM457/boards/1192?modal=detail&selectedIssue=CRM457-27


app/forms/steps/claim_details_form.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    attribute :prosecution_evidence, :string
+    attribute :defence_statement, :string
+    attribute :number_of_witnesses, :integer
+    attribute :time_spent_hours, :integer
+    attribute :time_spent_mins,  :integer
+
+    validates :prosecution_evidence, presence: true
+    validates :defence_statement, presence: true
+    validates :number_of_witnesses, presence: true
```
dwhenry, line 15: should also add numeric validations
```
```

app/forms/steps/claim_details_form.rb
```diff
@@ -0,0 +1,32 @@
...
+
+module Steps
+  class ClaimDetailsForm < Steps::BaseFormObject
+    BOOLEAN_FIELDS = %i[supplemental_claim preparation_time].freeze
+
+    attribute :prosecution_evidence, :string
+    attribute :defence_statement, :string
+    attribute :number_of_witnesses, :integer
+    attribute :time_spent_hours, :integer
+    attribute :time_spent_mins,  :integer
```
dwhenry, line 11: any reason we don't have validations for the time fields?

dwhenry (approved):
Would be good to have additional validation son the integer fields..

Can be left for later, if desired - just worth men tioning to @matty so that she is aware this is something we have to return to (we will have to review validations on all pages)

app/forms/steps/claim_details_form.rb
mhunt3142: think it was because these fields were conditional - oversight - thanks

app/forms/steps/claim_details_form.rb
mhunt3142: yep - added - thanks

mhunt3142:
additional integer validations

PR #34, opened 2023-06-05
otherinfo wip

mhunt3142:
## Description of change
add otherinfo form

## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/c/projects/CRM457/boards/1192?modal=detail&selectedIssue=CRM457-50&quickFilter=3076


app/forms/steps/claim_details_form.rb
```diff
@@ -13,8 +13,8 @@ class ClaimDetailsForm < Steps::BaseFormObject
     validates :prosecution_evidence, presence: true
     validates :defence_statement, presence: true
     validates :number_of_witnesses, presence: true, numericality: { only_integer: true, greater_than: 0 }
-    validates :time_spent_hours, numericality: { only_integer: true, greater_than: 0 }
-    validates :time_spent_mins, numericality: { only_integer: true, greater_than: 0 }
+    # validates :time_spent_hours, numericality: { only_integer: true, greater_than: 0 }
+    # validates :time_spent_mins, numericality: { only_integer: true, greater_than: 0 }
```
dwhenry, line 17: These should be optional validations - I was planning to mention I had spotted this issue on Friday.

i.e.
```ruby
validates :time_spent_mins, numericality: { only_integer: true, greater_than: 0 }, if: :preparation_time
```

app/forms/steps/other_info_form.rb
```diff
@@ -0,0 +1,28 @@
...
+
+module Steps
+  class OtherInfoForm < Steps::BaseFormObject
+    BOOLEAN_FIELDS = %i[concluded].freeze
+
+    attribute :other_info, :string
+    attribute :conclusion, :string
+
+    validates :other_info, presence: true
+    validates :conclusion, presence: true
```
dwhenry, line 11: this should be another optional validation (I believe)..

so
```ruby
validates :conclusion, presence: true, if: :concluded
```

db/migrate/20230523134506_claim_details.rb
```diff
@@ -7,5 +7,8 @@ def change
     add_column :claims, :preparation_time, :string
     add_column :claims, :time_spent_hours, :integer
     add_column :claims, :time_spent_mins, :integer
+    add_column :claims, :other_info, :string
+    add_column :claims, :conclusion, :string
+    add_column :claims, :concluded, :string
```
dwhenry, line 12: other_info and conclusion should be text instead of string fields as they are text-boxes in the UI so likely to be longer than sting can accommodate.

spec/steps/other_info/form_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      end
+    end
+  end
+
+  describe '#valid?' do
+    context 'when all fields are set' do
+      it 'is valid' do
+        expect(form).to be_valid
+      end
+    end
```
dwhenry, line 35: we need to be testing the other validation criteria, expecially for something like `conclusion` which needs conditional validation and `concluded` which needs to have value in a specific set.

dwhenry (commented):
This also needs an integration test as this gives confidence in the various parts all working together - currently we are only unit testing, which is good but does not give that confidence

app/forms/steps/claim_details_form.rb
mhunt3142: yep did try this witjh an unless but couldnt get it to work

app/forms/steps/other_info_form.rb
mhunt3142: thanks - will use in future

db/migrate/20230523134506_claim_details.rb
mhunt3142: cheers - string goes to 255 chars text to 30000 chars

spec/steps/other_info/form_spec.rb
mhunt3142: will do

app/forms/steps/claim_details_form.rb
mhunt3142: validates :time_spent_hours, numericality: { only_integer: true, greater_than: 0 }, if: :preparation_time
still doesnt work ?
in debugger when it is set to yes in persist
(ruby) preparation_time
#<YesNoAnswer:0x000000010a479e78 @value=:yes>
(ruby) preparation_time == true
false
(ruby) preparation_time.value == true
false

db/migrate/20230523134506_claim_details.rb
mhunt3142: database defintion changed to text but form doesnt seem to recognised :text - saved a 10k string  in database

spec/steps/other_info/form_spec.rb
mhunt3142: hopefully resolved for conditionals

dwhenry: So the issue with the tests here is that they only ever test for when the input is valid. we should also have tests for when the input is not valid..


i.e.
Scenario 1:
concluded = 'other' 
result => is invalid

Scenario 2:
concluded = 'yes' 
conclusion = nil
result => is invalid
 
Scenario 3:
concluded = 'no' 
conclusion = 'some text'
result => is valid, but conclusion is NOT saved to DB

dwhenry (commented):
It is an improvement,  but still a bit light on test coverage as we need negative paths to be covered in the unit tests...

app/forms/steps/claim_details_form.rb
mhunt3142: validates :time_spent_mins, numericality: { only_integer: true, greater_than: 0 },
      if: ->(form) { form.preparation_time == YesNoAnswer::YES }

spec/steps/other_info/form_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+      other_info:,
+      conclusion:,
+      concluded:,
+    }
+  end
+
+  let(:application) { instance_double(Claim, update!: true) }
+  let(:other_info) { 'other relevent information' }
+
+  describe '#save concluded yes' do
```
dwhenry, line 18: It is fine for now, but should consider that you are testing the exact same things with both the `save` and `valid` tests. Things to keep in mind with this:

* valid are a better cover for these sets of tests as the conditions are related to validity
* save is not a method on the form - it is inherited, as such we really want to use it to test the `persist!` method
* We are not actually testing the persist! method is doing what we want

In your defence this is how I wrote the first couple of sets of tests I did and you have just copied them. Leave these as is for now, and I will try to do a test review at some stage in the future.

dwhenry (approved):

spec/steps/other_info/form_spec.rb
mhunt3142: it did strike me i was testing the same thing twice - should really use common sense - cheers

PR #57, opened 2023-06-09
[CRM457-75] FE: Letters and calls

imtiazAhmd:
## What
This will add the feature to update total as user fills up the form
## Ticket
[FE: Letters and calls](https://dsdmoj.atlassian.net/browse/CRM457-75)
## Why
As user types the letters and calls, total for the items should reflect as per the design.
## Acceptance Criteria

- As a user, when I type in the number of letters (and the number of calls) in their respective input boxes and then tab out of the box, the total cost will automatically update.

- If a user changes any fields (letters or calls) after the initial update, it will trigger the update of the total cost.

- When a user adds a number into the percentage box and then tabs out of the box, the total cost will update based on the uplift percentage.

- If the user unticks the uplift box, the percentage box will be emptied, the reveal field will be removed, and the total cost will be updated accordingly.   

## Demo
[Video link](https://vimeo.com/834709371/2780f367be)

## ToDo

- [x] Update view to add Letters and calls unit price as data attributes
- [x] Add JS functionality to reflect the Acceptance criteria
- [x] Remove placeholders from the fields
- [x] Add unit tests for all the acceptance criteria

Screenshot 2023-06-13 at 13 41 34

### DIFF
--- a/app/javascript/controllers/letters_calls_form_controller.js
+++ b/app/javascript/controllers/letters_calls_form_controller.js
@@ -1,3 +1,78 @@
+// Get the value of an element attribute as a number or a default value
+function getAttributeAsNumber(element, attributeName, defaultValue = 0) {
+	return parseFloat(element.getAttribute(attributeName)) || defaultValue;
+}
+
+// Calculate the total based on the values of letters and phone_calls inputs
+function calculateTotal(lettersValue, phoneCallsValue, unitPriceForLetters, unitPriceForPhoneCalls) {
+	const total = lettersValue * unitPriceForLetters + phoneCallsValue * unitPriceForPhoneCalls;
+	return parseFloat(total.toFixed(2)); // Format total to two decimal places
+}
+
+// Calculate the total with uplift
+function calculateTotalWithUplift(currentTotal, upliftPercentage) {
+	const upliftToAdd = currentTotal * (upliftPercentage / 100);
+	const totalWithUplift = (currentTotal + upliftToAdd);
+	return totalWithUplift;
+}
+
+// Update the DOM element with the provided total
+function updateDomElementTotal(totalAmount, element) {
+	totalAmount = parseFloat(totalAmount).toFixed(2);
+	element.innerHTML = '£' + totalAmount;
+}
+
+// Check if all necessary field values exist
+function checkIfAllFieldValueExists(lettersValue, phoneCallsValue) {
+	return lettersValue && phoneCallsValue;
+}
+
+// Handle change events on the form
+export function handleFormChange(event, form, totalElement, lettersElement, phoneCallsElement, upliftCheckboxElement, upliftPercentageElement) {
+	const lettersValue = parseFloat(lettersElement.value) || 0;
+	const phoneCallsValue = parseFloat(phoneCallsElement.value) || 0;
+	const unitPriceForLetters = getAttributeAsNumber(lettersElement, 'data-rate-letters');
+	const unitPriceForPhoneCalls = getAttributeAsNumber(phoneCallsElement, 'data-rate-phone-calls');
+	const upliftPercentage = parseFloat(upliftPercentageElement.value) || 0;
+
+	if (checkIfAllFieldValueExists(lettersValue, phoneCallsValue)) {
+		const currentTotal = calculateTotal(lettersValue, phoneCallsValue, unitPriceForLetters, unitPriceForPhoneCalls);
+		// Triggered when letters or phone calls value changes, or when uplift percentage changes
+		if ([lettersElement, phoneCallsElement, upliftPercentageElement].includes(event.target)) {
+			// Update total
+			const totalWithUplift = calculateTotalWithUplift(currentTotal, upliftPercentage);
+			updateDomElementTotal(totalWithUplift, totalElement);
+		}
+		if (event.target === upliftCheckboxElement && !upliftCheckboxElement.checked) {
+			// Clear uplift percentage value
+			upliftPercentageElement.value = '';
+			// Update total
+			const totalWithUplift = calculateTotalWithUplift(currentTotal, 0);
+			updateDomElementTotal(totalWithUplift, totalElement);
+		}
+	}
+}
+
+// Attach event listener to the form for change events
+function initializeForm(
+	form,
+	totalElement,
+	lettersElement,
+	phoneCallsElement,
+	upliftCheckboxElement,
+	upliftPercentageElement
+) {
+	form.addEventListener('change', (event) => handleFormChange(
+		event,
+		form,
+		totalElement,
+		lettersElement,
+		phoneCallsElement,
+		upliftCheckboxElement,
+		upliftPercentageElement
+	));
+}
+
 const form = document.getElementById('new_steps_letters_calls_form');
 const total_for_letters_and_phone_calls = document.getElementById('letters-calls-total');
 const letters = document.getElementById('steps-letters-calls-form-letters-field');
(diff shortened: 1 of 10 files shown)


app/views/steps/letters_calls/edit.html.erb
```diff
@@ -8,8 +8,8 @@
     <%= govuk_error_summary(@form_object) %>
 
     <%= step_form @form_object do |f| %>
-      <%= f.govuk_number_field :letters, placeholder: 0, min: 1, step: 1, width: 3 %>
-      <%= f.govuk_number_field :calls, placeholder: 0, min: 1, step: 1, width: 3 %>
+      <%= f.govuk_number_field :letters, placeholder: 0, min: 1, step: 1, width: 3, data:{rate_letters: @form_object.pricing.letters} %>
+      <%= f.govuk_number_field :calls, placeholder: 0, min: 1, step: 1, width: 3, data:{rate_letters: @form_object.pricing.calls} %>
```
pking-moj, line 12: Have you incorrectly copy+pasted here? Should this be rate_phone_calls?

app/javascript/controllers/letters_calls_form_controller.js
pking-moj, line 1: If the uplift percentage is defaulted to 0, would it make more sense (and reduce code) to call `calculateTotalWithUplift` whenever we need a total? calculate total can then become a private method in support of calculating the total

app/javascript/controllers/letters_calls_form_controller.js
```diff
@@ -0,0 +1,60 @@
...
+export function calculateTotalWithUplift() {
+    const currentTotal = parseFloat(calculateTotal()) || 0;
+    const upliftPercentage = parseFloat(letter_calls_uplift_percentage.value) || 0;
+    const upliftToAdd = currentTotal * (upliftPercentage / 100);
+    const totalWithUplift = (currentTotal + upliftToAdd).toFixed(2); // Format total to two decimal places
+    updateDomElementTotal(totalWithUplift);
+}
+
+// Update the DOM element with the provided total, or calculate the total if not provided
+export function updateDomElementTotal(total) {
```
pking-moj, line 28: Can this be private?

app/javascript/controllers/letters_calls_form_controller.js
imtiazAhmd: I need this methods (`calculateTotal` and `calcultaTotalWithUplift`) Public for JEST testing. I will double check if this is the case or not while working on the [SPIKE](https://dsdmoj.atlassian.net/browse/CRM457-88) and update accordingly

imtiazAhmd: In addition, by keeping these functions separate, we maintain a clear separation of concerns and improve code readability, maintainability, and reusability. 

Each function has a single responsibility, making the code easier for us to understand and modify in the future.

In future, I would like to pluck these methods out, modify a bit and put it in helper functions if we need to use it in multiple areas.

pking-moj: Ive got no problem with the methods remaining as they are as they do perform a function. However i cant see a case in the code below where we couldnt call `calculateTotalWithUplift` as it provides a default 0 value. If these are used in isolation i support them both being public however if we soley use `calculateTotalWithUplift` then we should make `calculateTotal` private and testing is done from the upper most method.

app/javascript/controllers/letters_calls_form_controller.js
```diff
@@ -0,0 +1,60 @@
...
+function updateDomElementTotal(total) {
+    const totalAmount = total || calculateTotal();
+    total_for_letters_and_phone_calls.innerHTML = '£' + totalAmount;
+}
+
+// Handle change events on the form
+function handleFormChange(event) {
+    if (checkIfAllFieldValueExists()) {
+        if (event.target === letters || event.target === phone_calls) {
+            updateDomElementTotal();
```
dwhenry, line 37: This will write the wrong value if the number of letters is updated after the uplift percentage has been set.

dwhenry (dismissed):
If the user sets the uplift and then updated either letter or call count it will set the total without the uplift result in the wrong value being displayed.

app/javascript/controllers/letters_calls_form_controller.js
imtiazAhmd: Thanks David. I have added a check to make sure it changes the total even after the uplift amount set. I have now added JEST successfully, so will write unit tests for all acceptance criteria.

app/javascript/controllers/letters_calls_form_controller.js
imtiazAhmd: Got it thanks. I have optimised the code further in my last commit after our discussion.

app/javascript/controllers/letters_calls_form_controller.js
```diff
@@ -0,0 +1,96 @@
...
+			upliftPercentageElement.value = '';
+			// Update total
+			const totalWithUplift = calculateTotalWithUplift(currentTotal, 0);
+			updateDomElementTotal(totalWithUplift, totalElement);
+		}
+	} else {
+		// Called when letters or phone calls value is empty
+		updateDomElementTotal(0, totalElement)
+	}
+}
```
dwhenry, line 57: I found the multiple paths that could lead to an update to the HTML (it is called in 3 spots) slightly hard to follow. Can I suggest that refactoring to something like the below might make it easier to others to follow:

```javascript
// Handle change events on the form
export function handleFormChange(event, form, totalElement, lettersElement, phoneCallsElement, upliftCheckboxElement, upliftPercentageElement) {
	// check if we care about this event
	if ([lettersElement, phoneCallsElement, upliftPercentageElement, upliftCheckboxElement].includes(event.target)) {

		// reset uplift percentage if required
		if (event.target === upliftCheckboxElement && !upliftCheckboxElement.checked) {
			// Clear uplift percentage value
			upliftPercentageElement.value = '';
		}

		// load the values so we can check them
		const lettersValue = parseFloat(lettersElement.value) || 0;
		const phoneCallsValue = parseFloat(phoneCallsElement.value) || 0;
		let totalWithUplift;

		// calculate the total value
		if (checkIfAllFieldValueExists(lettersValue, phoneCallsValue)) {
			// load the remaining values as we are now going to use them
			const unitPriceForLetters = getAttributeAsNumber(lettersElement, 'data-rate-letters');
			const unitPriceForPhoneCalls = getAttributeAsNumber(phoneCallsElement, 'data-rate-phone-calls');
			const upliftPercentage = parseFloat(upliftPercentageElement.value) || 0;

			// update the values
			totalWithUplift = calculateTotalWithUplift(currentTotal, upliftPercentage);
		} else {
			totalWithUplift = 0			
		}

		//update screen with calculated value
		updateDomElementTotal(totalWithUplift, totalElement);
	}	
}
```

dwhenry (commented):
I have struggled in places to follow the flow so would prefer to leave it for @ivan to review as i believe he is/has been a JS developer

app/javascript/controllers/letters_calls_form_controller.js
```diff
@@ -0,0 +1,94 @@
...
+export function handleFormChange(event, form, totalElement, lettersElement, phoneCallsElement, upliftCheckboxElement, upliftPercentageElement) {
+	if ([lettersElement, phoneCallsElement, upliftPercentageElement, upliftCheckboxElement].includes(event.target)) {
+		// Reset uplift percentage value if uplift checkbox is unchecked
+		if (event.target === upliftCheckboxElement && !upliftCheckboxElement.checked) {
+			// Clear uplift percentage value
+			upliftPercentageElement.value = '';
+		}
+		// Get the values from the form
+		const lettersValue = parseFloat(lettersElement.value) || 0;
+		const phoneCallsValue = parseFloat(phoneCallsElement.value) || 0;
```
ivanELEC, line 40: Is it safe to default to 0 if the lettersElement value is null? Are there instances where this could indicate an error?

app/javascript/controllers/letters_calls_form_controller.js
```diff
@@ -0,0 +1,94 @@
...
+		let totalWithUplift;
+		// Check if all necessary field values exist for the calculation
+		if (checkIfAllFieldValueExists(lettersValue, phoneCallsValue)) {
+			// Get the remaining required values from the form for the calculation
+			const unitPriceForLetters = getAttributeAsNumber(lettersElement, 'data-rate-letters');
+			const unitPriceForPhoneCalls = getAttributeAsNumber(phoneCallsElement, 'data-rate-phone-calls');
+			const upliftPercentage = parseFloat(upliftPercentageElement.value) || 0;
```
ivanELEC, line 47: Same as above re: defaulting to 0 with null values on an element

ivanELEC (approved):

app/javascript/controllers/letters_calls_form_controller.js
```diff
@@ -0,0 +1,94 @@
+// Get the value of an element attribute as a number or a default value
+function getAttributeAsNumber(element, attributeName, defaultValue = 0) {
+	return parseFloat(element.getAttribute(attributeName)) || defaultValue;
+}
+
+// Calculate the total based on the values of letters and phone_calls inputs
+function calculateTotal(lettersValue, phoneCallsValue, unitPriceForLetters, unitPriceForPhoneCalls) {
+	const total = lettersValue * unitPriceForLetters + phoneCallsValue * unitPriceForPhoneCalls;
```
ivanELEC, line 8: another very minor gripe - we don't use brackets here to separate out the two multiplications we are adding, but we use brackets on line 15 for an addition - not necessarily anything that needs changing but another case for us potentially bringing in linting as a way for us to have consistency

app/javascript/controllers/letters_calls_form_controller.js
imtiazAhmd: same as above

app/javascript/controllers/letters_calls_form_controller.js
imtiazAhmd: changed already, simple math BODMAS. But Agree with you for the readability. changed accordingly.

mhunt3142 (approved):

PR #72, opened 2023-06-16
Add Sentry to Application 

ivanELEC:
## Description of change

- Add Sentry to application initialiser
- Add test error to verify that Sentry is working (another PR is needed to remove this after testing)

## Link to relevant ticket
[Add Sentry to Application
](https://dsdmoj.atlassian.net/browse/CRM457-158)

## How to manually test the feature:

- Go to MOJ Sentry and login
- View Issues
- There should be an issue with title "test message"

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,16 +4,16 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby File.read('.ruby-version').strip
 
 gem 'bootsnap', require: false
-gem 'dartsass-rails', '~> 0.5.0'
+gem 'dartsass-rails', '~> 0.4.0'
 gem 'importmap-rails'
 gem 'laa_multi_step_forms', path: './gems/laa_multi_step_forms'
 gem 'pg', '~> 1.5'
 gem 'puma', '~> 6.3'
 gem 'rails', '~> 7.0.5'
+gem 'sentry-rails'
 gem 'sentry-ruby'
 gem 'sprockets-rails'
 gem 'tzinfo-data'
-gem "sentry-ruby"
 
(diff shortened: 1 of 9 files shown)


dwhenry (approved):
It's a thumbs up for me... if we are adding this do we also need to be considering an uptime monitoring or do we get that by default from the govuk platform?

kubectl_deploy/deployment.tpl
```diff
@@ -20,6 +20,8 @@ spec:
         env:
           - name: ENV
             value: Development
+          - name: SENTRY_DSN
+            value: https://b3bce9495e484872ae0ab2aeeaedbd54@o345774.ingest.sentry.io/4505369863651328
```
dwhenry, line 24: do we want to have this here as a value? or should it be a secret?

Gemfile
```diff
@@ -4,12 +4,14 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby File.read('.ruby-version').strip
 
 gem 'bootsnap', require: false
-gem 'dartsass-rails', '~> 0.5.0'
+gem 'dartsass-rails', '~> 0.4.0'
```
dwhenry, line 7: something has gone wrong in the merge???

dwhenry (commented):
Yes, but no... so the platform change in the Gemfilke has bitten me in the past... having it set to ruby is for the best

With the change you made it can get angry when we don't have one of the selected platforms...

to fix this run:
```
bundle lock --add-platform ruby
bundle lock --remove-platform arm64-darwin-22
bundle lock --remove-platform x86_64-linux
```

I think that is the correct order..

PR #83, opened 2023-06-22
Crm457 166 claim details

mhunt3142:
## Description of change
add 2 radio buttons with condition before and after dates

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRM457-166

## Notes for reviewer
rubocop shows up issues too many memoized helpers


app/views/steps/claim_details/edit.html.erb
```diff
@@ -15,6 +15,12 @@
             <% if field == :preparation_time %>
               <%= f.govuk_period_field :time_spent, width: 'one-third' %>
             <% end %>
+            <% if field == :work_before %>
+             <%= f.govuk_date_field :work_before_date, width: 'one-third' %>
+            <% end %>
+            <% if field == :work_after %>
+             <%= f.govuk_date_field :work_after_date, width: 'one-third' %>
+            <% end %>
```
dwhenry, line 23: can we try an cleanup the code in this loop...

```ruby
        <%= f.govuk_radio_buttons_fieldset field, legend: { size: 'm' } do %>
          <%= f.govuk_radio_button field, YesNoAnswer::YES.to_s do%>
            <%= 
                     if field == :preparation_time
                       f.govuk_period_field :time_spent, width: 'one-third'
                     elsif %i[work_before work_after].include?(field)
                       f.govuk_date_field "#{field}_date", width: 'one-third' %>
                     end
            %>
          <%= f.govuk_radio_button field, YesNoAnswer::NO.to_s %>
        <% end%>
```
the goal here is to get something that it easy to read for a dev when they come in...

> NOTE: Happy for you to do this in a different way, just not what you had. :)

db/migrate/20230523134506_claim_details.rb
```diff
@@ -7,5 +7,9 @@ def change
     add_column :claims, :preparation_time, :string
     add_column :claims, :time_spent_hours, :integer
     add_column :claims, :time_spent_mins, :integer
+    add_column :claims, :work_before, :string
+    add_column :claims, :work_after, :string
+    add_column :claims, :work_before_date, :date
```
dwhenry, line 12: given we can determine if this work_before is required based on the presence of the `work_before_date` being populated in the DB. There is a strong argument that we don't need to store `work_before` as a yes no value (same for work_after).

This is the same as the approach for the uplift fields.

Can you look at this and see if you can get it to work.

dwhenry (commented):
Biggest issue here for me is storing fields that don;t add value (i.e. work_before/work_after flags).

app/forms/steps/claim_details_form.rb
```diff
@@ -2,26 +2,53 @@
...
       validates field, presence: true, inclusion: { in: YesNoAnswer.values }
       attribute field, :value_object, source: YesNoAnswer
     end
 
     def boolean_fields
-      self.class::BOOLEAN_FIELDS
+      self.class::VIEW_BOOLEAN_FIELDS
+    end
+
+    # state hasnt been set yet and of no use validate on save
```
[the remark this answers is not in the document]
mhunt3142: removed the without

app/forms/steps/claim_details_form.rb
```diff
@@ -2,26 +2,49 @@
...
+    def preparation_time
+      true
+    end
+
+    def work_before
+      true
+    end
+
+    def work_after
+      true
```
dwhenry, line 47: I think this will result in the fields being checked when the page load??? and if you submit the form with invalid data they will be selected again even if deselected previously..

You should follow the pattern for how the apply_uplift method works

app/forms/steps/claim_details_form.rb
```diff
@@ -2,26 +2,49 @@
...
     validates :prosecution_evidence, presence: true
     validates :defence_statement, presence: true
     validates :number_of_witnesses, presence: true, numericality: { only_integer: true, greater_than: 0 }
     validates :time_spent, numericality: { only_integer: true, greater_than: 0 }, time_period: true,
-      if: ->(form) { form.preparation_time == YesNoAnswer::YES }
+      if: ->(form) { @preparation_time == "yes"}
+    validates :work_before_date, presence: true, multiparam_date: { allow_past: true, allow_future: false },
+      if: ->(form) { @work_before == "yes"}
+    validates :work_after_date, presence: true, multiparam_date: { allow_past: true, allow_future: false },
+      if: ->(form) { @work_after == "yes"}
```
dwhenry, line 26: we should be using the `@` accessor here, that is what the methods below are for. This would result in issues as we use the validation process in the task list to determine if the step is complete.

Additionally we should be using YesNoAnswer::YES over hardcoding the `'yes'` string

spec/steps/claim_details/form_spec.rb
```diff
@@ -105,4 +112,18 @@
...
+      context 'when all fields are set' do
+        it { expect(subject).to be_valid }
+      end
+
+      context 'when work_before_date is blank' do
+        let(:work_before_date) { nil }
+
+        it { expect(subject).to be_valid }
+      end
+  end
```
dwhenry, line 128: so... we don't have a test when the value is `yes`, we are also missing tests for work_after

spec/steps/start_page/integration_spec.rb
```diff
@@ -8,12 +8,21 @@
...
   let(:rep_order_date) { Date.yesterday }
+  
+
+  it 'can do green path' do
+    visit steps_start_page_path(claim.id)
+
+    within('.moj-task-list__item', text: 'Your details') do
+      expect(page).to have_content('In progress')
+    end
+  end
```
dwhenry, line 19: is this a mistake?

dwhenry (commented):
Left some comments as this will add bugs to the system.

app/views/steps/claim_details/edit.html.erb
ivanELEC: This reads ok to me - my only confusion is the condition in line 18. Could this be a helper method that takes the field and the name of the method describes the intention of it?

ivanELEC (approved):
approved but added a comment for a potential improvement

spec/steps/claim_details/form_spec.rb
dwhenry: still missing test for when it value is yes here

dwhenry (commented):
Just the missing test.

I also assume you test the correct options are selected when revisiting the view?

ivanELEC (approved):

spec/steps/claim_details/form_spec.rb
```diff
@@ -31,8 +35,10 @@
       let(:preparation_time) { 'yes' }
       let(:time_spent_hours) { 2 }
       let(:time_spent_mins) { 40 }
+      let(:work_before_date) { Date.new(2023, 1, 1) }
+      let(:work_after_date) { Date.new(2023, 1, 1) }
 
-      it 'is valid' do
+      it 'is not valid' do
```
dwhenry, line 41: ??? that looks valid?

spec/steps/claim_details/form_spec.rb
```diff
@@ -105,4 +120,60 @@
       end
     end
   end
+
+  context 'when work_before is yes' do
```
dwhenry, line 124: so this is still missing a number of edge cases (specifically it has no tests for an y of the invalid criteria. Lets spend some time this afternoon paring on this...

dwhenry (commented):
As per comment lets try and find some time to pair today

config/locales/en/helpers.yml
```diff
@@ -37,8 +37,12 @@ en:
         defence_statement: Defence statement
         number_of_witnesses: Number of Defence witnesses
         supplemental_claim: Does this bill represent supplemental claim ?
-        preparation_time: Was any preparation time spent watching or listening to taped evidence ?
+        preparation_time: Did you spend time watching or listening to recorded evidence ?
```
pking-moj, line 40: space prepending the `?` at the end. Also the same for the entry above

config/locales/en/helpers.yml
```diff
@@ -37,8 +37,12 @@ en:
...
         time_spent: Running time of the recording
+        work_before_date: Date you started this work
+        work_after_date: Date of last court hearing
+        work_before: Did you do any work before the representation order date ?
```
pking-moj, line 44: space prepending the `?` at the end

db/migrate/20230523134506_claim_details.rb
```diff
@@ -4,8 +4,9 @@ def change
     add_column :claims, :defence_statement, :string
     add_column :claims, :number_of_witnesses, :integer
     add_column :claims, :supplemental_claim, :string
-    add_column :claims, :preparation_time, :string
     add_column :claims, :time_spent_hours, :integer
     add_column :claims, :time_spent_mins, :integer
+    add_column :claims, :work_before_date, :date
+    add_column :claims, :work_after_date, :date
```
dwhenry, line 10: This is will break the server as this migration has already been run in the past, so the new fields WILL NOT be added to the database.

The the following in the console:
```
rails g migration <name of migration>
```

This will generate a new file in then `db` directory and so can add the column changes there...

spec/steps/claim_details/form_spec.rb
```diff
@@ -31,8 +39,10 @@
       let(:preparation_time) { 'yes' }
       let(:time_spent_hours) { 2 }
       let(:time_spent_mins) { 40 }
+      let(:work_before_date) { Date.new(2023, 1, 1) }
+      let(:work_after_date) { Date.new(2023, 1, 1) }
 
-      it 'is valid' do
+      it 'is is valid' do
```
dwhenry, line 45: typo

spec/steps/claim_details/form_spec.rb
```diff
@@ -46,6 +56,8 @@
 
     context 'when preparation is yes' do
       let(:preparation_time) { 'yes' }
+      let(:work_before_date) { Date.new(2023, 1, 1) }
+      let(:work_after_date) { Date.new(2023, 1, 1) }
```
dwhenry, line 60: these additional lets are not required here as they are already added at the top level

spec/steps/claim_details/form_spec.rb
```diff
@@ -62,47 +74,77 @@
 
       context 'time_spent has a value in the database' do
         let(:time_spent_in_db) { 100 }
+        let(:work_before_date) { Date.new(2023, 1, 1) }
+        let(:work_after_date) { Date.new(2023, 1, 1) }
```
dwhenry, line 78: these additional lets are not required here as they are already added at the top level

spec/steps/claim_details/form_spec.rb
```diff
@@ -62,47 +74,77 @@
...
 
         it 'clears the database field' do
           expect(subject.save).to be_truthy
           expect(application.reload).to have_attributes(
-            time_spent: nil
+            time_spent: 160
```
dwhenry, line 83: so you have changed the test results here and I am not 100% sure why. This is a test to confirm the DB field is cleared, but that is not longer the result?

dwhenry (commented):
Close... only two main issues:

* The migration file which is a blocker
* Change to test results that doesn't make sense - happy to pair on this if it will help.

dwhenry (approved):

PR #84, opened 2023-06-22
Pking moj/helm deploy

pking-moj:
## Description of change

Provide the ability to deploy the application to multiple environments using Helm as the kubernetes tool to assist with multple envs. Also updated the gitactions to add in protection/checks when deploying

## Link to relevant ticket

n/a

## Notes for reviewer

Dev is still using the old method until we are happy with Helm. Prod can only be deployed to main with an approval from a Senior Engineer (Ivan and Paul currently)


dwhenry (approved):
trusting that it works as I don't really understand it. :)

.github/workflows/test-build-deploy.yml
```diff
@@ -110,7 +110,6 @@ jobs:
   build:
     runs-on: ubuntu-latest
     needs: [gem-test, test]
-    if: github.ref == 'refs/heads/main'
```
dwhenry, line 113: are you sure this should be removed... as that would mean it would deploy on every push/push with an open PR...  (I can never remember which)

dwhenry (changes requested):
would change what is deployed to dev

ivanELEC (approved):

ivanELEC (approved):

PR #89, opened 2023-06-26
Crm457-147: Claims Pagination

ivanELEC:
## Description of change

- Added Pagination to Claims Page 
- Setup Kaminari with config
- Setup GOV UK elements for pagination

## Link to relevant ticket
[Add Pagination to Claims Table](https://dsdmoj.atlassian.net/browse/CRM457-147)

## TO DO

- Add Unit Tests

### Before changes:
image

### After changes:
image

1. Go to application
2. Login
3. View Claims Page upon login


app/helpers/pagination_helpers.rb
```diff
@@ -0,0 +1,19 @@
...
+  private
+
+  # Override this method in your class to change the default
+  def default_page_size
+    5
+  end
+
+  def page_size
+    limit = params[:limit].to_i
+    limit.positive? ? limit : default_page_size
```
dwhenry, line 13: do we want to add a max page size as well?
```ruby
limit.positive? ? [limit, maximum_page_size].min : default_page_size
```

gems/laa_multi_step_forms/app/assets/stylesheets/laa_multi_step_forms/local/_navigation.scss
```diff
@@ -0,0 +1,56 @@
+// Navigation styles
+
+nav {
```
dwhenry, line 3: what did this affect?

dwhenry (dismissed):
paginated... couple of small questions

gems/laa_multi_step_forms/app/assets/stylesheets/laa_multi_step_forms/local/_navigation.scss
ivanELEC: I should be able to remove these now - I was originally following CCCD exactly which had some custom css that it used in place of the govuk-frontend styling. I redid it so that it would just use the govuk-frontend styling

app/helpers/pagination_helpers.rb
ivanELEC: I think this is a good idea - just to prevent users trying to load too much data at a time I guess

gems/laa_multi_step_forms/app/assets/stylesheets/laa_multi_step_forms/local/_navigation.scss
ivanELEC: may a max of 100?

app/helpers/pagination_helpers.rb
ivanELEC: I don't need to add this logic, there is a kaminari_config file in config/initialisers that has a setting for max_per_page

dwhenry (approved):

PR #91, opened 2023-06-26
Crm457 151 add laa reference

ivanELEC:
## Description of change

- Add laa_reference field to claims
- Add generate_laa_reference method to claims (private) to generate a reference with format LAA-[6 alphanumeric characters]
- Make laa_reference generate when a new claim is created
- Add LAA Reference as field on Claims table and Claim task list
- Order records on claim page by updated_at field

## Link to relevant ticket
[Add functionality to generate and store LAA Reference](https://dsdmoj.atlassian.net/browse/CRM457-151)

### Before changes:
image
image

### After changes:
image
image

Claim reference generation
- Go to claims page
- Click Make a New Claim
- Submit claim type
- Go back to claims page (should see claim there with an LAA reference generated)

Check claims page and claim task list to see laa reference


spec/controllers/claims_controller_spec.rb
```diff
@@ -5,8 +5,8 @@
     let(:scope) { double(:scope, not: [instance_double(Claim)]) }
 
     before do
-      allow(Claim).to receive(:where).and_return(scope)
       # rubocop:disable RSpec/MessageChain
```
dwhenry, line 8: do we want to disable this rule globally if it is helpful to use?

app/models/claim.rb
```diff
@@ -1,4 +1,5 @@
 class Claim < ApplicationRecord
+  before_create :generate_laa_reference
```
dwhenry, line 2: I'm not a big fan of these style of callbacks - will discuss in a call...

dwhenry (approved):
Would like a quick chat before you merge if possible. :)

dwhenry (approved):

PR #131, opened 2023-07-07
CRM457-66 - Upload Evidence

dwhenry:
## Description of change

Provides an Upload Evidence page within the application flow. This features a drag and drop field to be able to upload files as supporting evidence to the claim

## Link to relevant ticket

[CRM457-66](https://dsdmoj.atlassian.net/browse/CRM457-66)

[CRM457-66]: https://dsdmoj.atlassian.net/browse/CRM457-66?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/views/steps/supporting_evidence/edit.html.erb
```diff
@@ -0,0 +1,65 @@
+<% title t('.page_title') %>
+<% step_header %>
+
+<%= tag.div class: "govuk-grid-row" do -%>
```
dwhenry, line 4: where there is no logic I am inclinded to use raw html... so the above line becomes:
```
<div class="govuk-grid-row">
```

app/views/steps/supporting_evidence/edit.html.erb
```diff
@@ -0,0 +1,65 @@
...
+      <%= tag.li t('.para2.attendance') %>
+      <%= tag.li t('.para2.reading') %>
+      <%= tag.li t('.para2.documentation') %>
+      <%= tag.li t('.para2.prosecution') %>
+      <%= tag.li t('.para2.emails') %>
+      <%= tag.li t('.para2.representation') %>
+    <% end -%>
+    <%= form_with model: @form_object do |form| %>
+      <%# Postal address component %>
+      <%= tag.div class: "govuk-form-group" do -%>
```
dwhenry, line 39: This can all be render using the form builder helpers:
```
<%= f.govuk_check_boxes_fieldset :postal_field, multiple: false, legend: nil do %>  
   <%= f.govuk_check_box :postal_field, 'true', 'false', multiple: false do %>
        <div class="govuk-form-group">
            <pre class="govuk-body govuk-!-margin-bottom-6">
              <%= t('.postal_address') %>
            </pre>
        </div>
  <% end %>
<% end %>

```

app/javascript/file-upload.js
```diff
@@ -0,0 +1,11 @@
+import MOJFrontend from '@ministryofjustice/frontend'
+import $ from 'jquery'
+window.$ = $
+
+if (typeof MOJFrontend.MultiFileUpload !== "undefined") {
+    new MOJFrontend.MultiFileUpload({
+        container: $(".moj-multi-file-upload"),
```
imtiazAhmd, line 7: If we are using jquery only for the element selection, can we try not to use jQuery, we can try rewrite this as : `document.querySelectorAll('.moj-multi-file-upload')`

app/controllers/steps/supporting_evidence_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      SupportingEvidence.destroy(params[:resource_id])
+    end
+
+    private
+
+    def decision_tree_class
+      Decisions::SimpleDecisionTree
+    end
+
+    def supporting_evidence
```
dwhenry, line 39: this does not appear to be used anywhere?

app/javascript/file-upload.js
```diff
@@ -0,0 +1,66 @@
...
+        </tr>`
+
+        return html;
+    };
+
+    MOJFrontend.MultiFileUpload.prototype.onFileDeleteClick = function (e) {
+        e.preventDefault(); // if user refreshes page and then deletes
+        var button = $(e.currentTarget);
+        $.ajax({
+            url: `${this.params.deleteUrl}?resource_id=${button[0].attributes[2].value}`,
```
dwhenry, line 26:
```
button[0].attributes[2].value
```
what does the 2 mean? just seems dangerous

app/models/main_offence.rb
```diff
@@ -52,7 +52,7 @@ def self.all
       new('Use of firearm to resist arrest'),
       new('Possession of firearm with criminal intent'),
       new('Carrying loaded firearm in public place'),
-      new('Possession of firearm without certificate '),
+      new('Possession of firearm without certificate'),
```
dwhenry, line 55: fixed elsewhere so should disappear

app/presenters/tasks/other_info.rb
```diff
@@ -6,5 +6,9 @@ class OtherInfo < Generic
     def path
       edit_steps_other_info_path(application)
     end
+
+    def completed?
+      Steps::OtherInfoForm.build(application).valid?
```
dwhenry, line 11: you get this for free by setting the `FORM` value

config/initializers/aws.rb
```diff
@@ -0,0 +1,10 @@
+require 'aws-sdk-s3'
+
+if Rails.env.production?
+  Aws.config.update({
+                      region: ENV.fetch('AWS_REGION', nil),
+                      credentials: Aws::Credentials.new(ENV.fetch('AWS_ACCESS_KEY_ID', nil), ENV.fetch('AWS_SECRET_ACCESS_KEY', nil))
+                    })
+
+  S3_BUCKET = Aws::S3::Resource.new.bucket(ENV.fetch('S3_BUCKET', nil))
```
dwhenry, line 9: is this thread safe?

I wouldn't be confident unless it was specifically said to be as such I would be included to have a method to create a new bucket object when I needed it.

spec/decisions/simple_decision_tree_spec.rb
```diff
@@ -139,6 +139,8 @@
   it_behaves_like 'a generic decision', :solicitor_declaration, :claim_confirmation, Steps::SolicitorDeclarationForm,
                   action_name: :show
 
+  it_behaves_like 'a generic decision', :supporting_evidence, :check_answers, Steps::SupportingEvidenceForm, action_name: :show
+
```
dwhenry, line 143: prefer to have in order of pages as it makes finding the test slightly easier. :)

dwhenry (commented):
hopefully the comments are helpful.. :)

app/controllers/steps/supporting_evidence_controller.rb
pking-moj: This is called in the before_action so that we can display the existing evidence on the page

app/controllers/steps/supporting_evidence_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      @supporting_evidence = latest_evidence.nil? ? SupportingEvidence.new : latest_evidence
+    end
+
+    def latest_evidence
+      SupportingEvidence.where(claim_id: current_application.id)&.map
```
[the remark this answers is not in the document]
pking-moj: Provides us with an enumerator of SupportingEvidence to assist with showing this on the page

app/javascript/file-upload.js
pking-moj: Possibly a better way of doing this but need to work with @imtiazAhmd to get this working as we would like

app/models/supporting_evidence.rb
```diff
@@ -0,0 +1,13 @@
+class SupportingEvidence < ApplicationRecord
+  self.table_name = 'supporting_evidence'
+  belongs_to :claim
+
+  attribute :file_name
+  attribute :file_type
+  attribute :file_size
+  attribute :case_id
+  attribute :id
+  attribute :send_by_post
```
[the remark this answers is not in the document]
pking-moj: - send_by_post is hangover
- I would rather specify the attributes to avoid the "magic" - Its easy to see what the model shape is this way

app/presenters/tasks/other_info.rb
pking-moj: I believe my build broke and this sorted it. Will remove and see how i get on

app/models/supporting_evidence.rb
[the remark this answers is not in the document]
dwhenry: Not 100% against the attributes to avoid magic... but unless we are doing that everywhere I would push back... also very not rails approach to it should should be documented why we did this and ensure it is maintained in the future.

app/controllers/steps/supporting_evidence_controller.rb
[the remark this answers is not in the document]
dwhenry: so two things... The map shouldn't be required, and it can never be nil as it returns an empty association when it does not find anything.

app/javascript/file-upload.js
pking-moj: Resolved in a cleaner fashion

ivanELEC (approved):

ivanELEC (approved):

imtiazAhmd (dismissed):

imtiazAhmd (approved):

PR #147, opened 2023-07-13
[CRM457-213] Solicitor declaration step

imtiazAhmd:
## What
This is the page on the form that allows a user to confirm what they have submitted is correct. 

## Link to relevant ticket
[SOLICITOR DECLARATION: Build "Confirm the following" page](https://dsdmoj.atlassian.net/browse/CRM457-213)

## Notes for reviewer
After submission of the claim, status is set to 'completed'. We do need to confirm what to show when user clicks on the 'SUBMITTED' claim in the Main Claim list page. [ref.Screenshot below]. I left a placeholder with a TODO tag for us to change the URL later.
Screenshot 2023-07-13 at 12 26 38

## Screenshot
Screenshot 2023-07-13 at 12 22 45

### DIFF
--- a/app/controllers/steps/solicitor_declaration_controller.rb
+++ b/app/controllers/steps/solicitor_declaration_controller.rb
@@ -1,0 +1,20 @@
+module Steps
+  class SolicitorDeclarationController < Steps::BaseStepController
+    def edit
+      @form_object = SolicitorDeclarationForm.build(
+        current_application
+      )
+      Rails.logger.info(current_application[:status].to_json)
+    end
+
+    def update
+      update_and_advance(SolicitorDeclarationForm, as: :solicitor_declaration)
+    end
+
+    private
+
+    def decision_tree_class
+      Decisions::SimpleDecisionTree
+    end
+  end
+end
--- a/config/locales/en/steps.yml
+++ b/config/locales/en/steps.yml
@@ -165,6 +165,30 @@ en:
       edit:
         page_title: Other Relevant information
         heading: Other Relevant information
+    solicitor_declaration:
+      edit:
+        page_title: Confirm the following
+        heading: Confirm the following
+        subheading_html: |
+          <p class="govuk-body">You agree that:</p>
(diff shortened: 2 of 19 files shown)


config/locales/en/steps.yml
```diff
@@ -165,6 +165,30 @@ en:
...
+            <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+            <strong class="govuk-warning-text__text">
+              <span class="govuk-warning-text__assistive">Warning</span>
+              If you provide wrong or incomplete information, or are found to have committed benefit fraud, you may:
+              <ul class="govuk-list govuk-list--bullet govuk-!-font-weight-bold">
+                <li>be prosecuted</li>
+                <li>need to pay a financial penalty</li>
+              </ul>
+            </strong>
+          </div>
```
pking-moj, line 189: QQ: Should we be putting HTML into the locales files? Would this not be better in the view with the text here?

app/controllers/steps/solicitor_declaration_controller.rb
```diff
@@ -0,0 +1,20 @@
+module Steps
+  class SolicitorDeclarationController < Steps::BaseStepController
+    def edit
+      @form_object = SolicitorDeclarationForm.build(
+        current_application
+      )
+      Rails.logger.info(current_application[:status].to_json)
```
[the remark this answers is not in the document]
imtiazAhmd: Debugging log, Thanks for pointing out.

config/locales/en/steps.yml
imtiazAhmd: I would like to put this in the ERB, only reason I've done this way as line 75 'subheading_html' has HTML content and I thought this is the approach we are taking. Happy to change if my understanding is wrong.

imtiazAhmd: Saying so, I am changing this as suggested, translation to WELSH would be much simpler.

imtiazAhmd: Done!

PR #151, opened 2023-07-13
Crm457 19 check your answers pt 1

ivanELEC:
## Description of change:
Start of Check Your Answers Page

Done: 
- Your Details
- Defendant Details
- Case Details
- Hearing Details

## Link to relevant ticket
[CHECK YOUR ANSWERS: Build the page](https://dsdmoj.atlassian.net/browse/CRM457-210)

## Notes for reviewer

- Only review sections that are done, anything in TO DO may be done in another PR
- Need to understand what's going wrong with the other_info integration test

image

## TO DO 

- Case Disposal
- Claim Justification
- Claim Details
- Work Items
- Letters and phone calls
- Disbursement costs
- Other relevant information 
- Upload supporting evidence (will do this in separate PR)


app/controllers/steps/check_answers_controller.rb
```diff
@@ -0,0 +1,16 @@
+module Steps
+  class CheckAnswersController < Steps::BaseStepController
+    def show
+      @form_object = Steps::Shared::NoOpForm.build(
```
dwhenry, line 4: shouldn't need a form for a show page as it will not be submitting anything...

app/presenters/check_answers/base.rb
```diff
@@ -0,0 +1,19 @@
+module CheckAnswers
+  class Base
+    def translate_table_key(table, key, **opt)
+      I18n.t("steps.check_answers.show.sections.#{table}.#{key}.key", **opt)
```
dwhenry, line 4: so we have a separate `summary.yml` that I created for to store the translations for the summary page... I lean towards having these either in that file or their own. :)

app/presenters/check_answers/case_details_card.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    def title
+      I18n.t('steps.check_answers.groups.about_case.case_details.title')
+    end
+
+    # TO DO: update remittal to include date of remittal when CRM457-172 is done
+    # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+    def rows
+      [
```
dwhenry, line 22: this could be streamlined:
```ruby
def rows
   row_data.map do |key, value|
      {
          key: { text: translate_table_key(KEY, key), classes: 'govuk-summary-list__value-width-50' },
          value: { text: value }
        }
   end
end

def row_data
  {
    'main_offence' => case_details_form.main_offence,
    'main_offence_date' => case_details_form.main_offence_date&.strftime('%d %B %Y'),
   ...
  }
end
```
Then rows can be part of the base class and you only need to test the row_data... this would also help simplify your tests of noise..

app/presenters/check_answers/case_disposal_card.rb
```diff
@@ -0,0 +1,17 @@
...
+    def initialize(claim)
+      @case_disposal_form = Steps::CaseDisposalForm.build(claim)
+    end
+
+    def route_path
+      'case_disposal'
+    end
+
+    def title
+      I18n.t('steps.check_answers.groups.about_case.case_disposal.title')
```
dwhenry, line 14: you have a helper for this... any reason your not using it?

app/presenters/check_answers/defendant_details_card.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    def rows
+      all_rows = []
+      all_rows.concat main_defendant_rows
+      additional_defendants.each_with_index do |defendant, index|
+        all_rows.concat additional_defendant_rows(defendant, index + 1)
+      end
+      all_rows
+    end
```
dwhenry, line 26: suggestion
```ruby
main_defendant_rows + 
  additional_defendants.map.with_index do |defendant, index|
    additional_defendant_rows(defendant, index + 1)
  end
```

app/presenters/check_answers/letters_calls_card.rb
```diff
@@ -0,0 +1,17 @@
+module CheckAnswers
+  class LettersCallsCard < Base
```
dwhenry, line 2: this should try and leverage what was done for the summary if possible

app/presenters/tasks/generic.rb
```diff
@@ -8,14 +8,20 @@ def can_start?
...
     end
 
     private
 
     def record
       application
     end
+
+    def associated_form
+      return unless defined? self.class::FORM
```
dwhenry, line 22: could also have done this by passing in nil as the form.. but as I said this isn;'t really required

app/services/decisions/simple_decision_tree.rb
```diff
@@ -10,11 +10,12 @@ class SimpleDecisionTree < BaseDecisionTree
       claim_details: :work_item,
       work_item: :work_items,
       disbursement_cost: :disbursements,
-      other_info: :solicitor_declaration,
+      check_answers: :solicitor_declaration
```
dwhenry, line 13: This doesn't have to use the simple decision tree... as there si no input it could just be a hardcoded link on the screen if that is easier.....

dwhenry (commented):
Looking good... hopefully my comments are helpful... happy to chat if you have any questions on my meaning

app/presenters/check_answers/letters_calls_card.rb
ivanELEC: I was initially going to do this, but so much of it is different that I thought to separate it out

Cost summary version: 
image
No summation in the title, an overarching heading, a totals tow, heading column text different

Check answers version: 
image
We are showing the number of each and not the cost of each

app/presenters/check_answers/case_details_card.rb
ivanELEC: this is an option but it also limits what we can do with the key of each table, sometimes I need to pass in opt to the key e.g. in additional defendants

app/presenters/check_answers/defendant_details_card.rb
ivanELEC: tried doing something like this but it kept breaking the page, couldn't immediately figure out why so moved on in the interest of time to refactor later

PR #218, opened 2023-08-11
Crm457 381 other information update

ivanELEC:
## Description of change

- Update Text Formatting in Other Information Page
- Add Totals Section to Work Items in Check Answers
- Add Totals Section to Disbursements in Check Answers

## Link to relevant ticket

[Other information - content design updates to Figma
](https://dsdmoj.atlassian.net/browse/CRM457-381)

[Check Answers - Content Updates](https://dsdmoj.atlassian.net/browse/CRM457-365)


spec/factories/claims.rb
```diff
@@ -101,5 +113,9 @@
     trait :one_disbursement do
       disbursements { [build(:disbursement, :valid)] }
     end
+
+    trait :one_other_disbursement do
+      disbursements { [build(:disbursement, :valid_other_specific)] }
+    end
```
dwhenry, line 119: be carefull here we don't want 1000 traits, they become unmanagable

spec/factories/claims.rb
```diff
@@ -41,6 +41,18 @@
...
+      ufn { '20150612/001' }
+      main_offence { MainOffence.all.sample.name }
+      main_offence_date { Date.yesterday }
+
+      assigned_counsel { 'no' }
+      unassigned_counsel { 'no' }
+      agent_instructed { 'no' }
+      remitted_to_magistrate { 'yes' }
+      remitted_to_magistrate_date { Date.new(2023, 3, 1) }
+    end
```
dwhenry, line 54: so traits can be nested... so I would include the case_details trait to avoid duplicate declarations. :) 

or potentially this is just `:with_remittal` and has those fields and you use both trains in the test. :)

dwhenry (commented):
Just some comments on the traits...

But other than that I think the figma file is wrong, as my understanding here is that a new yes/no field would be added to ask if other information is required... and it is then only a required field if yes is selected (you can see the yes select, but not a no in the figma)....

otherwise this change dosen't really make sense.. :)

app/presenters/check_answers/other_info_card.rb
```diff
@@ -13,33 +13,44 @@ def initialize(claim)
     def row_data
       [
         {
-          head_key: 'other_info',
-          text: check_missing(claim.other_info.present?) do
-            ApplicationController.helpers.multiline_text(claim.other_info)
+          head_key: 'is_other_info',
```
dwhenry, line 16: this could potentially use the `process_boolean_value` method to simplify this code. :)

dwhenry (approved):
LGTM

PR #223, opened 2023-08-18
Crm457 404/notify app store

dwhenry:
## Description of change
build and send data message to app store. This is a WIP and class name/refactoring is still required before a full PR can be started

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRM457-404

## NOTES
reliant on https://github.com/ministryofjustice/laa-crime-application-store/pull/16 being merged first to ensure endpoints align


app/presenters/check_answers/base.rb
```diff
@@ -24,8 +24,8 @@ def row_data
     end
 
     def row_content(head_key, text, head_opts = {}, footer: false)
-      translated_heading = translate_table_key(section, head_key, **head_opts)
-      heading = translated_heading.start_with?('Translation missing:') ? head_key : translated_heading
+      heading = translate_table_key(section, head_key, **head_opts)
+      # heading = translated_heading.start_with?('Translation missing:') ? head_key : translated_heading
```
pking-moj, line 28: Any reason we are commenting this out?

dwhenry: Sorry, I did that as part of some refactoring I did as part of my original approach and then abandoned.. that being said we should have translations for everything and not be doing this...

I left it commented so I would remember to check and ensure removing it didn't break anything... and then forgot about it.. :(

app/services/notify_app_store/http_notifier.rb
```diff
@@ -0,0 +1,15 @@
+class NotifyAppStore
+  class HttpNotifier
+    include HTTParty
+
+    def post(message)
+      self.class.post("#{host}/application/", message).parsed_response
```
pking-moj, line 6: Just looking at the documentation - does this not need a content-type header? Also how well does this library handle auth?

dwhenry: umm... let me check that... :) actually yes, and also check the response instead of just ignore it....

app/presenters/check_answers/base.rb
dwhenry: have put this back in with a todo to look at it later...

app/services/notify_app_store/http_notifier.rb
pking-moj, line 1: The endpoint in environments has basic auth - we will need to cater for this in the headers.

app/services/notify_app_store/http_notifier.rb
```diff
@@ -0,0 +1,26 @@
...
+        :warning
+      else
+        raise "Unexpected response from AppStore - status #{response.code} for '#{message[:application_id]}'"
+      end
+    end
+
+    private
+
+    def host
+      ENV.fetch('APP_STORE_URL', 'http://localhost:8000')
```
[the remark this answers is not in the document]
dwhenry: if we want it to work on the dev environment - honestly, I thinking to only have it working locally until I have the caseworker side as that is when I will be able to work out any bugs...

but you are right there is no harm in inplementing this all now so I will do that (same with basic auth)

app/services/notify_app_store/http_notifier.rb
dwhenry: done

ivanELEC (approved):

app/services/notify_app_store.rb
```diff
@@ -0,0 +1,40 @@
+class NotifyAppStore
+  attr_reader :notifier, :message_builder
+
+  def initialize(claim:, scorer: NotifyAppStore::Scorer)
+    @message_builder = MessageBuilder.new(claim:, scorer:)
+    @notifier = notifier
+  end
+
+  def process
+    raise 'Sidekiq workers is not yet enabled' if ENV.key?('REDIS_URL')
```
pking-moj, line 10:
```suggestion
    raise 'Sidekiq workers is not yet enabled' if ENV.key?('REDIS_HOST')
```

dwhenry: so the ruby backed redis libraries will automatically config when a `REDIS_URL` is present... I don't think that works with `REIS_HOST`...

is there a specific reason for this request?

pking-moj: As discussed, previous projects we have developed used the REDIS_URL which came from infra code however we came unstuck when a major version of Redis changed how the url was formed (I think it was a username issue) so we moved the formation into the application to give us more power in the future.

pking-moj (approved):

ivanELEC (approved):

PR #230, opened 2023-08-22
Add redis to CI pipeline

dwhenry:
## Description of change
Add redis to CI so that tests will pass in https://github.com/ministryofjustice/laa-claim-non-standard-magistrate-fee-backend/pull/223

## Link to relevant ticket
N/A

### DIFF
--- a/.github/workflows/test-build-deploy-main.yml
+++ b/.github/workflows/test-build-deploy-main.yml
@@ -59,7 +59,13 @@ jobs:
         ports:
           - 5432:5432
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
-
+      redis:
+        image: redis
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
@@ -99,6 +105,8 @@ jobs:
 
       - name: Run testing framework
         run: yarn unit-test --passWithNoTests
+        env:
+          REDIS_HOST: redis
 
       - name: Upload rspec coverage (if failure)
         if: failure()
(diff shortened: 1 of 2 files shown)


.github/workflows/test-build-deploy-main.yml
```diff
@@ -99,6 +105,8 @@ jobs:
 
       - name: Run testing framework
         run: yarn unit-test --passWithNoTests
+        env:
+          REDIS_HOST: redis
```
ivanELEC, line 109: you should be able to get the whole REDIS_URL from kubernetes using: 

- name: REDIS_URL
              valueFrom:
                secretKeyRef:
                  name: nsm-elasticache
                  key: url
                  
  Instead of setting the host and constructing the URL

config/initializers/sidekiq.rb
```diff
@@ -25,6 +25,8 @@
   redis_url = "rediss://:#{ENV.fetch('REDIS_PASSWORD',
                                      nil)}@#{ENV.fetch('REDIS_HOST',
                                                        nil)}:6379"
+elsif ENV['REDIS_HOST'].present? && ENV['CI'] # for running on CI as no password
+  redis_url = "rediss://#{ENV.fetch('REDIS_HOST', nil)}:6379"
 end
```
ivanELEC, line 30: (This code is taken from CCCD)

we can use an environment variable called INLINE_SIDEKIQ to not have to run in separate process in development. Can add this: 
```
if ENV['INLINE_SIDEKIQ'].eql?('true')
  raise 'Sidekiq must be run using redis in production' unless Rails.env.development?
  require 'sidekiq/testing'
  Sidekiq::Testing.inline!
end
```

and have INLINE_SIDEKIQ=true in the local environment variables

ivanELEC: OR just have the conditional statement for this be if Rails.env.development?

ivanELEC: Also, there's an extra s here in redis on line 29

dwhenry: @ivanELEC  So we alreadu have Sidekiq set to inline, however one of the test is failed as it tries to connect to redis (see linked PR) I have not idea why this is happening.. and debugging it would be a length of time... adding the redis server makes the tests pass and can be done quickly....

rediss is basically https

PR #243, opened 2023-08-30
Update file upload

pking-moj:
## Description of change

- Moves File upload from ActiveStorage to a factory method so that we can save the location for retrieval on the other applications
- Update JS so that we can return messages based on upload/removal state


app/controllers/steps/supporting_evidence_controller.rb
```diff
@@ -3,43 +3,41 @@
...
     def update
       update_and_advance(SupportingEvidenceForm, as: :supporting_evidence)
     end
 
     def destroy
-      SupportingEvidence.destroy(params[:evidence_id])
-      render json: {
-        success: true
-      }
+      evidence = SupportingEvidence.find_by(id: params[:evidence_id])
```
dwhenry, line 34: prefer to use `claim.supporting_evidences.find_by` as this ensures scope (i.e. can't delete someone else evidence)

app/controllers/steps/supporting_evidence_controller.rb
```diff
@@ -3,43 +3,41 @@
 module Steps
   class SupportingEvidenceController < Steps::BaseStepController
     skip_before_action :verify_authenticity_token
-    before_action :supporting_evidence
+    before_action :supporting_evidence, :file_uploader
```
dwhenry, line 6: we don't need these here if we use the method to access it (over the `@` variable as it will just lazy load... :)

app/controllers/steps/supporting_evidence_controller.rb
```diff
@@ -3,43 +3,41 @@
...
+      unless SUPPORTED_FILE_TYPES.include? params[:documents].content_type
+        return return_error(nil, { message: 'Incorrect file type provided' })
+      end
 
-      render json: {
-        success: {
-          evidence_id: evidence.id
-        }
-      }, status: :ok
+      file_path = @file_uploader.upload(params[:documents])
```
dwhenry, line 22: so this becomes:
```
file_path = file_uploader.upload(params[:documents])
```

app/controllers/steps/supporting_evidence_controller.rb
```diff
@@ -3,43 +3,41 @@
...
 
     def create
-      evidence = SupportingEvidence.create(
-        file_name: params[:documents].original_filename,
-        file_type: params[:documents].content_type,
-        file_size: params[:documents].tempfile.size,
-        claim: current_application,
-        file: params[:documents]
-      )
+      unless SUPPORTED_FILE_TYPES.include? params[:documents].content_type
```
dwhenry, line 18: It's at this point I would suggest moving the logic here into a service to try and keep the controller as small as possible (also helps with testing...

```
def create
  creator = SupportEvidenceCreator.new(claim, document)
  if creator.create
    return_success({ evidence_id: evidence.id, file_name: params[:documents].original_filename })
  else
    return_error(e, { message: creator.error_message })
  end
end

File: app/services/file_upload/file_uploader.rb [hunk 5]
```diff
@@ -0,0 +1,15 @@
+module FileUpload
```
dwhenry, line 1: feel like this file might be noise.. in that it could easily be abstracted as a method on the Base class removing the need for one additional layer of indirection..

However I don't feel super strongly about this...

dwhenry (commented):
Would approve the the evidence delete needs the scope set at a minimum.. :) 

hope the other comments make sense, happy to chat if not. :)

app/services/file_upload/file_uploader.rb
pking-moj: Used a factory because the underlying uploader can be whatever and the code itself doesnt care.

app/controllers/steps/supporting_evidence_controller.rb
pking-moj: 👍

app/controllers/steps/supporting_evidence_controller.rb
pking-moj: Given the size of this PR already, this can be future enhancement

app/javascript/file-upload.js
```diff
@@ -17,17 +18,15 @@ MOJFrontend.MultiFileUpload.prototype.uploadFile = function (file) {
...
-            }
+            feedback.html(this.getSuccessHtml(`${response.success.file_name} has been uploaded`));
+            this.status.html(`${response.success.file_name} has been uploaded`);
             item.find(`a.remove-link.moj-multi-file-upload__delete`).attr("value", response.success.evidence_id ?? file.name)
             this.params.uploadFileExitHook(this, file, response);
         }, this),
         error: $.proxy(function (jqXHR, textStatus, errorThrown) {
+            feedback.html(this.getErrorHtml(jqXHR.responseJSON.error.message));
+            this.status.html(jqXHR.responseJSON.error.message);
+            this.feedbackContainer.find('.govuk-table__row.moj-multi-file-upload__row').last().remove();
```
dwhenry, line 29: I don't like this much as it assumes synchronous responses from the server... would prefer to remove by ID or some other identifier if possible... :)

dwhenry (approved):
not sure if I have a valid concern or not.. :)

dwhenry (approved):

ivanELEC (approved):

PR #302, opened 2023-10-26
CRM457-645 - Translate reasons for claim

pking-moj:
## Description of change

- Translate reasons for claim
- Update test

## Link to relevant ticket

[CRM457-645](https://dsdmoj.atlassian.net/browse/CRM457-645)

### DIFF
--- a/spec/services/notify_app_store/message_builder_spec.rb
+++ b/spec/services/notify_app_store/message_builder_spec.rb
@@ -86,7 +86,12 @@ RSpec.describe NotifyAppStore::MessageBuilder do
           'preparation_time' => nil,
           'prosecution_evidence' => nil,
           'reason_for_claim_other_details' => nil,
-          'reasons_for_claim' => ['enhanced_rates_claimed'],
+          'reasons_for_claim' => [
+            {
+              en: 'Enhanced rates claimed',
+              value: 'enhanced_rates_claimed'
+            }
+          ],
           'remitted_to_magistrate' => 'no',
           'remitted_to_magistrate_date' => nil,
           'rep_order_date' => '2023-01-01',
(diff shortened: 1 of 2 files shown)


imtiazAhmd (approved):

spec/services/notify_app_store/message_builder_spec.rb
```diff
@@ -86,7 +86,12 @@
           'preparation_time' => nil,
           'prosecution_evidence' => nil,
           'reason_for_claim_other_details' => nil,
-          'reasons_for_claim' => ['enhanced_rates_claimed'],
+          'reasons_for_claim' => [
+            {
+              en: 'Enhanced rates claimed',
+              value: 'enhanced_rates_claimed'
+            }
+          ],
```
dwhenry, line 94: so in the case worker app it is currently expecting this in the following format:
```
     "reasons_for_claim": {
        "en": [
            "Enhanced rates claimed",
            "Councel or agent assigned"
        ],
        "value": [
          "enhanced_rates_claimed",
          "councel_or_agent_assigned"
        ]
      },
```

pking-moj: This makes no sense to do. The function that is being used outputs a consistent way of providing the translations so we would have to hack it to provide the above output. There is also the thought process of mapping a translation to a value as we are not using an ordered list. I would rather we kept things consistent across the whole board.

dwhenry: The reason I would like the suggested approach is that it makes the case worker side extreme easy to process https://github.com/ministryofjustice/laa-assess-non-standard-magistrate-fee/blob/main/app/view_models/v1/claim_justification.rb

Processing an array of translated object on the caseworker side seemed magnitudes more complex when I considered it.

ON the provider side we could achieve this with a small change to the `translations` method to allow an array to be passed in. (currently supports nil and string values)

dwhenry (approved):

PR #313, opened 2023-11-01
Add feedback form

pking-moj:
## Description of change

- Add Feedback form and controller to app
- Add Mailer models and class
- Add initialiser
- Update deployment for API Key

Screenshot 2023-11-02 at 10 35 00


ivanELEC (approved):

app/controllers/about/feedback_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+        application_env: application_environment
+      ).deliver_later!
+    end
+
+    def feedback_params
+      params.require(:feedback).permit(:user_feedback, :user_email, :user_rating)
+    end
+
+    def application_environment
+      ENV.fetch('ENV', Rails.env).to_s
```
dwhenry, line 35: are we better off using the `ENV_NAME` here?

app/controllers/about/feedback_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    private
+
+    def submit_feedback
+      FeedbackMailer.notify(
+        user_email: feedback_params[:user_email],
+        user_rating: feedback_params[:user_rating],
+        user_feedback: feedback_params[:user_feedback],
+        application_env: application_environment
+      ).deliver_later!
```
dwhenry, line 27:
```
      FeedbackMailer.notify(
        application_env: application_environment
        **feedback_params
      ).deliver_later!
```

is the same. :)

dwhenry (approved):
looks good. :)

app/controllers/about/feedback_controller.rb
pking-moj: The env var we use is ENV on the helm deploy.

app/controllers/about/feedback_controller.rb
pking-moj: This throws an exception as expecting the params - there are default params so that might be the case

PR #316, opened 2023-11-06
CRM457-645 - Add Cookie Banner

pking-moj:
## Description of change

- Adds Turbo for inline cookie banner interactivity (This is opt-in to not break existing functionality
- Adds cookie banner with translation files
- Adds cookie concern to support
- Add partial for Google Analytics
- Adds and updates tests

## Link to relevant ticket

[CRM457-645](https://dsdmoj.atlassian.net/browse/CRM457-645)

## Notes for reviewer

First view of website will display cookie banner until accepted/rejected. To view again delete cookies

[CRM457-645]: https://dsdmoj.atlassian.net/browse/CRM457-645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -187,6 +187,7 @@ GEM
     globalid (1.2.1)
       activesupport (>= 6.1)
     google-protobuf (3.23.3)
+    google-protobuf (3.23.3-x86_64-linux)
     govuk-components (4.1.1)
       html-attributes-utils (~> 1.0.0, >= 1.0.0)
       pagy (~> 6.0)
@@ -262,6 +263,8 @@ GEM
     nio4r (2.5.9)
     nokogiri (1.15.4)
       mini_portile2 (~> 2.8.2)
+      racc (~> 1.4)
+    nokogiri (1.15.4-x86_64-linux)
       racc (~> 1.4)
     notifications-ruby-client (5.4.0)
       jwt (>= 1.5, < 3)
@@ -408,6 +411,8 @@ GEM
     sass-embedded (1.63.5)
       google-protobuf (~> 3.23)
       rake (>= 13.0.0)
+    sass-embedded (1.63.5-x86_64-linux-gnu)
+      google-protobuf (~> 3.23)
     sentry-rails (5.12.0)
       railties (>= 5.0)
       sentry-ruby (~> 5.12.0)
@@ -445,6 +450,10 @@ GEM
       patience_diff
     thor (1.2.2)
     timeout (0.4.0)
+    turbo-rails (1.4.0)
+      actionpack (>= 6.0.0)
+      activejob (>= 6.0.0)
+      railties (>= 6.0.0)
     tzinfo (2.0.6)
       concurrent-ruby (~> 1.0)
     tzinfo-data (1.2023.3)
@@ -467,6 +476,7 @@ GEM
 
 PLATFORMS
   ruby
+  x86_64-linux
 
 DEPENDENCIES
   aws-sdk-s3 (~> 1.136)
@@ -506,6 +516,7 @@ DEPENDENCIES
   simplecov-rcov
   sprockets-rails
   super_diff
+  turbo-rails (~> 1.4.0)
   tzinfo-data
 
 RUBY VERSION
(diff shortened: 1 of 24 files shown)


Gemfile.lock
```diff
@@ -467,6 +476,7 @@ GEM
 
 PLATFORMS
   ruby
+  x86_64-linux
```
dwhenry, line 479: I would prefer to not have this in here - I don't actually know why it happens, but what I do know if adding specific platforms seems to break stuff... :(

app/controllers/concerns/cookie_concern.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    render(partial: 'layouts/cookie_banner/success',
+           locals: { analytics_cookies_accepted: cookies[:analytics_cookies_set] })
+  end
+
+  def set_default_cookies
+    return set_default_analytics_cookies unless cookies[:cookies_preferences_set]
+
+    @analytics_cookies_accepted = ActiveModel::Type::Boolean.new.cast(cookies[:analytics_cookies_set])
+    @cookie_preferences_set = show_hide_cookie_banners
```
dwhenry, line 21: this is always true as you have a guard up about. :) (line 18)

app/controllers/concerns/cookie_concern.rb
```diff
@@ -0,0 +1,61 @@
...
+    update_analytics_cookies(cookie_params[:cookies_preference])
+
+    render(partial: 'layouts/cookie_banner/success',
+           locals: { analytics_cookies_accepted: cookies[:analytics_cookies_set] })
+  end
+
+  def set_default_cookies
+    return set_default_analytics_cookies unless cookies[:cookies_preferences_set]
+
+    @analytics_cookies_accepted = ActiveModel::Type::Boolean.new.cast(cookies[:analytics_cookies_set])
```
dwhenry, line 20: you use `ActiveModel::Type::Boolean` here, but are checking against a value of `yes` down below. not a big issue but would prefer a consistent approach. :)

app/javascript/application.js
```diff
@@ -3,8 +3,12 @@
 // https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#javascript
 import { initAll } from 'govuk-frontend'
 import accessibleAutocomplete from 'accessible-autocomplete'
+import '@hotwired/turbo-rails'
+
 initAll()
 
+Turbo.setFormMode('optin')
```
dwhenry, line 10: Question: does this mean we can remove the `data: { turbo: 'false' }` stuff we have been adding everywhere?

app/views/shared/_flash_banner.html.erb
```diff
@@ -10,7 +10,7 @@
     </div>
     <div class="govuk-notification-banner__content">
       <p class="govuk-notification-banner__heading">
-        <%= msg %>
+        <%= msg.html_safe %>
```
dwhenry, line 13: Pretty sure this would make us fail any and all security checks...

worth having a look at what has been done for caseworker...

```<p><%= sanitize(detail, tags: %w[a]) %></p>```

the key here is that it only allow the `a` tag to be passed in (can extend the list if required)

**Review by dwhenry — COMMENTED** [review]
Some hopefully helpful suggestions... 

only show stopped is the `html_safe` call on the flash message. :)

File: app/controllers/concerns/cookie_concern.rb (hunk 2 above)
Comment (pking-moj) [reply app/controllers/concerns/cookie_concern.rb:21]: First time you view the website or when the cookies expire this will be false

File: app/javascript/application.js (hunk 4 above)
Comment (pking-moj) [reply app/javascript/application.js:10]: Thats my understanding!

File: app/controllers/concerns/cookie_concern.rb (hunk 2 above)
Comment (dwhenry) [reply app/controllers/concerns/cookie_concern.rb:21]: no I mean the value will always be true if it gets to this line of code... so could replace this line with
```
@cookie_preferences_set = show_hide_cookie_banners
```

dwhenry (approved):
please see my clarification on the previous comment. 🍪

PR #327, opened 2023-11-13
CRM457-682 - Add equality questions to view only

pking-moj:
## Description of change

- Made view only report
- Added equality card
- Clean up some strings

## Link to relevant ticket

[Show Equality monitoring questions when viewing a claim](https://dsdmoj.atlassian.net/browse/CRM457-682)
[CONTENT Change - taped evidence](https://dsdmoj.atlassian.net/browse/CRM457-407)
[Update header](https://dsdmoj.atlassian.net/browse/CRM457-634)


spec/services/notify_app_store/message_builder_spec.rb
```diff
@@ -64,7 +64,7 @@
             'town' => 'Lawyer Town'
           },
           'first_hearing_date' => '2023-03-01',
-          'gender' => nil,
+          'gender' => 'm',
```
dwhenry, line 67: so these are the values -> they should be mapped to translations (I think)

pking-moj: These will need to be added to another ticket for mapping in that case

ivanELEC (approved):

PR #340, opened 2023-11-21
CRM457-546 - Fix bad translations

pking-moj:
## Description of change

- FIxes translations from strings to translation objects
- Update tests
- Adds helper method to valueobjects to help get name by id

## Link to relevant ticket

[Provider App: Add Translations to Application Version Payload](https://dsdmoj.atlassian.net/browse/CRM457-546)

### DIFF
--- a/app/models/matter_type.rb
+++ b/app/models/matter_type.rb
@@ -22,6 +22,10 @@ MatterType = Struct.new(:id, :description) do
   end
   # rubocop:enable Metrics/MethodLength
 
+  def self.name_by_id(matter_id)
+    @all.first { |item| item.id == matter_id }.name
+  end
+
   def name
     "#{id} - #{description}"
   end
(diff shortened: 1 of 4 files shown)


imtiazAhmd (approved):

app/models/outcome_code.rb
```diff
@@ -28,6 +28,10 @@ def self.all
   end
   # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
 
+  def self.name_by_id(outcome_id)
+    @all.first { |item| item.id == outcome_id }.name
```
dwhenry, line 32: can we change this to 
```
@all.first { |item| item.id == outcome_id }.description
```

as otherwise we are forcing the ID in the return value. :)

app/models/matter_type.rb
```diff
@@ -22,6 +22,10 @@ def self.all
   end
   # rubocop:enable Metrics/MethodLength
 
+  def self.name_by_id(matter_id)
+    @all.first { |item| item.id == matter_id }.name
+  end
```
dwhenry, line 27: same for here... `description` instead

PR #374, opened 2023-12-11
CRM457-783 - Add VAT flag to check answers cards

pking-moj:
## Description of change

- Add VAT registered flag to check answers card

## Link to relevant ticket

[CRM457-783](https://dsdmoj.atlassian.net/browse/CRM457-783)


dwhenry:
My only question here - and it may have already been addressed... but if we have a VAT flag on the firm... do we still need the VAT flag on the disbursements? does it become optional depending on if firm value is set?

dwhenry (approved):

PR #388, opened 2023-12-14
Add Brewfile and .gitignore some files

jsugarman:
## Description of change
Add Brewfile and .gitignore some files

The README contains instruction to brew bundle
but no Brewfile exists. This adds what appears to
be a minimal requirement of services to install
for this app.

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,0 +1,5 @@
+tap 'homebrew/cask'
+brew 'postgresql@14'
+brew 'redis', restart_service: true
+brew 'node'
+brew 'yarn'
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,11 @@ tmp/*
 /node_modules
 /yarn-error.log
 
+yarn-debug.log*
+.yarn-integrity
+package-lock.json
+Brewfile.lock.json
+
 /public/assets
 .byebug_history
 


Brewfile
```diff
@@ -0,0 +1,5 @@
+tap 'homebrew/cask'
+brew 'postgresql@14'
+brew 'redis', restart_service: true
```
pking-moj, line 3: Some developers do not run these on their system and opt to run them through docker. Not sure how you would like to handle this

jsugarman: Thanks @pking-moj 

The Procfile.dev, run by bin/dev, is reliant on sidekiq running, and therefore `redis` being installed. I think that if docker is being used to run the app(s) then it is, i believe, irrelevant whether you have redis installed locally or not.

So to my mind, this brew file just makes setup on mac for non-docker running simpler, and for docker running it would not make any difference.

jogilder: @jsugarman 

I agree if we are going to use brew bundle then we should have a Brewfile, either that or we have to specify what is needed for the application to run in the readme for local development. It also needs to include node and yarn.

Might it be preferable to remove bundle and instead have a docker setup for the other services? 

Can you split this PR so we can add the .gitignore and we can query how everyone prefers to work (docker vs local) and then add as required?

pking-moj (approved):

Brewfile
jsugarman: I have not got round to a docker setup running all the apps in conjunction as yet but as a newbie on the project it is pretty normal to come to a rails repo and be able to go through readme and do a `rails s` or `bin/dev` and be able to see the app.

happy to remove "brew bundle" (i did not actually need it as I had it all installed anyway) from setup. 

i only added the `.gitignore` to not store the `Brewfile.lock.json` primarily. the others were just additional things that are useful to ignore (taken from other repos/expreience). happy to split out regardless

PR #396, opened 2023-12-19
Crm457 786 - travel and waiting totals

ivanELEC:
## Description of change

- Add Costs model for all the Claim cost totals
- Create calculator for costs
- Add step when submitting claim to update claim with costs

## Link to relevant ticket
[Provider: Submit Travel and Waiting Total Data to App Store](https://dsdmoj.atlassian.net/browse/CRM457-840)


app/lib/cost_calculator.rb
```diff
@@ -0,0 +1,28 @@
+module CostCalculator
+  class << self
+    def cost(type, object, vat)
+      case type
+      when :travel_and_waiting_total
+        travel_and_waiting_total(vat, object)
+      end
```
jsugarman, line 7: **nit** could simplify until other "cases" come up

```ruby
travel_and_waiting_total(vat, object) if type == :travel_and_waiting_total
```

app/lib/cost_calculator.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    private
+
+    def travel_and_waiting_total(vat, object)
+      pricing = Pricing.for(object)
+      vat_rate = pricing.vat
+      total_without_vat = get_work_total(WorkItem.where(claim_id: object[:id], work_type: %w[travel waiting]), pricing)
```
jsugarman, line 15: extract and isolate query

```ruby
...
  total_without_vat = get_work_total(work_items(object), pricing)
end

def work_items(object)
  WorkItem.where(claim_id: object[:id], work_type: %w[travel waiting])
end
```

app/lib/cost_calculator.rb
```diff
@@ -0,0 +1,28 @@
...
+      total = vat ? total_without_vat * (1 + vat_rate) : total_without_vat
+      total.round(2)
+    end
+
+    def get_work_total(items, pricing)
```
jsugarman, line 20: naming, avoid use of get/set

```suggestion
    def work_items_total(items, pricing)
```

app/lib/cost_calculator.rb
```diff
@@ -0,0 +1,28 @@
...
+      items.sum { |i| get_item_total(i, pricing) }
+    end
+
+    def get_item_total(item, pricing)
```
jsugarman, line 24: naming

```suggestion
    def item_total(item, pricing)
```

db/migrate/20231219122845_change_cost_column_type.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeCostColumnType < ActiveRecord::Migration[7.1]
+  def change
+    rename_column :costs, :type, :cost_type
+  end
+end
```
jsugarman, line 5: **nit** either way, but could  combine this migration with the on above as they are part of the same PR. use of the term `type` (rather than cost_type) could, in any event provide a simple means to use single table inheritance pattern in future if deemed necessary. but i don't know enough about the intent and possible extension here so either way on the naming too.

db/schema.rb
```diff
@@ -78,6 +78,7 @@
     t.integer "defence_statement"
     t.integer "number_of_witnesses"
     t.string "supplemental_claim"
+    t.string "preparation_time"
```
jsugarman, line 81: **nit** seems to be a result of local db differences and could be excluded from diff

jsugarman (commented):
feel free to ignore the nits but a couple of other suggestions added that i think are worthwhile.

Also, a general point  related to the upcoming notion of Prior authority "applications" I am concerned about the use of the term "application" for "claims", because we may need to differentiate in the future 😬 .

spec/steps/solicitor_declaration/form_spec.rb
```diff
@@ -10,13 +10,16 @@
     }
   end
 
-  let(:application) { instance_double(Claim, update!: true, 'status=': true) }
+  let(:application) { create(:claim, :complete) }
```
jsugarman, line 13: Not in the PR alone but I see the use of the term application and claim being used interchangeably. This concerns me as with crm4 coming along we may need to differentiate 😬

jsugarman (commented):
general comment, Is this a `cost_totals` relation in reality where each record is a total of a specific cost type?  naming wise it does not reflect this as a "cost" object is really a total of other individual "costs" grouped by type.

db/migrate/20231218154837_create_costs.rb
```diff
@@ -0,0 +1,10 @@
+class CreateCosts < ActiveRecord::Migration[7.1]
+  def change
+    create_table :costs, id: :uuid do |t|
+      t.references :claim, null: false, foreign_key: true, type: :uuid
+      t.string :type, null: false
+      t.float :amount, null: false
+      t.float :amount_with_vat
+    end
```
dwhenry, line 8: did you consider storing the data in a flat structure instead of multiple records? 

I ask as that is what I had originally imagined - it doesn't make a heap of difference here, but would potentially be an easier data structure to use/query in the CW side. :)

app/lib/cost_calculator.rb
ivanELEC: going to extend this in my next ticket so want to leave it for now

db/migrate/20231218154837_create_costs.rb
ivanELEC: My preference is to keep is as a many relationship where the type differentiates it so that we don't have to keep extending the table if there are new totals that get added - and we can easily reuse this table if we use it for CRM4/5 etc. However probably true that it's easier to query on the CW side.

spec/steps/solicitor_declaration/form_spec.rb
ivanELEC: Yeah I think we need to have a wider chat about this as even I'm getting confused now on what the difference is between and claim and an application - as they seem to be interchangeable/definition is changing a lot of the time

ivanELEC:
> general comment, Is this a `cost_totals` relation in reality where each record is a total of a specific cost type? naming wise it does not reflect this as a "cost" object is really a total of other individual "costs" grouped by type.

Yeah I like this idea, costs feels more specific to the individual items and not the totals, will make this change

db/migrate/20231219122845_change_cost_column_type.rb
ivanELEC: Using 'type' got me into too many binds when trying to code this because it's a reserved column name

app/lib/cost_calculator.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    private
+
+    def travel_and_waiting_total(vat, object)
+      pricing = Pricing.for(object)
+      vat_rate = pricing.vat
+      total_without_vat = work_items_total(work_items(object), pricing)
+      total = vat ? total_without_vat * (1 + vat_rate) : total_without_vat
+      total.round(2)
```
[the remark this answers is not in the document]
dwhenry: actually this calculation need to be done at a line item level not teh summary.. :( so it is worst than this...

app/lib/cost_calculator.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    def work_items_total(items, pricing)
+      items.sum { |i| work_item_total(i, pricing) }
+    end
+
+    def work_item_total(item, pricing)
+      (item.time_spent / 60.to_f) * pricing[item.work_type] * (1 + (item[:uplift] / 100.to_f))
```
dwhenry, line 25: as per the above comment... this needs to be rounded
```
(item.time_spent.to_f / 60) * pricing[item.work_type] * (1 + (item[:uplift].to_f / 100)).round(2)
```

db/migrate/20231219122845_change_cost_column_type.rb
dwhenry: you can still use `type` as a columkn name and avoid the STI issues if change this value

https://apidock.com/rails/ActiveRecord/ModelSchema/ClassMethods/inheritance_column

app/lib/cost_calculator.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    def travel_and_waiting_total(vat, object)
+      pricing = Pricing.for(object)
+      total_without_vat = work_items_total(work_items(object), pricing)
+
+      return total_without_vat unless vat
+
+      total_with_vat = total_without_vat * (1 + pricing.vat)
+      total_with_vat.floor(2)
+    end
```
dwhenry, line 20:
```
    def travel_and_waiting_total(vat, object)
      pricing = Pricing.for(object)
      line_items = work_items(object)

      multiplier = vat ? (1.0 + pricing.vat) : 1.0
      items.sum { |i| (work_item_total(i, pricing)* multiplier).floor(2) }
    end
```
might work.... it also removed the work_items_total method

dwhenry (approved):
Ideally I would wrap the persist! implementation in a  transaction as it is now doing multiple DB events, but not really needed for as unlikely the second one would ever fail...

ivanELEC:
> Ideally I would wrap the persist! implementation in a transaction as it is now doing multiple DB events, but not really needed for as unlikely the second one would ever fail...

will update this when I do disbursements

PR #399, opened 2023-12-20
Crm457 841 disbursment totals 

ivanELEC:
## Description of change
- Added to CostCalculator calculation for disbursement totals
- Added to cost_totals object in application to include disbursements=
## Link to relevant ticket
[Provider: Submit Disbursement Total Data to App Store
](https://dsdmoj.atlassian.net/browse/CRM457-841)

### DIFF
--- a/app/forms/steps/solicitor_declaration_form.rb
+++ b/app/forms/steps/solicitor_declaration_form.rb
@@ -8,10 +8,12 @@ module Steps
     private
 
     def persist!
-      application.status = :submitted
-      build_costs
-      application.update!(attributes)
-      NotifyAppStore.new.process(claim: application)
+      Claim.transaction do
+        application.status = :submitted
+        build_costs
+        application.update!(attributes)
+        NotifyAppStore.new.process(claim: application)
+      end
     end
 
     def build_costs
@@ -35,8 +37,8 @@ module Steps
         },
         {
           'cost_type' => 'disbursements',
-          'amount_with_vat' => 0,
-          'amount' => 0
+          'amount_with_vat' => CostCalculator.cost(:disbursement_total, application, true),
+          'amount' => CostCalculator.cost(:disbursement_total, application, false)
         }
       ]
     end
--- a/app/lib/cost_calculator.rb
+++ b/app/lib/cost_calculator.rb
@@ -22,14 +22,14 @@ module CostCalculator
 
     def work_item_total(item, pricing, vat)
       rounded_cost = (item.time_spent.to_f / 60) * pricing[item.work_type] * (1 + (item[:uplift].to_f / 100)).round(2)
-      return (rounded_cost * (1 + pricing.vat)).floor(2) if vat == true
+      return (rounded_cost * (1 + pricing.vat)).floor(2) if vat
 
       rounded_cost
     end
 
     def disbursements_total(vat, object)
       pricing = Pricing.for(object)
-      if vat == true
+      if vat
         total = disbursements(object).sum do |i|
           vat_multiplier = i.apply_vat == 'true' ? pricing.vat : 0
           (i.total_cost_without_vat * (1 + vat_multiplier)).floor(2)
--- a/app/lib/cost_calculator.rb
+++ b/app/lib/cost_calculator.rb
@@ -5,7 +5,7 @@ module CostCalculator
       when :travel_and_waiting_total
         travel_and_waiting_total(vat, object)
       when :disbursement_total
-        disbursement_total(vat, object)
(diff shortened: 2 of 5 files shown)


jsugarman (commented):
Couple of suggestions but will leave to david to approve.

more generally, perhaps we should be storing amount with vat as well. but that is a bigger piece of work. CCCD does this i believe.

app/lib/cost_calculator.rb
```diff
@@ -25,8 +27,25 @@ def work_item_total(item, pricing, vat)
       rounded_cost
     end
 
+    def disbursements_total(vat, object)
+      pricing = Pricing.for(object)
+      if vat == true
+        total = disbursements(object).sum do |i|
+          vat_multiplier = i.apply_vat == 'true' ? pricing.vat : 0
+          (i.total_cost_without_vat * (1 + vat_multiplier)).floor(2)
+        end
```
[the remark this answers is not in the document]
ivanELEC: updated with the alternative option - as you said in Slack we would need total cost with vat to do the first option - which we could add in, but I'm hesitant to include ant vat changes before we have more general clarification

app/lib/cost_calculator.rb
```diff
@@ -25,8 +27,25 @@ def work_item_total(item, pricing, vat)
       rounded_cost
     end
 
+    def disbursements_total(vat, object)
+      pricing = Pricing.for(object)
+      if vat == true
```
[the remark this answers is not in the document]
ivanELEC: updated

app/forms/steps/solicitor_declaration_form.rb
```diff
@@ -35,8 +37,8 @@ def cost_totals(application)
         },
         {
           'cost_type' => 'disbursements',
-          'amount_with_vat' => 0,
-          'amount' => 0
+          'amount_with_vat' => CostCalculator.cost(:disbursement_total, application, true),
+          'amount' => CostCalculator.cost(:disbursement_total, application, false)
```
dwhenry, line 41: as a side note for a future refactoring it make more sense to change the CostCalculator interface so that:
```
amount_without_vat, amount_with_vat = CostCalculator.cost(:disbursement_total, application)
```
as this helps avoid having to run the same loops twice.

Thinking about doing something similar in CW to make that nicer as well...

app/lib/cost_calculator.rb
```diff
@@ -20,13 +22,33 @@ def work_items_total(items, pricing, vat)
...
+        disbursements_total_cost_with_vat(object, pricing).round(2)
+      else
+        disbursements(object).sum(&:total_cost_without_vat).round(2)
+      end
+    end
+
+    def disbursements_total_cost_with_vat(object, pricing)
+      disbursements(object).sum do |i|
+        vat_multiplier = i.apply_vat == 'true' ? pricing.vat : 0
+        (i.total_cost_without_vat * (1 + vat_multiplier)).floor(2)
```
dwhenry, line 42: so we actually already store a VAT amount for disbursements :(

we should work out if we actually want that and use it here or remove it as not required so that we have a single source of truth

app/lib/cost_calculator.rb
```diff
@@ -20,13 +22,33 @@ def work_items_total(items, pricing, vat)
...
+      end
+    end
+
     def work_items(object)
       WorkItem.where(claim_id: object[:id], work_type: %w[travel waiting])
     end
+
+    def disbursements(object)
+      Disbursement.where(claim_id: object[:id])
```
dwhenry, line 51: so could do this just as:

```
object.disbursements
```

dwhenry (approved):
Lots of little notes, but nothing that is actually a blocker.

I do think we should sit down and try to:
a. simplify our CostCalculator code
b. try and sync the implementation between provider and CW so that we are confident they give identical results.

app/lib/cost_calculator.rb
ivanELEC: yeah the reason I recalculated it here was because I was unsure of whether the VAT amounts stored in the table would be in line with the new calculations for summing VAT - I agree that we should have a single source of truth, thought I may leave this as it is for now until the stuff surrounding VAT is more clear

app/lib/cost_calculator.rb
ivanELEC: will rectify when we pick this back up

app/forms/steps/solicitor_declaration_form.rb
ivanELEC: agreed, will do this when this is all finalised in the core costs ticket

PR #453, opened 2024-01-22
Use govuk_phase_banner component

jsugarman:
## Description of change
Use govuk_phase_banner component

Simplify the partial to use the govuk-component
libraries phase banner component. For brevity and to
delegate styling out to this well maintained gem.

### DIFF
--- a/gems/laa_multi_step_forms/app/helpers/laa_multi_step_forms/application_helper.rb
+++ b/gems/laa_multi_step_forms/app/helpers/laa_multi_step_forms/application_helper.rb
@@ -26,6 +26,12 @@ module LaaMultiStepForms
       "app-environment-#{ENV.fetch('ENV', 'local')}"
     end
 
+    def phase_name
+      return t('layouts.phase_banner.phase') if ENV.fetch('ENV', 'local') == 'production'
+
+      ENV.fetch('ENV', 'local').capitalize
+    end
+
     def format_period(period)
       return if period.nil?
 
--- a/gems/laa_multi_step_forms/app/views/layouts/_phase_banner.html.erb
+++ b/gems/laa_multi_step_forms/app/views/layouts/_phase_banner.html.erb
@@ -1,14 +1,2 @@
-<div class="govuk-phase-banner" role="complementary">
-  <p class="govuk-phase-banner__content">
-    <strong class="govuk-tag govuk-phase-banner__content__tag">
-      <% if ENV.fetch('ENV', 'local') == 'production' %>
-        <%= t('.phase') %>
-      <% else %>
-        <%= ENV.fetch('ENV', 'local').capitalize %>
-      <% end %>
-    </strong>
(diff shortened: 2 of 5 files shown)


ivanELEC (approved):

gems/laa_multi_step_forms/app/views/layouts/_phase_banner.html.erb
```diff
@@ -1,14 +1,2 @@
...
-        <%= ENV.fetch('ENV', 'local') %>
-      <% end %>
-    </strong>
-    <span class="govuk-phase-banner__text">
-      <%= t('.phase_description', feedback: main_app.about_feedback_index_path).html_safe %>
-    </span>
-  </p>
-</div>
+<%= govuk_phase_banner(tag: { text: phase_name },
+                       text: t('.phase_description', feedback: main_app.about_feedback_index_path).html_safe) %>
```
patrick-laa, line 2: Nit: `html_safe`ing the output of `t` can occasionally get you into trouble, as it raises the possibility of XSS attacks if a user-entered string is ever passed in as an interpolation variable.

It's a good habit to get into to use `t('.blah_html', x: y)` instead of `t('.blah', x: y).html_safe`, as it will mean only the static template, not the interpolated variables, will be marked as safe.

I do appreciate that this particular line has no possibility of XSS and that you're only reusing the code that was already there, though!

gems/laa_multi_step_forms/app/helpers/laa_multi_step_forms/application_helper.rb
```diff
@@ -26,6 +26,12 @@ def app_environment
       "app-environment-#{ENV.fetch('ENV', 'local')}"
     end
 
+    def phase_name
+      return t('layouts.phase_banner.phase') if ENV.fetch('ENV', 'local') == 'production'
+
+      ENV.fetch('ENV', 'local')
```
patrick-laa, line 32: Now that we're on frontend v5, you'll want to capitalize this output as the tag helper won't do it for you.

gems/laa_multi_step_forms/config/locales/en/layouts.yml
```diff
@@ -0,0 +1,5 @@
+---
+en:
+  layouts:
+    phase_banner:
+      phase: BETA
```
patrick-laa, line 5: Likewise the new v5 convention is for `Beta` rather than `BETA` for the phase banner.

patrick-laa (approved):
Two tiny things that will crop up when you update to main, and one even tinier thing about `t` safeguards, but LGTM!

gems/laa_multi_step_forms/app/views/layouts/_phase_banner.html.erb
jsugarman: done, good spot

PR #459, opened 2024-01-24
Crm457 957 reuse defendants

patrick-laa:
## Description of change
Made `defendants` table polymorphic
Replaced full_name with first_name and last_name and amended NSM flow to use two fields
Updated PriorAuthority to use a defendants object rather than store client details on the main record

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRM457-957

## Notes for reviewer
To keep compatibility, the message builder is sending defendant full name to the app store as a single field. If we want to change that we'll need to update the caseworker app to cope with two separate fields

### DIFF
--- a/app/models/defendant.rb
+++ b/app/models/defendant.rb
@@ -1,7 +1,11 @@
 # NOTE: position value is used purely to maintain insertion order
 # and is not used to number or identify the defendant.
 class Defendant < ApplicationRecord
-  belongs_to :claim
+  belongs_to :defendable, polymorphic: true
 
   validates :id, exclusion: { in: [Nsm::StartPage::NEW_RECORD] }
+
+  def full_name
+    [first_name, last_name].join(' ')
+  end
 end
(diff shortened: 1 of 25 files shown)


imtiazAhmd (approved):
LGTM

app/models/defendant.rb
```diff
@@ -1,7 +1,11 @@
 # NOTE: position value is used purely to maintain insertion order
 # and is not used to number or identify the defendant.
 class Defendant < ApplicationRecord
-  belongs_to :claim
+  belongs_to :defendable, polymorphic: true
 
   validates :id, exclusion: { in: [Nsm::StartPage::NEW_RECORD] }
+
+  def full_name
```
jogilder, line 8: Is there a need to maintain full_name here?

jogilder: Ah, for compatibility with caseworker app - can we record this as tech debt and have another ticket to remove full_name from caseworker and this model if agreed by product?

PR #463, opened 2024-01-25
Crm457 955 feature flags

ivanELEC:
## Description of change

- Add feature flag library from caseworker app (use feature_flags.yml as config file instead of settings.yml)
- Use feature flag lib for omniauth_test_mode

## Link to relevant ticket
[Formalise Use of Feature Flags](https://dsdmoj.atlassian.net/browse/CRM457-955)

### DIFF
--- a/config/feature_flags.yml
+++ b/config/feature_flags.yml
@@ -1,6 +1,6 @@
 feature_flags:
   omniauth_test_mode:
-    local: true
+    local: <%= ENV.fetch("OMNIAUTH_TEST_MODE", false) %>
     development: true
     uat: false
     production: false # must never be enabled in the live service
--- a/config/feature_flags.yml
+++ b/config/feature_flags.yml
@@ -1,0 +1,6 @@
+feature_flags:
+  omniauth_test_mode:
+    local: true
+    development: true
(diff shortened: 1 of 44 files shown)


config/feature_flags.yml
```diff
@@ -0,0 +1,6 @@
+feature_flags:
+  omniauth_test_mode:
+    local: true
```
patrick-laa, line 3: The other app does something handy where for local environments the value gets pulled from an env var - while in K8s-land we want to avoid env vars, having an immediate mechanism to switch features on and off without having to touch a version-controlled file we don't want to accidentally update for everyone is quite helpful.

ivanELEC: yup agreed, will add this in shortly :)

patrick-laa: I think we're missing the final step here - this file should check the env var that you've added back in and use it to set the `local` value - otherwise the env var will just be ignored.

ivanELEC: my bad, I have this change locally and hadn't saved the file

patrick-laa (approved):
LGTM

PR #464, opened 2024-01-25
CRM457-973: Update button text and wording validation

patrick-laa:
## Description of change
Update wording of validation messages as per the error spreadsheet
Update UI of new application button as per the prototype

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRM457-973

### DIFF
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -129,8 +129,8 @@ en:
         prior_authority/steps/ufn_form:
           attributes:
             ufn:
-              blank: Enter the unique reference number
-              invalid: Unique reference number must be in the correct format, for example 310224/001
+              blank: Enter the unique file number
+              invalid: Unique file number must be in the correct format, for example 310224/001
         prior_authority/steps/youth_court_form:
           attributes:
             youth_court:
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -129,8 +129,8 @@ en:
         prior_authority/steps/ufn_form:
           attributes:
             ufn:
-              blank: Enter your unique file number for this application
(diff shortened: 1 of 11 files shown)


spec/forms/prior_authority/steps/ufn_form_spec.rb
```diff
@@ -25,7 +25,7 @@
       it 'has a validation error on the field' do
         expect(form).not_to be_valid
         expect(form.errors.of_kind?(:ufn, :blank)).to be(true)
-        expect(form.errors.messages[:ufn]).to include('Enter your unique file number for this application')
+        expect(form.errors.messages[:ufn]).to include('Enter the unique reference number')
```
jsugarman, line 28: i would  question some of these changes. this one in particular struck me as odd when ii saw it on the spreadsheet as it is referring to the UFN using two differently terminologies, which seems unnecessarily confusing.

patrick-laa: Good point! I've checked with Chris and he confirmed we should standardise on 'file' instead of 'reference'. I've updated.

config/locales/en/errors.yml
```diff
@@ -98,8 +98,8 @@ en:
         prior_authority/steps/ufn_form:
           attributes:
             ufn:
-              blank: Enter your unique file number for this application
-              invalid: UFN must be a 9-digit number
+              blank: Enter the unique file number
+              invalid: Unique file number must be in the correct format, for example 310224/001
```
jsugarman, line 102: this one i also find odd. We have already have a hint on the field "For example 310224/001" so that seems redundant, making it unwieldily long.

jsugarman (approved):
one other i would question but ✅  regardless

PR #474, opened 2024-01-29
Crm457 988/why pa

dwhenry:
## Description of change
Add why Prior Authority page with with JS drag and drop

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRM457-988

Screenshot 2024-01-29 at 12 35 19


ivanELEC:
Decision tree may need updating

app/models/supporting_document.rb
```diff
@@ -0,0 +1,8 @@
+class SupportingDocument < ApplicationRecord
+  SUPPORTING_DOCUMENT = 'supporting_document'.freeze
+  belongs_to :documentable, polymorphic: true
+
+  has_one_attached :file
```
patrick-laa, line 5: This isn't actually being used - you can delete this line and the only thing that fails is the test setup in one controller spec, because we're not actually using the Active Storage API.

app/views/prior_authority/steps/reason_why/edit.html.erb
```diff
@@ -0,0 +1,56 @@
...
+    <span class="govuk-caption-xl"><%= t(".caption") %></span>
+    <%= step_form @form_object do |form| %>
+      <%= form.govuk_text_area :reason_why,  max_chars: 2000, rows: 16, label: { size: 'xl', text: t(".page_title") }, hint: { text: t('.reason_why_hint_html')}  %>
+
+      <h1 class="govuk-heading-m"><%= t(".supporting_documentation") %></h1>
+      <span class="govuk-caption-m govuk-!-margin-bottom-8"><%= t(".supporting_documentation_caption") %></span>
+
+      <div class="moj-multi-file-upload">
+        <div class="moj-multi-file-upload__upload">
+          <%= form.govuk_file_field :documents, label: { text: t('.file_upload.heading'), size: 'm' }, class: 'moj-multi-file-upload__input', multiple: true, accept: "application/msword,application/vnd.openxmlformats-officedocument.wordprocessingml.document,application/rtf,image/jpeg,image/bmp,image/png,image/tiff,application/pdf" %>
```
patrick-laa, line 18: Is there any way of reusing ReasonWhyController::SUPPORTED_FILE_TYPES here? I feel like at some point otherwise we'll scratch our heads why a particular file type looks like it should be supported but isn't when the list changes.

spec/system/prior_authority/reason_why_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    fill_in 'Why is prior authority required?', with: 'important reasons'
+
+    click_on 'Save and continue'
+
+    expect(application.reload).to have_attributes(
+      reason_why: 'important reasons'
+    )
+  end
+
+  # TODO: implement test 'can have a reason with attachments'
```
patrick-laa, line 24: Love this comment - really clear and concise explaining what needs doing and why it can't be done yet.

app/views/prior_authority/steps/reason_why/edit.html.erb
```diff
@@ -0,0 +1,56 @@
...
+
+<% title t('.page_title') %>
+<% decision_step_header %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary(@form_object) %>
+    <span class="govuk-caption-xl"><%= t(".caption") %></span>
+    <%= step_form @form_object do |form| %>
+      <%= form.govuk_text_area :reason_why,  max_chars: 2000, rows: 16, label: { size: 'xl', text: t(".page_title") }, hint: { text: t('.reason_why_hint_html')}  %>
```
patrick-laa, line 11: Complex field hints are an absolute pain and I've not yet found a nice way of handling them. Putting HTML in the I18n yml is a decent enough workaround. One other thing you can do is:

```
hint: -> { content_for(:reason_why_hint) })
```

And then you can keep the HTML in your view by adding at the top:

```
<% content_for(:reason_why_hint) do %>
  <p>blah blah blah</p>
<% end %>
```
If you don't feel like that's any nicer then feel free to stick with what you've done.

patrick-laa (approved):
Looks good to me. A couple of suggestions but no blockers.

app/views/prior_authority/steps/reason_why/edit.html.erb
dwhenry: I want to have a chat about translations as we did it a different way for the crm7 work which was slightly nicer/easier.. will bring this up on our Thursday meeting.

app/models/supporting_document.rb
dwhenry: Not against removing this, but started and felt it was out of scope for this PR so have left it for now.

app/views/prior_authority/steps/reason_why/edit.html.erb
dwhenry: extraced to a common PORO so easier to use in required places

PR #475, opened 2024-01-29
CRM457-1017: Additional cost screens

patrick-laa:
## Description of change
Add the additional cost screens

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRM457-1017

## Notes for reviewer
- The 'add another' pattern of NSM is quite unwieldy, requiring repeated checks of the ID against a 'fake' one of 0000-0000-000000. The built-in Rails `resources` paradigm basically does what we wanted, so I chose that for this.
- That did mean making a couple of changes, mainly changing `:id` to `:application_id` in the URL structure, which had a couple of knock-on effects, but nothing significant - it turns out a lot of the code that set this ID explicitly was covering up some stuff that Rails inferred automatically
- The wiring up of these screens to the rest of the flow is incomplete, because the screens these connect to don't exist yet.
- Speaking of which, further changes are brewing. I'll keep an eye on the Slack for the next couple of days to see where they land. At least one of these screens might be going away.


app/forms/prior_authority/steps/additional_costs/detail_form.rb
```diff
@@ -0,0 +1,70 @@
...
+        validates :cost_per_item, presence: true, numericality: { greater_than: 0 }, if: :per_item?
+
+        validates :period, presence: true, time_period: true, unless: :per_item?
+        validates :cost_per_hour, presence: true, numericality: { greater_than: 0 }, unless: :per_item?
+
+        def formatted_total_cost
+          NumberTo.pounds(per_item? ? item_cost : time_cost)
+        end
+
+        def method
```
jsugarman, line 34: slight concern over the name `method` for a method. from a grepability point of view and name clash avoidance point of view perhaps `form_method`,`request_method` or `method_name`

jsugarman (commented):
Just a naming nit, see comment, but otherwise LGTM 👍

patrick-laa:
@jsugarman thanks for that, I've amended as you suggested. After a fun couple of hours of rebasing I've now got this sitting properly on top of all the other primary quote stuff. If you wouldn't mind a re-review that would be great.

gems/laa_multi_step_forms/Gemfile.lock
```diff
@@ -113,7 +113,7 @@ GEM
       html-attributes-utils (~> 1.0.0, >= 1.0.0)
       pagy (~> 6.0)
       view_component (>= 3.9, < 3.10)
-    govuk_design_system_formbuilder (4.1.1)
+    govuk_design_system_formbuilder (5.0.0)
```
jsugarman, line 116: eek, missed that when i bumped this gem for the app some time ago. I will let QA determine any impact. It cause no issues before....

app/controllers/prior_authority/steps/additional_costs_controller.rb
```diff
@@ -0,0 +1,25 @@
+module PriorAuthority
+  module Steps
+    class AdditionalCostsController < BaseController
+      def edit
+        @form_object = AdditionalCosts::OverviewForm.build(
+          current_application
+        )
+        # We never want 'yes' to be pre-selected as it makes for confusing UX.
+        # 'No' being pre-selected is fine otherwise users have to re-enter the same information
+        # every time they come back to this screen.
```
jsugarman, line 10: if there is a test that locks this behaviour down I would not bother with the comment but no bother either way

jsugarman (approved):
LGTM 👍

PR #476, opened 2024-01-29
Crm457 977 primary quote: service details

ivanELEC:
## Description of change

- Primary Quote added to task list
- Primary Quote form created
- Decision Tree Updated
- usages of accessible autocomplete updated to use autoselect option
- TODO: finish tests

## Link to relevant ticket
[CRM4 Provider - Primary: Service details](https://dsdmoj.atlassian.net/browse/CRM457-1015)


db/schema.rb
```diff
@@ -190,6 +190,9 @@
     t.boolean "psychiatric_liaison"
     t.string "psychiatric_liaison_reason_not"
     t.boolean "next_hearing"
+    t.boolean "hourly_travel_costs"
```
patrick-laa, line 193: This doesn't belong in this PR, I don't think - it's from the travel cost migration.

app/javascript/application.js
```diff
@@ -17,28 +17,50 @@ if ($inputs) {
...
+
+function convertSelectToAutocomplete(){
+  //enhance all tagged select elements to be accessible-autocomplete elements
+  const $acElements = document.querySelectorAll('[data-module="accessible-autocomplete"]')
+  if ($acElements) {
+    for (let i = 0; i < $acElements.length; i++) {
+      var convertedToAutocomplete = $acElements[i].getAttribute('data-converted')
+      if(!convertedToAutocomplete){
+        const name = $acElements[i].getAttribute('data-name')
+        const autoselect = trueFalseToBoolean($acElements[i].getAttribute('data-autoselect'))
```
patrick-laa, line 42: Adding a 10-line method twice feels a bit overkill when you could just do  `$acElements[i].getAttribute('data-autoselect') === true` here

app/forms/prior_authority/steps/primary_quote_form.rb
```diff
@@ -0,0 +1,31 @@
...
+        self.service_name = value
+      end
+
+      private
+
+      def persist!
+        record.update!(attributes.merge(attributes_to_reset))
+      end
+
+      def attributes_to_reset
```
patrick-laa, line 24: I think this  method name is wrong - we're not resetting anything, that's the terminology we use when a change to this form causes us to want to delete now-invalid attributes related to another form. Maybe 'default_attributes' would be a better name here.

db/schema.rb
```diff
@@ -215,6 +218,18 @@
     t.index ["auth_provider", "uid"], name: "index_providers_on_auth_provider_and_uid", unique: true
   end
 
+  create_table "quotes", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "service_name"
```
patrick-laa, line 222: I think it's a mistake to store the user-facing string value of service names in the database (at the moment if I choose "Plastic Surgeon" in the UI, the DB record will have the value "Plastic Surgeon", not "plastic_surgeon"). The reason being that there some business logic that we will need on subsequent screens where content/fields vary based on the service chosen by the user. We don't want to have to be doing string comparisons against user-facing strings because (a) if ever the wording gets updated we will have to update our business logic and. that will break all the old records and (b) even if we don't think the service needs to work in Welsh, the fact that our business logic would break in Welsh should be a sign we're on the wrong path.

I would strongly suggest that we have two fields here: `service_type` and `custom_service_name`. So then either service_type is something like "plastic_surgeon" and custom_service_name is nil, or service_type is "custom" and custom_service_name is whatever the user-entered string is.

app/presenters/prior_authority/tasks/primary_quote.rb
```diff
@@ -0,0 +1,16 @@
+module PriorAuthority
+  module Tasks
+    class PrimaryQuote < ::Tasks::Generic
+      PREVIOUS_TASK = CaseAndHearingDetail
```
patrick-laa, line 4: I know that CaseAndHearingDetail _is_ the previous task, but what this config actually does is say what has to be completed before PrimaryQuote can be started. And I don't think we actually need them to complete that to start this. If you look in PriorAuthority::Tasks::ClientDetail you'll see we've worked around this there by lying about the previous task but adding an explanatory comment. I wonder if it would be worth doing that here.

app/value_objects/prior_authority/quote_services.rb
```diff
@@ -0,0 +1,118 @@
+module PriorAuthority
+  # rubocop:disable Metrics/ClassLength
+  class QuoteServices < ValueObject
+    VALUES = [
+      ACCOUNTANT = new(:accountant),
+      ANAESTHATIST = new(:anaesthatist),
```
patrick-laa, line 6: Typo here and throughout - it's 'Anaesthetist', not Anaesthatist

config/locales/en/helpers.yml
```diff
@@ -365,4 +367,110 @@ en:
...
+      disability_consultant: Disability Consultant
+      dna_per_person: DNA (per person)
+      doctor_gp: Doctor (GP)
+      drug_expert: Drug Expert
+      employment_consultant: Employment Consultant
+      encro_chat: Encro Chat
+      engineer: Engineer
+      enquiry_agent: Enquiry Agent
+      ent_surgeon: ENT Surgeon
+      enviromental_expert: Enviromental expert
```
patrick-laa, line 397: Another typo - there's an n before the m in Environmental

config/locales/en/helpers.yml
```diff
@@ -365,4 +367,110 @@ en:
...
+    quote_services:
+      ae_consultant: A&E Consultant
+      accident_reconstruction: Accident Reconstruction
+      accountant: Accountant
+      anaesthatist: Anaesthatist
+      animal_behaviourist: Animal Behaviourist
+      architect: Architect
+      back_calculations: Back Calculations
+      benefit_expert: Benefit Expert
+      culture_expert: Culture expert
```
patrick-laa, line 379: There are some casing inconsistencies here - Culture expert doesn't have a capital E but Benefit Expert does. Is there an official list available? I suspect that the source you're working from was hand-transcribed at some point, as a few things like this have crept in.

config/locales/en/helpers.yml
```diff
@@ -365,4 +367,110 @@ en:
...
+      handwriting_expert: Handwriting Expert
+      human_trafficking: Human Trafficking
+      interpreters: Interpreters
+      intermediary: Intermediary
+      intermediary_after_apr_2022: Intermediary submitted after 1st April 2022
+      jewellery_valuation: Jewellery valuation
+      legal_prison_expert_extradition: Legal/Prison expert (extradition)
+      lip_reader_signer: Lip-Reader / Signer
+      mortgage_expert: Mortgage Expert
+      mediators_fees: Mediators Fees
```
patrick-laa, line 419: Suspect there should be a possessive apostrophe before the s in Mediator's.

config/locales/en/helpers.yml
```diff
@@ -365,4 +367,110 @@ en:
...
+      paediatrician: Paediatrician
+      pathologist: Pathologist
+      pharmacologist: Pharmacologist
+      photocopying_per_sheet: Photocopying per sheet
+      photographer: Photographer
+      physiotherapist: Physiotherapist
+      plastic_surgeon: Plastic Surgeon
+      process_server: Process Server
+      private_investigator: Private Investigator
+      psychiatrist_inc_chld: Psychiatrist (inc Chld)
```
patrick-laa, line 447: Missing I in child

spec/forms/prior_authority/steps/primary_quote_form_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    let(:record) { create(:quote, :blank, prior_authority_application: create(:prior_authority_application)) }
+
+    context 'with valid quote details' do
+      let(:service_name) { 'Forensics Expert' }
+      let(:contact_full_name) { 'Joe Bloggs' }
+      let(:organisation) { 'LAA' }
+      let(:postcode) { 'CR0 1RE' }
+
+      # rubocop:disable RSpec/ExampleLength
```
patrick-laa, line 76: To me, this is a sign instead of disabling the cop inline we should be configuring rubocop to allow examples of this length. If we think this is a reasonable length for an example (which I do), then it's pointless to have a rule telling us off for example of this length that we have to keep ignoring. If we don't think this is a reasonable length for an example, we should change this example to make it more concise.

spec/system/nsm/hearing_details/suggestion_autocomplete_spec.rb
```diff
@@ -14,12 +14,12 @@
     click_on 'Save and continue'
 
     fill_in 'Which court was the last case hearing heard at?', with: 'Aldershot'
-    expect(page).to have_field('Which court was the last case hearing heard at?', with: "Aldershot Magistrates' Court")
+    expect(page).to have_field('Which court was the last case hearing heard at?', with: 'Aldershot')
```
patrick-laa, line 17: Changing the expectation like this changes the entire test. The test says 'can select a value from the autocomplete', but what this test is describing is _not_ selecting a value from the autocomplete, but instead entering a custom value. I think this spec should now be 2 specs, one showing that if you enter a custom value, by default it remains a custom value. The other spec should show how you actually use autocomplete to get from entering 'Aldershot' to clicking on 'Aldershot Magistrates Court' to the value being recorded being "Aldershot Magistrates Court".

spec/system/support/prior_authority/step_helpers.rb
```diff
@@ -2,6 +2,7 @@
 
 module PriorAuthority
   module StepHelpers
+    # rubocop:disable Metrics/CyclomaticComplexity
```
patrick-laa, line 5: (For comparison, I'd be more inclined to let this one slide. In general was want to keep cyclomatic complexity down. In this _specific_ case, unusually, we're ok with it, because while it might be "cyclomatically" complex, it's actually very simple, and very _obviously_ simple.)

patrick-laa (commented):
A few thoughts on this. The only really significant one is that I think how we store service types needs a little rethink, otherwise we'll end up with a lot of fragility in the system.

app/javascript/application.js
ivanELEC: rails stores all data attributes as strings - but yeah I think the method I did is very inelegant, I'll do your suggestion (with the minor change that the conditional statement compares against the string "true" instead of a boolean)

db/schema.rb
ivanELEC: yeah I think this is a good suggestion - this extra logic flew past my radar

config/locales/en/helpers.yml
ivanELEC: the list comes from here: https://docs.google.com/spreadsheets/d/1xtepqupYvQpb79CgCyDE-8G90Ol4xvXVtKb1gh08f0A/edit#gid=0

I'll talk to Aimee to check if there's a cleaned up official version

app/presenters/prior_authority/tasks/primary_quote.rb
ivanELEC: yeah I think this is necessary context - before I do this I'm going to look into whether maybe there's actually an explicit can_start? method that I could put together because thinking about it, when the flow is on "non prison law", there are other forms that need completing before we can get to primary quote I think  like youth court/psychiatric liason (I'll double check in ticket and with Mary)

ivanELEC: yeah after talking with Mary, the "last step" of this section depends on whether Prison Law is Yes or No so I'm going to make an explicit can_start? override instead of relying on PREVIOUS_TAST

PR #477, opened 2024-01-30
CRM457-982, 1014: Find and delete an application

patrick-laa:
## Description of change
- Added draft/submitted/assessed tabs
- Added server-side sorting to the contents of each tab
- Added pagination

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRM457-982

## Notes for reviewer
I chose not to DRY up the views because they're just different enough that I thought it was less messy than trying to squidge them into a one-size-fits-all template.

I've not applied the server-side sorting to the NSM code. I think we agreed that would be the subject of a separate ticket.

Screenshot 2024-01-30 at 13 17 52

From the console, do `require 'factory_bot_rails'` and then you can do things like `20.times { FactoryBot.create :prior_authority_application, updated_at: _1.days.ago, provider: Provider.first, status: 'draft' }` to seed some data


app/helpers/sortable_table_helper.rb
```diff
@@ -0,0 +1,19 @@
...
+      reorder_form(path, column, next_direction, i18n_stem, index)
+    end
+  end
+
+  def reorder_form(path, column, next_direction, i18n_stem, index)
+    tag.form(action: path, method: 'get') do
+      safe_join([tag.input(type: 'hidden', name: 'sort_by', value: column),
+                 tag.input(type: 'hidden', name: 'sort_direction', value: next_direction),
+                 tag.button(type: 'submit', 'data-index': index) do
+                   I18n.t("#{i18n_stem}.#{column}")
```
jsugarman, line 15: i have been bitten by missing translation that can result from dynamic lookups likes this, however I get that they make things so much simpler so i guess we just go this route (that I have seen on so many other apps)

You can use a i18n scope arg here, although that does not deal with the dynamic lookup aspect

```ruby
I18n.t(column, scope: i18n_stem)
```
https://guides.rubyonrails.org/i18n.html#basic-lookup-scopes-and-nested-keys

jsugarman (approved):
minor comment and suggestion, but ✅ regardless

ivanELEC:
lgtm will review after merge conflicts resolved

PR #479, opened 2024-01-30
Crm457 933/user auth

dwhenry:
## Description of change
Extend Gatekeeper to support separate enabling of crm4/crm7

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRM457-933

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,17 +6,19 @@ class ApplicationController < LaaMultiStepForms::ApplicationController
   before_action :set_default_cookies
   before_action :can_access_service
 
-  def cas_access_service
-    return if Provider::Gatekeeper.provider_enrolled?(service:)
+  private
 
-    Rails.logger.warn "Not enrolled provider access attempt, UID: #{auth_hash.uid}, " \
-                      "accounts: #{auth_hash.info.office_codes}, service: #{service}"
+  def can_access_service
+    return unless current_provider
+    return if Providers::Gatekeeper.new(current_provider).provider_enrolled?(service:)
 
-    redirect_to laa_msf.not_enrolled_errors_path, flash: { notice: 'You do not have access to that srevice'}
+    Rails.logger.warn "Not enrolled provider access attempt, UID: #{current_provider.id}, " \
+                      "accounts: #{current_provider.office_codes}, service: #{service}"
 
+    redirect_to root_path
   end
 
   def service
-    Provider::Gatekeeper::ANY
+    Providers::Gatekeeper::ANY_SERVICE
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,4 +4,19 @@ class ApplicationController < LaaMultiStepForms::ApplicationController
   include CookieConcern
 
   before_action :set_default_cookies
+  before_action :can_access_service
+
+  def cas_access_service
+    return if Provider::Gatekeeper.provider_enrolled?(service:)
+
+    Rails.logger.warn "Not enrolled provider access attempt, UID: #{auth_hash.uid}, " \
+                      "accounts: #{auth_hash.info.office_codes}, service: #{service}"
+
+    redirect_to laa_msf.not_enrolled_errors_path, flash: { notice: 'You do not have access to that srevice'}
+
+  end
+
+  def service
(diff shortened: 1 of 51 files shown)


app/controllers/prior_authority/steps/base_controller.rb
```diff
@@ -1,14 +1,8 @@
 module PriorAuthority
   module Steps
     class BaseController < ::Steps::BaseStepController
-      include ApplicationHelper
-      include CookieConcern
-
```
dwhenry, line 6: This is already included in base classes

app/controllers/application_controller.rb
```diff
@@ -4,4 +4,21 @@ class ApplicationController < LaaMultiStepForms::ApplicationController
   include CookieConcern
 
   before_action :set_default_cookies
+  before_action :can_access_service
+
+  private
+
+  def can_access_service
+    return unless current_provider
+    return if Providers::Gatekeeper.new(current_provider).provider_enrolled?(service:)
```
patrick-laa, line 13: It makes me a bit nervous that Gatekeeper expects to be initialised with an `auth_info` object, and we're passing in a provider here. It happens to be the case that both objects have an `office_codes` array attribute, but I think it might be worth tweaking the OmniauthCallbacksController so that it sends the provider instance to Gatekeeper, so we always know what sort of thing we're dealing with.

config/initializers/omniauth.rb
```diff
@@ -16,7 +27,8 @@
     # * info[office_codes] => ["1A123B", "2A555X"]
     if FeatureFlags.omniauth_test_mode.enabled?
       config.before_callback_phase do |env|
-        env['omniauth.auth'].merge!(
+        AuthMemory.put(env['omniauth.auth'].dup)
+        env['omniauth.auth'] = AuthMemory.get.merge(
```
patrick-laa, line 31: I'm. a bit confused by this. What's the benefit of this compared to simply doing:

```
env['omniauth.auth'] = env['omniauth.auth'].dup.merge(
```
?

spec/steps/nsm/provider/select_controller_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe Nsm::Steps::Office::SelectController, type: :controller do
-  let(:provider) { Provider.new }
+  let(:provider) { build_stubbed(:provider) }
```
patrick-laa, line 4: Huh. I didn't know about `build_stubbed`. That's pretty cool.

patrick-laa (commented):
Couple of queries. As always, feel free to tell me I've missed the point!

config/initializers/omniauth.rb
dwhenry: So this is a fun fact about how the test mode appears to work as it caches what is set in `env['omniauth.auth']` between requests (at least in the tests)... hence I am storing the initial value in memory to avoid this glitch...

app/controllers/application_controller.rb
dwhenry: We chat separately about this... 

The issue I see (might not be an issue) is that this would create the provider records even if they are not able to use the system....

config/initializers/omniauth.rb
patrick-laa: Weird. For me the tests pass with the change I suggested above, as all it needed was the `dup` - I still don't see how assigning the `dup`'d value to an attribute and taking it out again would change things. But if you're seeing a problem without that extra step then fair enough.

patrick-laa: I'm still intrigued by this. When I [remove AuthMemory](https://github.com/ministryofjustice/laa-claim-non-standard-magistrate-fee-backend/commit/10b5adf948122f233f49b6c9223e9340c1b97b35), both locally and [on CI](https://app.circleci.com/pipelines/github/ministryofjustice/laa-claim-non-standard-magistrate-fee-backend/1221/workflows/761282a1-5cdd-47ec-b47f-d0d2f036f64d/jobs/6777) everything passes. Are you sure it's necessary on your local machine? Are you working in Docker?

dwhenry: arrrg... assumptions... so `merge!` broke the tests so assumed setting the value would as well... but it doesn't... you don`t even need to dup....

patrick-laa (approved):

PR #483, opened 2024-01-31
Crm457 1061 add dynamic branching

jogilder:
## Description of change

Add in creation of a dynamic url and pods to dev environment

## Link to relevant ticket

https://dsdmoj.atlassian.net/browse/CRM457-1061

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -256,7 +256,7 @@ jobs:
             ${AWS_ECR_REGISTRY_ID}.dkr.ecr.${ECR_REGION}.amazonaws.com
       - aws-ecr/build-image:
           push-image: true
-          tag: $ECR_TAG
+          tag: $IMAGE_TAG
           region: $ECR_REGION
           repo: $ECR_REPOSITORY
           extra-build-args: |
@@ -292,7 +292,7 @@ jobs:
             kubectl config set-context ${K8S_CLUSTER_NAME} --cluster=${K8S_CLUSTER_NAME} --user=circleci --namespace=${K8S_NAMESPACE}
             kubectl config use-context ${K8S_CLUSTER_NAME}
             kubectl --namespace=${K8S_NAMESPACE} get pods
-            source .circleci/define_build_environment_variables << parameters.namespace >> << parameters.dynamic_hostname >>
+            source .circleci/define_build_environment_variables << parameters.dynamic_hostname >>
             echo "export IMAGE_TAG=$IMAGE_TAG" >> $BASH_ENV
             echo "export RELEASE_HOST=$RELEASE_HOST" >> $BASH_ENV
       - run:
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -341,40 +332,48 @@ workflows:
...
-      - hold-dev-crm4:
-          type: approval
-          requires:
-            - build-to-ecr
-      - deploy-dev-crm4:
+#      - hold-dev-crm4:
+#          type: approval
+#          requires:
+#            - build-to-ecr
+      - deploy-to-environment:
```
patrick-laa, line 360: I think part of the benefit of dynamic releases is we don't need a "dev-crm4" environment at all - it was a sticking plaster put in place to deal with not having enough dev environments, so I'm hoping we can delete every reference to it in this PR.

.circleci/config.yml
```diff
@@ -294,29 +265,49 @@ jobs:
...
     steps:
-      - deploy-to-environment:
-          environment: prod
+      - checkout
+      - run:
+          name: Authenticate with cluster
+          command: |
+            .circleci/authenticate_with_kubernetes_cluster
+      - run:
+          name: Delete dynamic release
```
patrick-laa, line 308: The difficulty with trying to trigger this in CI is that [CircleCI doesn't have a trigger for branch delete or PR close events](https://discuss.circleci.com/t/github-pr-close-or-branch-delete-events/27062), meaning that is a PR is closed without merging it, its dynamic cloud platform environment won't be deleted. I _think_ this means we need to be triggering this via a GitHub action.

helm_deploy/values-dev.yaml
```diff
@@ -78,4 +70,4 @@ secrets:
   appStore: https://laa-crime-application-store-dev.apps.live.cloud-platform.service.justice.gov.uk
 
 service_account:
-  name: laa-claim-non-standard-magistrate-fee-dev-irsa
\ No newline at end of file
+  name: laa-claim-non-standard-magistrate-fee-dev-irsa
```
patrick-laa, line 73: I think we need a way of providing dynamic dev environments with their own DB. Otherwise a dodgy migration in one PR will break every other environment. In CCQ we used a postgres pod with a persistent volume claim. I guess you could also use RDS, creating a new database with a custom name inside the dev RDS instance for each branch and then deleting it on cleanup?

patrick-laa (commented):
These are mostly notes to self to help me understand this all.

.circleci/config.yml
jogilder: @patrick-laa makes sense - I think we could go with the github action.

.circleci/config.yml
jogilder: I also agree with this - I was using it as a testing ground instead of messing up everyone's dev. Eventually it would go.

patrick-laa:
CLosing as superseded by #568

PR #485, opened 2024-02-01
CRM457-1067: Accessibility test

patrick-laa:
## Description of change
Add Axe accessibility tests that can be run on CI but skipped locally

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRM457-1067

```
INCLUDE_ACCESSIBILITY_SPECS=1 bundle exec rspec spec/system/accessibility_spec.rb
```


spec/rails_helper.rb
```diff
@@ -20,6 +21,12 @@
   exit 1
 end
 
+Capybara.register_driver :headless_chrome do |app|
+  options = Selenium::WebDriver::Chrome::Options.new
+  options.add_argument('--headless=new')
+  Capybara::Selenium::Driver.new(app, browser: :chrome, options: options)
+end
+
```
dwhenry, line 29: so we already have cubrite setup `spec/system/support/cuprite_setup.rb`

I happy to switch to selenium but would prefer not to have both.. :)

dwhenry (commented):
Just the comment on the driver..

also would be nice to have `INCLUDE_ACCESSIBILITY_SPECS` in the `.env` files... might require the check to be `ENV['INCLUDE_ACCESSIBILITY_SPECS'] == 'true'` to make that work.

spec/rails_helper.rb
patrick-laa: I hadn't even spotted cuprite! It doesn't play nicely with axe, so I've removed it, using selenium's headless chrome as the replacement. It has the advantage that it supports capybara's `drop` so we're a step closer to adding that missing spec in `reason_why_spec.rb` (I started to build it but don't understand enough of the infrastructure so I've left a TODO in place to finish it off).

jsugarman (approved):
👍 having dynamic generated tests may become painful to debug but lets suck it and see.

ivanELEC (approved):

PR #488, opened 2024-02-02
CRM457-1015: Add Upload to Primary Quote

ivanELEC:
## Description of change

- Added max upload size for single file uploads as env var (since this doesn't use the MOJ multi file upload JS)
- Added file upload element to primary quote form including error handling 
 - NOTE: the errors for file uploads are shown as alert flashes, I have done it this way due the fact that file uploads are handled in the controller, not the base form object and not sure of the best way to pass errors into the form after a redirect other than showing the issue via a flash - we may need to enhance this code so that the submission and error handling are consistent with other parts of the app
 - Update Docker Image from ruby:3.2.2-alpine3.18 to ruby:3.2.2-alpine3.19 to fix [high vulnerability](https://security.snyk.io/vuln/SNYK-ALPINE318-LIBXML2-6245694)

## Link to relevant ticket
[Primary Quote - Service Details
](https://dsdmoj.atlassian.net/browse/CRM457-1015)## Notes for reviewer


app/models/supporting_document.rb
```diff
@@ -1,8 +1,11 @@
 class SupportingDocument < ApplicationRecord
   SUPPORTING_DOCUMENT = 'supporting_document'.freeze
+  QUOTE_DOCUMENT = 'quote_document'.freeze
+
   belongs_to :documentable, polymorphic: true
 
   has_one_attached :file
 
   scope :supporting_documents, -> { where(document_type: SUPPORTING_DOCUMENT) }
+  scope :quote_ocument, -> { where(document_type: QUOTE_DOCUMENT) }
```
patrick-laa, line 10: Typo - `ocument` - which suggests to me we're not using this scope

app/models/quote.rb
```diff
@@ -1,3 +1,10 @@
 class Quote < ApplicationRecord
   belongs_to :prior_authority_application
+  has_one :document, lambda {
```
patrick-laa, line 3: We'll need to update `QuotePayloadBuilder::PriorAuthority::SubmitToAppStore` to add this new stuff to the payload - we're trying to keep this up to date as we go.

app/views/prior_authority/steps/primary_quote/edit.html.erb
```diff
@@ -9,12 +9,32 @@
...
+                            options: { include_blank: true }, label: { size: "m" }, hint: { size: "s", text: t(".service_name_hint")},
                             data: { autoselect: false } %>
       <h2 class="govuk-heading-m"><%= t(".service_details") %></h2>
-      <%= form.govuk_text_field :contact_full_name, label: { text: t(".contact_full_name"), size: 's' } %>
-      <%= form.govuk_text_field :organisation, label: { text: t(".organisation"), size: 's' } %>
-      <%= form.govuk_text_field :postcode, label: { text: t(".postcode"), size: 's' } %>
+      <%= form.govuk_text_field :contact_full_name, label: { text: t(".contact_full_name"), size: "s" } %>
+      <%= form.govuk_text_field :organisation, label: { text: t(".organisation"), size: "s" } %>
+      <%= form.govuk_text_field :postcode, label: { text: t(".postcode"), size: "s" } %>
+      <%= form.govuk_file_field :document, label: { text: t(".file_upload"), size: "m" }, hint: { text: t(".file_upload_hint") }, accept: SupportedFileTypes::QUOTE_DOCUMENT.join(',') %>
```
patrick-laa, line 18: You'll also need to add this file to app/views/prior_authority/steps/primary_quote_summary/_service_cost_card.html.erb

spec/system/prior_authority/primary_quote_spec.rb
```diff
@@ -42,4 +42,5 @@
     click_on 'Save and come back later'
     expect(page).to have_content 'Your applications'
   end
+  # TODO: Add system tests for uploading files
```
patrick-laa, line 45: Joel had a great suggestion for TODOs - it would be good to get into the discipline of only adding a TODO if we can also add a ticket number for the ticket that will address the TODO. That way it's harder for them to linger indefinitely.

patrick-laa (commented):
Looking promising - this stuff is really fiddly, isn't it? I think we're mostly there.

app/controllers/prior_authority/steps/primary_quote_controller.rb
```diff
@@ -1,16 +1,31 @@
...
           application: current_application
         )
+
+        @primary_quote_document = current_application.primary_quote.document
       end
 
       def update
         record = primary_quote
+        @primary_quote_document = current_application.primary_quote.document
+        if params[:prior_authority_steps_primary_quote_form][:document]
```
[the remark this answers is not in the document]
ivanELEC: Ah yes, good point, I originally did this thinking the upload wasn't compulsory and forgot to check back to confirm if  this was the case or not - will add this in

app/models/quote.rb
ivanELEC: added

app/controllers/prior_authority/steps/primary_quote_controller.rb
```diff
@@ -1,16 +1,31 @@
 module PriorAuthority
   module Steps
     class PrimaryQuoteController < BaseController
+      skip_before_action :verify_authenticity_token, only: [:update]
```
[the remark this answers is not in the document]
ivanELEC: made the mistake of using the multifile upload code as a starting point of a single file upload and working backwards, there were a few things I forgot to take out

spec/services/file_upload/file_uploader_spec.rb
```diff
@@ -10,12 +12,77 @@
 
     context 'production environment' do
       before do
-        allow(Rails).to receive(:env).and_return(ActiveSupport::StringInquirer.new('production'))
+        allow(Rails).to receive_message_chain(:env, :production?).and_return(true)
```
patrick-laa, line 15: I really don't like this... Not your change, but rather the whole context here. Using Rails.env.production is a bit iffy in general, as it rather goes against the general 12-factor approach that an app shouldn't know its own environment. And then _overriding_ that... Yeesh. If we want to sometimes use AwsFileUploader in the test environment, then the logic to decide whether to use AwsFileUploader shouldn't be controller with Rails.env.production.

But this might be a battle for another day...

spec/system/prior_authority/primary_quote_summary_spec.rb
```diff
@@ -1,13 +1,14 @@
 require 'system_helper'
 
-RSpec.describe 'Prior authority applications - add service costs' do
+RSpec.describe 'Prior authority applications - add service costs', :javascript, type: :system do
```
patrick-laa, line 3: I'm really not a fan of making every system spec a JS spec. It slows the test suite down a lot. Can we not use `attach_file('file-field-selector', File.join(Rails.root, '/spec/fixtures/blah'))` to use the basic field attachment field that should be there when JS isn't available?

patrick-laa (commented):
Some tiny nitpicks. The only thing I really want to query is whether our system tests _have_ to have JS enabled in order to be able to upload a file on the primary quote screen?

If we really can't do that in a non-JS way, then I think to get to the state needed for post-primary-quote screens we should save a Factory-bot created application in the database  rather than building it up screen by screen. See the `reason_why_spec` for an example. Otherwise we slow our test suite down a lot.

config/routes.rb
```diff
@@ -124,12 +124,12 @@
         edit_step :ufn
         edit_step :case_contact
         edit_step :client_detail
+        upload_step :primary_quote
```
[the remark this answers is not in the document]
ivanELEC: yeah you're right - I was keeping it in there in case we needed the ability to delete the uploaded file but the file is compulsory so agreed with Mary it should just be an overwrite

spec/system/prior_authority/primary_quote_summary_spec.rb
ivanELEC: The issue isn't the file upload, the issue is the suggestion box for the services which uses JS - it starts off as a completely different element which is a select element and is then converted to the searchable input field after the js runs - js needs to be enabled on every test which has going through the primary quote form in the before block (once we have the offences list in we will need to do this for that form too)

patrick-laa: @ivanELEC ah but we had that solved - because without JS that field is just a select, which is why the step helper selected a value from the select. Can't we just keep doing that so that we don't slow down the test suite?

db/schema.rb
```diff
@@ -204,12 +204,13 @@
     t.boolean "psychiatric_liaison"
     t.string "psychiatric_liaison_reason_not"
     t.boolean "next_hearing"
+    t.integer "travel_time"
```
patrick-laa, line 207: I think something has gone a bit wrong with your rebase - we don't want to be reverting this change, and especially not by simply deleting the migration files, as that will break every environment where those migrations have already been run.

gems/laa_multi_step_forms/app/controllers/steps/base_step_controller.rb
```diff
@@ -40,8 +37,6 @@ def update_and_advance(form_class, opts = {})
         # Validations will not be run when saving a draft
         @form_object.save!
         redirect_to opts[:after_commit_redirect_path] || nsm_after_commit_path(id: current_application.id)
-      elsif params.key?(:reload)
```
patrick-laa, line 43: I won't keep commenting on these throughout, but again - looks like you're accidentally undoing a bunch of work from previous tickets at the moment....

patrick-laa (approved):

PR #490, opened 2024-02-02
WIP - [CRM457-1048] better-task-list

dwhenry:
## Description of change
General improvements to the tasks list so that we can better support the requirements:

* Add TaskStatus class to ensure status is only calculated once per task to avoid nested calls
* Add option to set `PREVIOUS_TASKS` when more than 1 dependancy (`PREVIOUS_TASK` still works)
* some refactoring of code 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRM457-1048


gems/laa_multi_step_forms/app/presenters/tasks/base_task.rb
```diff
@@ -29,8 +30,12 @@ def default_url_options
     end
 
     def status
+      task_statuses.status(self.class, application:)
+    end
+
+    def current_status
       return TaskStatus::NOT_APPLICABLE if not_applicable?
-      return TaskStatus::UNREACHABLE unless can_start?
```
dwhenry, line 33: So this means that if a page is reachable at any? point it should still be reachable in the future even if required pages are now not fully complete.

app/presenters/prior_authority/tasks/case_and_hearing_detail.rb
```diff
@@ -11,28 +11,20 @@ def path
         end
       end
 
-      def completed?
-        results = forms_for_completion_hash.each_with_object([]) do |(_k, form_hash), arr|
-          arr << form_hash[:form].build(application).valid? if form_hash[:required]
-        end
-        results.all?(true)
+      def completed?(_rec = record, _form = nil)
+        required_forms.all? { |form| super(record, form) }
```
jsugarman, line 15: neat, but would it be clearer if the usable argument is used for consistency?

```rb
```
```suggestion
      def completed?(rec = record, _form = nil)
        required_forms.all? { |form| super(rec, form) }
```

app/presenters/tasks/generic.rb
```diff
@@ -24,6 +24,14 @@ def completed?(rec = record, form = associated_form)
 
     private
 
+    def previous_tasks
+      if self.class.const_defined?(:PREVIOUS_TASKS)
+        self.class::PREVIOUS_TASKS
+      else
+        [self.class::PREVIOUS_TASK]
+      end
+    end
```
jsugarman, line 33: not necessarily in this PR, but had been wanting to rename/pluralise all PREVIOUS_TASK to PREVIOUS_TASKS and enforce as always being an array, even if only one element. This could make this logic terser and more consistent.

jsugarman (commented):
I like the "previous task" refactor and suggest for future, or now, wider refactor to use`PREVIOUS_TASKS` for all.

One thing i would consider changing (use of rec arg in completed? method), but otherwise just some nits.

jsugarman (approved):
rebase means PREVIOUS_TASKS change has broken a few things but ✅

PR #501, opened 2024-02-06
CRM457-988 addendum: Test file upload

patrick-laa:
## Description of change
We had a TODO because the file upload mechanism was really hard to test using our old capybara driver. Now we've switched to Selenium we can actually do the test, so here's an attempt at it.

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-988)

## Notes for reviewer
I had a go at setting up Clamby both locally and on CI and got nowhere, so I've let myself off the hook and just bypassed it for this test. More than happy to commit to getting it running in test env in future but I think that may need to be a separate ticket.


jogilder (approved):
LGTM

PR #502, opened 2024-02-06
CRM457-1015: Give priority to text field rather than hidden select

patrick-laa:
## Description of change
The bug was that if you type a value into the service cost but don't click on a pre-selected option, the hidden field value doesn't update, and we were giving priority to the hidden field over the text field. Fixing this means that users aren't forced to click for a change to 'take'

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1015)

### DIFF
--- a/app/forms/prior_authority/steps/primary_quote_form.rb
+++ b/app/forms/prior_authority/steps/primary_quote_form.rb
@@ -13,8 +13,8 @@ module PriorAuthority
       validates :postcode, presence: true, uk_postcode: true
 
       def service_type_suggestion=(value)
-        if value.in?(translated_values)
-          self.service_type = service_type.value
+        if value.in?(translations.keys)
+          self.service_type = translations[value]
           self.custom_service_name = nil
         else
           self.service_type = 'custom' if value.present?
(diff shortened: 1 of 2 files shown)


app/forms/prior_authority/steps/primary_quote_form.rb
```diff
@@ -13,8 +13,8 @@ class PrimaryQuoteForm < ::Steps::BaseFormObject
       validates :postcode, presence: true, uk_postcode: true
 
       def service_type_suggestion=(value)
-        if value.in?(translated_values)
-          self.service_type = service_type.value
+        if value.in?(translations.keys)
+          self.service_type = translations[value]
```
dwhenry, line 17: I found this slightly confusing as expected the `value` to contact the untranslated key.... 

It looks like it is returning the translated text which it shouldn't be... ???

or am I missing something.

patrick-laa: @dwhenry the `value` of service_type_suggestion is the contents of the text field, which is the translated value. This is separate from the value of `service_type` which is the untranslated value, derived from the hidden HTML <select> element. The bug occurs when a user types in a value to the text field but doesn't click the autocomplete. In which case the service_type_suggestion contains a "true" translated value, but the service_type dropdown contains a false untranslated value, so the translated value needs to take precedence.

dwhenry (approved):

PR #503, opened 2024-02-06
CRM457-975: Contact detail onwards navigation

patrick-laa:
## Description of change
From QA:
- 'Save and come back later' should redirect to 'Your applications', not 'Your application progress' - this is a universal update
- 'Save and continue' from 'case contact' should redirect to 'Your application progress', not 'Client details'

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-975)

### DIFF
--- a/spec/system/support/prior_authority/step_helpers.rb
+++ b/spec/system/support/prior_authority/step_helpers.rb
@@ -64,6 +64,7 @@ module PriorAuthority
     end
 
     def fill_in_client_detail
+      click_on 'Client details'
       fill_in 'First name', with: 'John'
       fill_in 'Last name', with: 'Doe'
 
--- a/app/controllers/prior_authority/steps/base_controller.rb
+++ b/app/controllers/prior_authority/steps/base_controller.rb
@@ -10,7 +10,7 @@ module PriorAuthority
       end
 
       def after_commit_redirect_path
-        prior_authority_after_commit_path(id: current_application.id)
+        prior_authority_root_path(anchor: 'drafts')
       end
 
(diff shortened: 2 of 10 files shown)


spec/system/support/prior_authority/step_helpers.rb
```diff
@@ -64,6 +64,7 @@ def fill_in_case_contact
     end
 
     def fill_in_client_detail
+      click_on 'Client details'
```
jsugarman, line 67: i did not realise you were working on this and started a [branch this am to handle the flow generally](https://github.com/ministryofjustice/laa-claim-non-standard-magistrate-fee-backend/pull/507).

Having looked into this a bit I found that having the click_on in here made the individuals spec less flexible in testing the checks for tasklist status e.g. "Not started", plus just from a convention point of view I think the fill_in... methods should only fill_in the fields (not perform navigation (to avoid confusing our future selves. I therefore added the click_on  in the fill_in_until_step method.

patrick-laa: Good point about `click_on`. Ok, how shall we divvy this up? I'm working on this to get CRM457-975 to pass QA. Do we want a single branch that handles both:
- The new behaviour for 'Save and come back later'
- Standardising returning to the task list after completing a task?

patrick-laa: Ok @jsugarman I've updated to move the 'click_on's out of the 'fill_ins'

jsugarman (approved):
👍

PR #510, opened 2024-02-08
crm457-1109: Fix missing translations on CYA page

jsugarman:
## Description of change
Fix missing translations on CYA page

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1109)

The check your answers page for NSM encounters
multiple missing translation errors when running
in local/development. This error does not occur on
hosted environments because any missing translations fall back to using a humanized version of whatever “key” has been supplied.

Screenshot 2024-02-08 at 11 52 34

### DIFF
--- a/app/presenters/nsm/check_answers/disbursement_costs_card.rb
+++ b/app/presenters/nsm/check_answers/disbursement_costs_card.rb
@@ -30,7 +30,7 @@ module Nsm
       def disbursement_rows
         disbursements.rows.map do |row|
           {
-            head_key: row[:key][:text],
+            head_key: row[:key][:text].parameterize.underscore,
             text: row[:value][:text]
           }
         end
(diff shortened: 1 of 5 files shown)


app/presenters/nsm/check_answers/disbursement_costs_card.rb
```diff
@@ -30,7 +30,7 @@ def header_rows
       def disbursement_rows
         disbursements.rows.map do |row|
           {
-            head_key: row[:key][:text],
+            head_key: row[:key][:text].parameterize.underscore,
```
dwhenry, line 33: so `row[:key][:text]` is the translated value so not 100% sure converting it to look like a key is a good idea. Would mostly depend on where it is then used (which I am not sure of)

but this is likely to break what is shown on the some of the views

config/locales/en/nsm/check_answers.yml
```diff
@@ -82,6 +82,7 @@ en:
               total: Total
               total_inc_vat: Total (including VAT)
             disbursements:
+              car: Car
```
dwhenry, line 85: :( we need to be better at this.......

patrick-laa:
Closing as now fixed in another way

PR #511, opened 2024-02-08
CRM457-1104: Basic alternative quote flow

patrick-laa:
## Description of change

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1104)

## Notes for reviewer
The basic 'add another' flow is similar to additional costs on the primary quote, so I've abstracted the logic for that into a base controller

There is an open question as to what the summary UI will look like when there are no quotes but only because a quote has been deleted - the current designs don't quite work for that as they don't allow a 'reason for no quote' field. So for the time being I've gone with simple and consistent, but this may change.

Screenshot 2024-02-08 at 13 14 53


app/controllers/prior_authority/steps/add_another_controller.rb
```diff
@@ -0,0 +1,68 @@
+module PriorAuthority
+  module Steps
+    class AddAnotherController < BaseController
```
dwhenry, line 3: can we put this at the lower level so it can be used for NSM Add Another forms? 

is it worth looking at that for this PR or leave it to review later?

spec/system/prior_authority/alternative_quotes_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'system_helper'
+
+RSpec.describe 'Prior authority applications - alternative quote' do
+  context 'when I visit the quote summary screen' do
+    before do
+      fill_in_until_step(:primary_quote_summary)
```
dwhenry, line 6: my worry is that these tests are getting slower and slower as they build each pay up one at a time... any reason not to use a factory and just build a valid object for this start point... it would be much quicker?

dwhenry (approved):
✅ LGTM

app/controllers/prior_authority/steps/add_another_controller.rb
patrick-laa: Inheritance was making this complicated, so I've moved this into a concern so it can be reused. Until we have a full add-another flow working and signed off in PA I'm hesitant to try to rework NSM to use the same system, but once it's up and running I think it would be great to see if we use this for NSM.

spec/system/prior_authority/alternative_quotes_spec.rb
patrick-laa: Fair play - I've updated to use a factory instead.

PR #512, opened 2024-02-08
CRM457-1040: Fix bug in navigation stack logic

patrick-laa:
## Description of change

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1040)

## Notes for reviewer
Previously if you completed steps A and B then went back and edited step A, step B would be marked as 'not started'. Which would be extra problematic if step B came before step A in the task list.

As we have our own logic for determining where changes in A invalidate step B, the simplest thing is to simply remove the 'prune' mechanism that's causing the app to forget about previous interactions.

### DIFF
--- a/app/controllers/prior_authority/steps/base_controller.rb
+++ b/app/controllers/prior_authority/steps/base_controller.rb
@@ -22,6 +22,10 @@ module PriorAuthority
       def service
         Providers::Gatekeeper::PAA
       end
+
+      # The default behaviour of prune_navigation_stack assumes linear completion of tasks
+      # which doesn't apply to Prior Authority
+      def prune_navigation_stack; end
     end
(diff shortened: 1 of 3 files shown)


jsugarman (approved):

app/controllers/prior_authority/steps/base_controller.rb
```diff
@@ -22,6 +22,11 @@ def as
       def service
         Providers::Gatekeeper::PAA
       end
+
+      def prune_navigation_stack
```
dwhenry, line 26: would be good to get ride of this entirely.. but that will have to wait until we redo CRM7 task list

dwhenry (approved):

PR #513, opened 2024-02-08
Add shared date and time formats and use

jsugarman:
## Description of change
Add shared date and time formats and use

 [came out of ](https://dsdmoj.atlassian.net/browse/CRM457-979)

There are multiple locations where the date format
`'%d %B %Y'` is used. Adding and using a shared format
date/time  stamp is something rails provides and can help to ensure
consistency, as well as the potential for simple amendment.

```rb
\# instead of
my_date.strftime('%d %B %Y')
\# use
my_date.to_fs(:stamp)
```

## Notes for reviewer
This has only be used to apply the format in the multiple
places that the `'%d %B %Y'` date/time format is already used. 

There are no performance benefits to my knowledge, just consistency.

### DIFF
--- a/app/mailers/claim_submission_mailer.rb
+++ b/app/mailers/claim_submission_mailer.rb
@@ -66,7 +66,7 @@ class ClaimSubmissionMailer < GovukNotifyRails::Mailer
   end
 
   def submission_date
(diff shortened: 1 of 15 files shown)


patrick-laa (approved):
Very tidy!

PR #516, opened 2024-02-09
crm457-979: namespace tasklist section headers

jsugarman:
## Description of change
Namespace tasklist section headers

 [Came out of ticket](https://dsdmoj.atlassian.net/browse/CRM457-979)

To provide consistency with prior authority and,specifically, namespaced translation lookup for NSM
and PA applications and allow use of the same header
sub-key if desired.


dwhenry (approved):
LGTM

PR #518, opened 2024-02-09
CRM457-1110: Default to sorting by most recent

patrick-laa:
## Description of change
Put most recently created applications at the top of the list.

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1110)

### DIFF
--- a/app/controllers/prior_authority/applications_controller.rb
+++ b/app/controllers/prior_authority/applications_controller.rb
@@ -79,7 +79,7 @@ module PriorAuthority
     }.freeze
 
     def order_and_paginate(query)
-      order_template = ORDERS.fetch(params[:sort_by], 'created_at ?')
+      order_template = ORDERS.fetch(params[:sort_by], 'updated_at ?')
       direction = DIRECTIONS.fetch(params[:sort_direction], 'DESC')
       pagy(query.includes(:defendant).order(order_template.gsub('?', direction)))
     end
--- a/app/controllers/prior_authority/applications_controller.rb
+++ b/app/controllers/prior_authority/applications_controller.rb
@@ -65,7 +65,7 @@ module PriorAuthority
     }.freeze
 
     def order_and_paginate(query)
-      order_template = ORDERS.fetch(params[:sort_by], 'created_at ?')
+      order_template = ORDERS.fetch(params[:sort_by], 'created_at DESC')
       direction = params[:sort_direction] == 'descending' ? 'DESC' : 'ASC'
       pagy(query.includes(:defendant).order(order_template.gsub('?', direction)))
     end
--- a/app/controllers/prior_authority/applications_controller.rb
+++ b/app/controllers/prior_authority/applications_controller.rb
@@ -1,6 +1,7 @@
 module PriorAuthority
   class ApplicationsController < ApplicationController
     before_action :authenticate_provider!
+    before_action :set_default_table_sort_options
     before_action :load_drafts, only: %i[index draft]
     before_action :load_assessed, only: %i[index assessed]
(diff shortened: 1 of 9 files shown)


app/controllers/prior_authority/applications_controller.rb
```diff
@@ -65,7 +65,7 @@ def service
     }.freeze
 
     def order_and_paginate(query)
-      order_template = ORDERS.fetch(params[:sort_by], 'created_at ?')
+      order_template = ORDERS.fetch(params[:sort_by], 'created_at DESC')
```
dwhenry, line 68: so this set a default sort direct when no sort by is passed in. This might be alright but would require deep knowledge of what params are passed when to be sure....

can I suggest

```
  def order_and_paginate(query)
      order_by = ORDERS[params[:sort_by]] ? params[:sort_by] : 'created_at'
      direction = if params[:sort_direction]
          params[:sort_direction] == 'ascending' ? 'ASC' : 'DESC'
        else
          order_by == 'created_at' ? 'DESC' : 'ASC'
        end        
      pagy(query.includes(:defendant).order(order_by => direction))
    end
```

NOTE this sets the default sort direction if none is passed in based on the field being sorted

NOTE: ORDERS hash is less important but could now be used to store default sort direction

spec/system/prior_authority/application_lists_spec.rb
```diff
@@ -3,10 +3,10 @@
...
   before do
     visit provider_saml_omniauth_callback_path
-    create(:prior_authority_application, laa_reference: 'LAA-AAAAA', status: 'submitted', provider: Provider.first)
-    create(:prior_authority_application, laa_reference: 'LAA-BBBBB', status: 'submitted', provider: Provider.first)
-    create(:prior_authority_application, laa_reference: 'LAA-CCCCC', status: 'granted', provider: Provider.first)
-    create(:prior_authority_application, laa_reference: 'LAA-DDDDD', status: 'draft', provider: Provider.first)
+    create(:prior_authority_application, laa_reference: 'LAA-AAAAA', status: 'submitted', created_at: 1.day.ago)
+    create(:prior_authority_application, laa_reference: 'LAA-BBBBB', status: 'submitted', created_at: 2.days.ago)
+    create(:prior_authority_application, laa_reference: 'LAA-CCCCC', status: 'granted', created_at: 3.days.ago)
+    create(:prior_authority_application, laa_reference: 'LAA-DDDDD', status: 'draft', created_at: 4.days.ago)
```
dwhenry, line 9: so we should have a test that only applications for the current office are shown in the list - I have been getting errors when trying to access applicartion but it fails due to office code filtering later in the process. :)

dwhenry (approved):
Some suggestions - include if you think they help. :)

app/controllers/prior_authority/applications_controller.rb
patrick-laa: So I started off with something like that, but the issue I hit is that when sorting by client, we need to sort by both first name and last name, so `.order(column => direction)` doesn't work.

But you're right, my solution hinges on the fact that I know that the only situation where  `sort_by` is not provided is a situation where `sort_direction` is also not provided, and that's not made explicit in the code. So a more verbose working-out of `direction` is warranted. I'll sort that.

spec/system/prior_authority/application_lists_spec.rb
patrick-laa: Yeah... it's because to this point we've been ignoring office codes and simply looking at what applications the provider directly owns... I'll go ahead and fix that...

PR #534, opened 2024-02-15
CRM457-977: Stop auto-casting params to integers

patrick-laa:
## Description of change
The 'i' in the field name of multiparameter attributes [causes Rails](https://github.com/rails/rails/blob/d90b4e2/activerecord/lib/active_record/base.rb#L1884) to auto-cast to an integer. Which means we can't detect non-numerical values because by the time they reach us they are converted to "0".

This PR stops doing that, and adds validations to give custom errors for non-numbers and non-integer number

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-977) (see QA comments on subtasks where Mary has repeatedly pointed this out).


gems/laa_multi_step_forms/app/validators/time_period_validator.rb
```diff
@@ -43,7 +40,36 @@ def validate_period(time_period)
       add_error(:invalid)
     end
   end
-  # rubocop:enable Metrics/AbcSize
+
+  def validate_hours(hours)
+    add_error(:blank_hours) if hours.nil?
```
dwhenry, line 45: are we sure this is a nil and not an empty string? i.e. would this be better with `hours.blank?`

config/locales/en/errors.yml
```diff
@@ -432,3 +417,13 @@ en:
               invalid: |
                 Name must only include letters a to z, and special characters such as hyphens, spaces and apostrophes.
                 Name must be 2 words or more.
+      messages:
```
dwhenry, line 420: how does this work to link the fields?

dwhenry (approved):
LGTM - just a quick check on nil v blank

gems/laa_multi_step_forms/app/validators/time_period_validator.rb
patrick-laa: Interestingly Rails Magic (TM) means that by the time the value makes it to our custom attribute type, a blank string has been converted into a nil. But since it's magic I think you're right, better to check for blanks since blanks are explicitly what we're interested in. I've updated this.

config/locales/en/errors.yml
patrick-laa: If ActiveModel can't find a message for a specific error type on a specific field it turns out it falls back to searching `language.active_model.errors.messages.error_type`. Obviously for error types like `blank`, `greater_than` etc there's no point having a fallback because the message will be so different for different fields. But as we're using some very specific, explicit error type names, like `blank_hours`, we can provide sensible defaults here and only override them if needed.

I was going to use the `&shared_time_errors` / `<<: *shared_time_errors` pattern, but that only works if the shared fields are all in the same file, and we recently took the decision to start splitting errors out of one massive `errors.yml` file, so the time errors in `alternative_quotes.yml` couldn't be shared that way.

PR #544, opened 2024-02-17
CRM457-979: Move `actions` header resposibility to cards and add row actions capability

jsugarman:
## Description of change
Refactor CYA report and card objects and enable row actions

The Change action links in the card "header" are the responsibility
of the individual card objects rather than the report object.

One card, Application details, should not have this link on
the header but on the Ufn row instead, according to designs
and, by moving the responsibility to the card object, this
becomes easy.

In addition, the govuk_summary_list accepts an ` actions:` key
on the individual rows so we use this to add a link to a specifc
row (UFN).

### Before changes:

Screenshot 2024-02-17 at 18 28 22

### After changes:

Screenshot 2024-02-17 at 18 27 39

### DIFF
--- a/app/presenters/prior_authority/check_answers/base.rb
+++ b/app/presenters/prior_authority/check_answers/base.rb
@@ -3,6 +3,9 @@ module PriorAuthority
   module CheckAnswers
     class Base
       include LaaMultiStepForms::CheckMissingHelper
+      include ActionView::Helpers::UrlHelper
+      include GovukLinkHelper
+      include GovukVisuallyHiddenHelper
       include ActionView::Helpers::TagHelper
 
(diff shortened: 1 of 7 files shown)


app/presenters/prior_authority/check_answers/base.rb
```diff
@@ -43,9 +46,24 @@ def row_content(head_key, text, head_opts = {})
...
+          actions: actions
         }
       end
+
+      def actions
+        helper = Rails.application.routes.url_helpers
+
+        [
+          govuk_link_to(
+            'Change',
```
patrick-laa, line 59: I18nify?

patrick-laa (approved):

PR #550, opened 2024-02-19
CRM457-979: Do not display travel cost reason on CYA for prison law matters

jsugarman:
## Description of change
Do not display travel cost reason on CYA for prison law matters

Prison law matters do not required the provider to answer
the question of whether the client is detained nor provide
a reason for travel if they are not detained.


patrick-laa (approved):
Good catch

PR #551, opened 2024-02-19
Crm457 1137/validation bug

dwhenry:
## Description of change
Fixes a bug with custom service type not allowing per_item billing to pass validation in the task list

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1137)

### DIFF
--- a/app/forms/prior_authority/steps/primary_quote_form.rb
+++ b/app/forms/prior_authority/steps/primary_quote_form.rb
@@ -1,50 +1,28 @@
 module PriorAuthority
   module Steps
     class PrimaryQuoteForm < ::Steps::BaseFormObject
-      def self.attribute_names
-        super - %w[service_type custom_service_name]
-      end
-
-      def initialize(attrs)
-        # We have logic to set these below if a service type suggestion is provided,
-        # and we don't want the arbitrary order of attribute assignment to overwrite that
-        if attrs.key?('service_type_suggestion')
-          attrs.delete('service_type')
-          attrs.delete('custom_service_name')
-        else
-
-          # This ensures that the 'service type suggestion' field in the UI is
-          # is pre-populated with the custom name
-          if attrs[:application].service_type == 'custom'
-            attrs[:application].service_type = attrs[:application].custom_service_name
-          end
-
-          # But otherwise, we need to pull in application-wide settings to this quote-specific model
-          attrs[:service_type] ||= attrs[:application].service_type
-          attrs[:custom_service_name] ||= attrs[:application].custom_service_name
-        end
-        super(attrs)
-      end
-
-      attribute :service_type, :value_object, source: QuoteServices
-      attribute :custom_service_name, :string
       attribute :contact_full_name, :string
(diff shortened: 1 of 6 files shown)


app/forms/prior_authority/steps/primary_quote_form.rb
```diff
@@ -1,50 +1,37 @@
...
-        # The complication is that if a user types in a value to the text field but doesn't click the autocomplete,
-        # service_type_suggestion will contain an accurate translated value, but service_type dropdown contains
-        # a false untranslated value.
-        # So to avoid being caught out, the translated value needs to take precedence, and we need to infer the
-        # untranslated value from that.
+      # Using local variable for service_type_autocomplete to avoid issues with
+      # assignment of value into two fields service_type and custom_service_name
+      attr_accessor :service_type, :custom_service_name, :local_values
+
+      def service_type_autocomplete
```
dwhenry, line 17: this feels like a smell but is required to support this working without JS

ivanELEC (approved):

app/forms/prior_authority/steps/primary_quote_form.rb
```diff
@@ -1,50 +1,37 @@
...
+        scope = local_values ? self : application
+        scope.service_type == 'custom' ? scope.custom_service_name : scope.service_type
+      end
+
+      # this should only be called when JS is disabled
```
patrick-laa, line 22: This comment is a tiny bit misleading: When JS is enabled this _is_ called, but it's called _before_ `service_type_autocomplete_suggestion=`, meaning the logic is overruled (I'm typing this slowly because I'm trying everything out before I say anything to make sure I'm not wrong _again_.

Whereas when JS is disabled, e.g. in when running `spec/system/prior_authority/service_cost_spec.rb:16`, this runs _and `service_type_autocomplete_suggestion=` _doesn't_ run, because the select "name" attribute doesn't have `_suggestion` appended (although it's "data-name" attribute _does_ have that appended)

This does lead to a slightly weird behaviour I just found. Repro steps:

1. JS enabled
2. Start typing a service name e.g. Lab fees, and click on the autosuggestion. Don't save yet
3. Then delete it, so service name is blank
4. Now click 'Save and come back later'
5. Then return to the primary quote page.

**Expected behaviour**: Service name is not prepopulated
**Actual behaviour**: Service name is prepopulated with 'Lab fees'

Why? Because thanks to your click in step 2, line 27 here assigns "lab_fees" to self.service_type. Then because the autosuggest value is set to a blank string thanks to the delete in step 3, the `if` statement on line 39 returns false, so service_type is _not_ overwritten.

It's such an edge case that I don't know if that counts as a bug or not. If it is, I guess the fix is to change line 39 to:


```ruby
 self.service_type = value.present? ?  'custom' : nil
 ```

But either way, we need to be wary that `service_type_autocomplete=` _is_ called when JS is enabled, so I think this comment is worth updating.

patrick-laa (approved):

PR #552, opened 2024-02-19
CRM457-1108: Enable submission to app store

jsugarman:
## Description of change
Enable submission to app store

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1108)

Submit on branch then check app store and or case worker app for the relevant application.


gems/laa_multi_step_forms/app/controllers/steps/base_step_controller.rb
```diff
@@ -33,7 +33,7 @@ def update_and_advance(form_class, opts = {})
       hash = permitted_params(form_class).to_h
       record = opts.fetch(:record, current_application)
       @form_object = form_class.new(
-        hash.merge(application: current_application, record: record)
+        hash.merge(application: current_application, record: record, commit_draft: params.key?(:commit_draft))
```
patrick-laa, line 36: I'm not entirely convinced by this change. I think it leads to the form object knowing a bit too much in almost every case, just to fix this one edge condition.

What you could do instead is overwrite ` def save!` on CheckAnswersForm, as that will be called instead of `save` if a user clicks 'save and come back later'. That way you can differentiate between the 2 behaviours in your form object without having to change the base step controller or the base form object.

PR #567, opened 2024-02-21
Crm457 1130/navigation flow

dwhenry:
## Description of change
Rework how navigation works when editing from the check answers page

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1130)


app/services/decisions/custom_wrapper.rb
```diff
@@ -3,5 +3,12 @@ class CustomWrapper < SimpleDelegator
     def any_invalid?(scope, *forms)
       scope.detect { |record| forms.any? { |f| !f.build(record, application:).valid? } }
     end
+
+    def source_was_check_answers?(rule, any_source: false)
```
patrick-laa, line 7: So I think this approach can still work, but we need to find a way of making it feel less like a hack.

I think instead of the DecisionTree saying "after screen X go to the start page" and then us hijacking it to go to the check answers page instead, the DecisionTree should say "after screen X go to whatever the appropriate summary page is", and then either here or elsewhere, our logic should run to say "the' appropriate summary page' is either the start page or the check answers page depending on the state history.

app/controllers/nsm/steps/start_page_controller.rb
```diff
@@ -15,6 +15,16 @@ def show
         @application = current_application
         render 'laa_multi_step_forms/task_list/show', locals: { header: -> { decision_step_header }, app_type: 'claim' }
       end
+
+      # we remove the check_answers entry from the list to reset the return
+      # location after edits once the task list has been displayed
+      def append_navigation_stack
+        current_application.navigation_stack -= [
```
patrick-laa, line 22: This feels a leetle bit like a abuse of the navigation stack. Essentially we're using it to set a flag saying either "check answers page" or "start page". Doing it by adding and removing from the navigation stack is maybe less tidy than, say, stashing something in the session, or in a standalone database field?

app/services/decisions/decision_tree.rb
```diff
@@ -99,25 +99,27 @@ class DecisionTree < DslDecisionTree
...
     # prison law flow
     from(:next_hearing)
-      .when(-> { application.navigation_stack[-1].match?('check_answers') })
-      .goto(edit: PRIOR_AUTHORITY_CHECK_ANSWERS)
       .goto(show: PRIOR_AUTHORITY_START_PAGE)
 
     # non-prison law flow
     from(:case_detail)
-      .when(-> { application.navigation_stack[-1].match?('check_answers') })
+      .when(->(rule) { source_was_check_answers?(rule, any_source: true) })
```
patrick-laa, line 122: In theory is this rendered unnecessary by the (ANY) clause above?

patrick-laa (commented):
I see what you mean. This could work, but I'm not in love with it. In CCQ, because all steps were handled by the same controller, we simply made the ChangeAnswersController subclass the main StepsController, so you always knew if you were in a "change answers loop" from the check your answers screen because the URL was different and the controller was different, so when a "loop" was complete it was trivial to work out whether to go to the check answers screen or not. If there was a way of passing the "we're in a change answers loop" state through the path or querystring that might simplify things, but I don't know if we'd be able to preserve that state through back buttons, form submissions etc. It's a tough one.

app/services/decisions/decision_tree.rb
dwhenry: :( I wish... as normally the loop is case_details -> hearing details... 

but both these section have an edit button on the summary screen so the assumption is you exit the loop early - just for this one...

app/controllers/nsm/steps/start_page_controller.rb
dwhenry: I have left this at the moment over adding a separate DB field.... 

my thoughts are the navigation stack contains the locations you have visited and are still allowed to visit.. (used in the `in_progress` check in task list
So visiting the task list removes the `permission` on check_answers... so this maybe what it was intended for

patrick-laa: Yeah all right

patrick-laa (approved):
You know what, that's actually quite elegant to my eyes. Nice one!

PR #583, opened 2024-02-23
CRM457-977: Primary quote QA fixes

patrick-laa:
## Description of change
When the service type changes, we don't want irrelevant previously-entered values to stick around. So we do more resetting of attributes on quote cost forms and the primary quote form

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-977)

### DIFF
--- a/app/forms/prior_authority/steps/primary_quote_form.rb
+++ b/app/forms/prior_authority/steps/primary_quote_form.rb
@@ -58,6 +58,7 @@ module PriorAuthority
         return false unless save_file
 
         save_quote
+        reset_quote_cost_fields
         application.update(service_type:, custom_service_name:) if service_type
 
         # If a change to service type has rendered any alternative quotes invalid,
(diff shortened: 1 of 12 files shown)


app/forms/prior_authority/steps/primary_quote_form.rb
```diff
@@ -82,6 +83,41 @@ def save_quote
...
+        return unless service_type_changed?
+
+        values_to_reset = {}
+
+        # If we have changed cost type, e.g. from per_item to per_hour, or we have changed item type,
+        # e.g. from pages to words, then values previously entered shouldn't carry across
+        values_to_reset.merge!(QUOTE_ITEM_FIELD_RESETS) if previous_service_rule.item != current_service_rule.item
+
+        # Cost type is per_item, per_hour, or variable
+        values_to_reset.merge!(QUOTE_ITEM_FIELD_RESETS) if current_service_rule.cost_type != :per_item
```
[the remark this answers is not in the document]
patrick-laa: I don't think we want to start filling in values that the user has not chosen. The steer from Mary was to lean towards making the user re-enter more information if they change the service type rather than 'try to make sense' of the data we already have. I was on the fence about what to do if we change from per_x to variable. I think we want to ensure that they see that they now have a choice of cost types and make them confirm which one they want, not just assume based on what they entered when they had no choice in the matter. But if I previously filled in item fields for an enforced per_item service, then I change to a variable service, if I then actively select per_item, should my previously entered values be waiting for me in the conditional reveal? My gut is no, but I could be convinced otherwise.

spec/forms/prior_authority/steps/primary_quote_form_spec.rb
```diff
@@ -249,6 +249,64 @@
...
+
+        it 'clears out old item data for all quotes' do
+          expect(application.reload.primary_quote).to have_attributes(
+            items: nil,
+            cost_per_item: nil
+          )
+        end
+
+        it 'removes alternative quotes entirely' do
+          expect(application.alternative_quotes.count).to eq 0
```
jsugarman, line 306: nit i find expecting a change to be clearer to read, rather than relying on the before { save } mechanism
```rb
expect { save }.to change(application.alternative_quotes, :count).from(1).to(0)
```

spec/forms/prior_authority/steps/primary_quote_form_spec.rb
```diff
@@ -249,6 +249,64 @@
...
+                 quotes: [
+                   build(:quote, :primary, period: 30, cost_per_hour: 1.23),
+                   build(:quote, :alternative, period: 60, cost_per_hour: 1.25)
+                 ])
+        end
+
+        before { save }
+
+        it 'clears out old item data for all quotes' do
+          expect(application.reload.primary_quote).to have_attributes(
```
jsugarman, line 299: **nit** i find expecting a change to be clearer to read, rather than relying on the before { save } mechanism

```rb
  expect { save }.to change { application.primary_quote.reload.attributes }
    .from(hash_including())
    .to(hash_including())
```

jsugarman (approved):
some testing nits but otherwise LGTM

PR #584, opened 2024-02-23
CRM457-979: Make check answers look more like the prototype

patrick-laa:
## Description of change
These discrepancies are also affecting CRM457-1158 (in progress: readonly view of submitted applications)

### Before changes:
Screenshot 2024-02-23 at 09 40 55

### After changes:
Screenshot 2024-02-23 at 09 40 33

### DIFF
--- a/app/presenters/prior_authority/check_answers/base.rb
+++ b/app/presenters/prior_authority/check_answers/base.rb
@@ -49,7 +49,6 @@ module PriorAuthority
         {
           key: {
             text: heading,
-            classes: 'govuk-summary-list__value-width-50'
           },
           value: {
             text:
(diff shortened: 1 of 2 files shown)


app/presenters/prior_authority/check_answers/base.rb
```diff
@@ -49,7 +49,6 @@ def row_content(head_key, text, actions = [], head_opts = {})
         {
           key: {
             text: heading,
-            classes: 'govuk-summary-list__value-width-50'
```
dwhenry, line 52: without this I thought each section would be a different width based on the size of the text leading to a non-uniform appearance???

dwhenry (approved):
left a comment, but happy if design is happy

app/presenters/prior_authority/check_answers/base.rb
patrick-laa: No, the standard width for the key is 30%

PR #596, opened 2024-02-26
Setup and clamav test amendments

jsugarman:
## Description of change
Setup and clamav test amendments

- bump postgres14 to 16 for brew
- Amend db setup config inline with repo rename
- Add clamav install script and tweak tests

For clamav This adds a tested install script that also performs the
necessary configuration for having clamav work locally on
mac os.

In addition the tests for the file uploader, which were failing
locally, have been fixed to stub out the actual clamav calls
since we are not using an actual "infected" file (and the tests
are more to exercise the ENV var and host environement config
impact on scanning in any event).

pull branch locally, run `bin/install_clamav_on_mac`, run rails server, attempt to 
upload a viral file on one of the upload pages. An [Eicar](https://www.eicar.org/download-anti-malware-testfile/
) fake malicious file has been 
added to fixtures for manual (and automated) testing use.


bin/install_clamav_on_mac
```diff
@@ -0,0 +1,148 @@
...
+  printf "."
+
+  if (( $(date +%s)-start > 60 ))
+  then
+    printf "timed out!\n"
+    break;
+  fi
+
+  sleep 5;
+done;
```
dwhenry, line 148: WTF...

dwhenry (approved):
WTF is that install file...

bin/install_clamav_on_mac
jsugarman: something we used in apply and i had tweaked to, generally, just work. Have you tried it?

jsugarman: this bit is just waiting until clamav restarts with the new config.

PR #606, opened 2024-02-27
Setup for local e2e development

jsugarman:
## Description of change
Document setup for e2e development

[Comes out of ticket](https://dsdmoj.atlassian.net/browse/CRM457-992)

### DIFF
--- a/docs/development-e2e-setup.md
+++ b/docs/development-e2e-setup.md
@@ -1,0 +1,89 @@
+# Development end-to-end setup
+
+
+## Setup
+To run all the applications in a fashion that enables a provider to submit a claim or application and then see that in the case worker app for assessment you can follow this guide.
+
+For provider app submission you will need to ensure you have the following env vars in your `.env.development.local`. These will enable authenticated communication with the local running version of the app store. Ensure it is the `...local` version and that no secrets are inadvertently committed to the repository.
+
+```sh
+# App store credentials (for authenticating to a local app store currently)
+# TODO: Oauth may be removed from local/development app store in the future
+APP_STORE_CLIENT_ID=not-a-real-key
+APP_STORE_CLIENT_SECRET=not-a-real-key
+APP_STORE_TENANT_ID=not-a-real-key
+```
+Ask a dev where these can be retrieved, and do not use production versions
+
+To receive emails to you own email address, or avoid errors on submission, you should supply the following ENV VAR
+```sh
+GOVUK_NOTIFY_API_KEY=not-a-real-key
+```
+You can generate your own "test" key (that does not send emails) or a limited key (that only sends to a controlled list of emails). You will need access to https://www.notifications.service.gov.uk/ or our existing secrets to supply this value. Ask a dev.
+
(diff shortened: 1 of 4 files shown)


docs/development-e2e-setup.md
```diff
@@ -0,0 +1,85 @@
...
+
+- Submit a claim or prior authority application in the provider app
+- See a successful cron job being processed in the Sidekiq WebUI (at http://localhost/sidekiq).
+- See the successful authentication and processing in the app store server output.
+- See the submitted claim or application in the caseworker app list of claims or list of applications.
+
+
+## Trouble shooting
+
+- Missing translation errors in provider app
```
patrick-laa, line 70: This issue is now ticketed as CRM457-1183. Might be worth sticking that ticket number in this doc to make it more likely we remember to update this document to take the stale info out once that ticket is done.

docs/development-e2e-setup.md
```diff
@@ -0,0 +1,85 @@
...
+SIDEKIQ_WEB_UI_USERNAME=sidekiq
+SIDEKIQ_WEB_UI_PASSWORD=sidekiq
+```
+
+## Running e2e development stack
+
+### Run the provider App:
+  - Follow above readme to setup the development requirements
+
+  - Amend Procfile.dev to run rails server on a unique port of your choice
```
patrick-laa, line 36: I'd almost recommend we hard-code 3001 into the procfile. If we're all going to have both caseworker and provider up and running simultaneously more and more, maybe we should all get into the habit of having one on 3001 and the other on 3002 or something, by having those as standard in the procfile.

patrick-laa (approved):
Feel free to ignore my suggestions - I'm not 100% convinced myself that they add benefit.

docs/development-e2e-setup.md
jsugarman: i did muck about with `${PORT:-3000}` type logic but there are so many ways it seems to go wrong (puma config, foreman start command, Procfile.dev) that i have never found a robust solution, here or elsewhere.

So i would vote for hardcoding 3001/3002 into provider and caseworker 👍

docs/development-e2e-setup.md
```diff
@@ -0,0 +1,85 @@
...
+```sh
+# App store credentials (for authenticating to a local app store currently)
+# TODO: Oauth may be removed from local/development app store in the future
+APP_STORE_CLIENT_ID=not-a-real-key
+APP_STORE_CLIENT_SECRET=not-a-real-key
+APP_STORE_TENANT_ID=not-a-real-key
+```
+Ask a dev where these can be retrieved, and do not use production versions
+
+To receive emails to you own email address, or avoid errors on submission, you should supply the following ENV VAR
```
jsugarman, line 18:
```suggestion
To receive emails to your own email address, or avoid errors on submission, you should supply the following ENV VAR
```

docs/development-e2e-setup.md
jogilder: Agreed with this

PR #609, opened 2024-02-28
Fix erroneous append of js command in `Procfile.dev`

jsugarman:
## Description of change
Correct Procfile.dev 

Fix erroneous append of js command


patrick-laa (approved):
Thank you!

PR #623, opened 2024-03-04
[CRM457-988] Wait for ready

dwhenry:
## Description of change
When reloading the page the file drop area would occasionally not load, this was due to the JS being called before the html had rendered

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-988)


app/javascript/file-upload.js
```diff
@@ -124,11 +124,13 @@ MOJFrontend.MultiFileUpload.prototype.getErrorHtml = function(message) {
...
 };
 
-if (typeof MOJFrontend.MultiFileUpload !== "undefined") {
-    new MOJFrontend.MultiFileUpload({
-        container: $(".moj-multi-file-upload"),
-        uploadUrl: window.location.href,
-        deleteUrl: window.location.href
-    })
-}
+// $(function() {
```
patrick-laa, line 127: I _think_ this isn't supposed to be commented out, but let me know if there's some clever JS magic going on that's gone above my head.

dwhenry: i was testing was an actual fix by turning it off... opps...

ivanELEC (approved):

PR #628, opened 2024-03-07
Crm457 867/ufn validation

dwhenry:
## Description of change
Ensure first 6 digits of UFN is a valid date
* update field label to include '(UFN)' text
* add error for dates which are in the future
* improve existing invalid error
* all changes applied to both NSM and PA

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-867)


app/validators/ufn_validator.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    @record = record
+    @attribute = attribute
+
+    matched, *date_parts_strings = *value.match(%r{\A(\d{2})(\d{2})(\d{2})/\d{3}\z})
+    date_parts = date_parts_strings.reverse.map(&:to_i)
+
+    if matched && Date.valid_date?(*date_parts)
+      # convert to 4 digit date
+      date_parts[0] += 2000
```
patrick-laa, line 16: Is there _any_ possibility that the system will ever deal with some very old case from the 90s? I don't think so, but worth just checking. If so I guess we say any year that's... greater than the current year... plus arbitrary constant years... is from the 20th century...?

patrick-laa (approved):

PR #630, opened 2024-03-07
CRM457-1101: Pull PA updates

patrick-laa:
## Description of change
Process updates to state of PA applications reported by the app store

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1101)

## Notes for reviewer
This is for a caseworker ticket, that has an AC saying when a PA application is granted etc, it should move into the appropriate tab on the provider app

### DIFF
--- a/app/jobs/pull_updates.rb
+++ b/app/jobs/pull_updates.rb
@@ -2,9 +2,6 @@ class PullUpdates < ApplicationJob
   # queue :default
 
   def perform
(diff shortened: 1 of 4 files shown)


jogilder (approved):
LGTM other than we are calling them 4 and 7 when NSM or PA would be better and maybe even having type as an enum but hey, time.

And do we think that we might need two convert_params one for each type? Or do we do that as required?

PR #631, opened 2024-03-08
SPIKE - thoughts on interrupt

dwhenry:
## Description of change
how I was thinking we could easily do teh interrupt screens

### DIFF
--- a/app/forms/prior_authority/steps/case_detail_form.rb
+++ b/app/forms/prior_authority/steps/case_detail_form.rb
@@ -4,6 +4,7 @@ module PriorAuthority
       attribute :rep_order_date, :multiparam_date
       attribute :client_detained, :boolean
       attribute :subject_to_poca, :boolean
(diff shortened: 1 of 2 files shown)


app/forms/prior_authority/steps/case_detail_form.rb
```diff
@@ -4,18 +4,28 @@ class CaseDetailForm < ::Steps::BaseFormObject
...
       validates :subject_to_poca, inclusion: { in: [true, false] }
+      validate :client_detained_interrupt
 
       attr_accessor :main_offence_id, :custom_main_offence_name, :local_main_offence_values,
                     :prison_id, :custom_prison_name, :local_prison_values,
                     :defendant_attributes
 
+      def client_detained_interrupt
+        return if interrupt_cleared
+        return unless client_detained != application.client_detail
```
jsugarman, line 23: should this be
```suggestion
        return unless client_detained != application.client_detained
```

PR #638, opened 2024-03-14
Crm457 998 Email Submit Notifications

ivanELEC:
## Description of change

 [CRM4 Provider - Submit notification](https://dsdmoj.atlassian.net/browse/CRM457-998)

- Started Namespacing Mailers
- Add Mailer for PA Submission
- Use Mailer when submitting PA application to app store
- Added functionality in quote, additional_cost and prior_authority_application for totals


app/models/prior_authority_application.rb
```diff
@@ -49,4 +51,12 @@ def youth_court_applicable?
   def psychiatric_liaison_applicable?
     court_type == PriorAuthority::CourtTypeOptions::CENTRAL_CRIMINAL.to_s
   end
+
+  def total_cost
+    [primary_quote&.total_cost, additional_costs.sum(&:total_cost), alternative_quotes.sum(&:total_cost)].compact.sum
```
patrick-laa, line 56: Bug: the cost of the alternative quotes is _not_ part of the total cost.

app/models/quote.rb
```diff
@@ -9,4 +9,23 @@ class Quote < ApplicationRecord
...
+      travel_cost + additional_cost_value
+    end
+  end
+
+  def travel_cost
+    travel_cost_per_hour * travel_time / 60
+  end
+
+  def additional_cost_value
+    additional_cost_total || 0
```
patrick-laa, line 29: `additional_cost_total` is only populated for alternative quotes. For the primary quote, additional costs have to be calculated by summing the value of additional cost objects.

app/services/submit_to_app_store.rb
```diff
@@ -35,6 +35,7 @@ def submit(submission)
   end
 
   def notify(submission)
-    ClaimSubmissionMailer.notify(submission).deliver_later! if submission.is_a?(Claim)
+    Nsm::SubmissionMailer.notify(submission).deliver_later! if submission.is_a?(Claim)
+    ::PriorAuthority::SubmissionMailer.notify(submission).deliver_later! if submission.is_a?(PriorAuthorityApplication)
```
patrick-laa, line 39: If REDIS_HOST is not defined, so we're _not_ doing this job asynchronously, I don't think we should deliver the emails asynchronously either. i.e. if we come to this from line 10 instead of line 20, we should be doing `deliver!` not `deliver_later!`

patrick-laa (commented):
In general I'm not a huge fan of adding all those maths methods to the model objects. We already have that maths worked out in a bunch of places, so let's just reuse that, because there are a lot of gotchas that otherwise you need to duplicate working around.

app/mailers/nsm/submission_mailer.rb
```diff
@@ -0,0 +1,77 @@
...
+    def case_reference
+      @claim.laa_reference
+    end
+
+    def unique_file_number
+      @claim.ufn
+    end
+
+    def main_defendant
+      @claim.defendants.find { |defendant| defendant.main == true }
```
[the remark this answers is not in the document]
patrick-laa: Actually, you can just do `@claim.main_defendant` as that method has already been created for you in the model.

app/mailers/nsm/submission_mailer.rb
```diff
@@ -0,0 +1,77 @@
...
+    def case_reference
+      @claim.laa_reference
+    end
+
+    def unique_file_number
+      @claim.ufn
+    end
+
+    def main_defendant
+      @claim.defendants.find_by(main: true)
```
patrick-laa, line 35: So this one is my fault - I was late to spot that there's already a `main_defendant` method on the Claim object you can use, but that's probably preferable, as otherwise methods like `defendant_name` will all cause a SQL query to occur.

app/models/additional_cost.rb
```diff
@@ -1,3 +1,11 @@
 class AdditionalCost < ApplicationRecord
   belongs_to :prior_authority_application
+
+  def total_cost
```
patrick-laa, line 4: I think we're better off reusing our existing logic here given my comment from my previous review. How about:

```ruby
PriorAuthority::Steps::AdditionalCosts::DetailForm.build(self, application: prior_authority_application).total_cost
```

patrick-laa (commented):
Think we're nearly there but you do have a couple of failing specs.

patrick-laa (approved):

PR #639, opened 2024-03-14
CRM457-992: Add custom item type specific cost error messages

jsugarman:
## Description of change
Add item cost error messages

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-992)

Inline designs/content for error messages and QA feedback. This can/ should also be applied to provider

Reference:
see [assess PR](https://github.com/ministryofjustice/laa-assess-crime-forms/pull/347)


patrick-laa (approved):
Neat. I like it.

PR #643, opened 2024-03-18
[CRM457-1240] Process all pages when polling

dwhenry:
## Description of change
Fix bug in branches when polling not working

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1240)

### DIFF
--- a/app/jobs/pull_updates.rb
+++ b/app/jobs/pull_updates.rb
@@ -4,7 +4,19 @@ class PullUpdates < ApplicationJob
   def perform
     json_data = HttpPuller.new.get_all(last_update)
 
-    json_data['applications'].each do |record|
+    loop do
+      json_data = HttpPuller.new.get_all(since:, count:)
+      break if json_data['applications'].none?
+
+      last_updated = process(json_data['applications'])
+      break if since == last_updated
+
+      since = last_updated
+    end
+  end
+
+  def process(applications)
+    applications.each do |record|
       case record['application_type']
       when 'crm7'
         update_claim(record['application_id'], convert_params(record))
@@ -12,6 +24,8 @@ class PullUpdates < ApplicationJob
         update_prior_authority_application(record['application_id'], convert_params(record))
       end
     end
+
+    Time.zone.parse(applications.last['updated_at'])
   end
 
   private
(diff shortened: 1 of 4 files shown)


spec/services/http_puller_spec.rb
```diff
@@ -24,10 +24,18 @@
...
         expect(described_class).to receive(:get)
-          .with('http://some.url/v1/applications?since=1',
+          .with('http://some.url/v1/applications?since=1&count=20',
                 headers: { authorization: 'Bearer test-bearer-token' })
 
-        subject.get_all(1)
+        subject.get_all(since: 1)
+      end
+
+      it 'can set the coutn varible to set page size' do
```
patrick-laa, line 33: typo - count

spec/jobs/pull_updates_spec.rb
```diff
@@ -20,6 +20,9 @@
 
   before do
     allow(HttpPuller).to receive(:new).and_return(http_puller)
+    allow(http_puller).to receive(:get_all).and_return('applications' => [])
+    allow(http_puller).to receive(:get_all).with(since: Time.zone.local(2023, 1, 1), count: 100)
```
patrick-laa, line 24: Nit: It's not at all clear from this spec why http_puller would receive a `since` of specifically `Time.zone.local(2023, 1, 1)`. The fallback to that date isn't something we mention in a different spec. I wonder if as a marginal improvement, PullUpdates should define that date as a constant and we could reference that constant here? But not crucial by any means.

patrick-laa (approved):

app/jobs/pull_updates.rb
```diff
@@ -1,17 +1,32 @@
...
+    applications.each do |record|
       case record['application_type']
       when 'crm7'
         update_claim(record['application_id'], convert_params(record))
       when 'crm4'
         update_prior_authority_application(record['application_id'], convert_params(record))
       end
     end
+
+    Time.zone.parse(applications.last['updated_at'])
```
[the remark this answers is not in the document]
dwhenry: It is order by default from the service, but you are correct we should rely on this as it could change in the future.

PR #646, opened 2024-03-19
Use pgrep to probe sidekiq for liveness/readiness

jsugarman:
## Description of change
Use pgrep to probe sidekiq for liveness/readiness

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1248)

We should really be testing if sidekiq is available for liveness and readiness, rather
than any root path available on the worker (which may just be the sidekiq webUI root path 🤷)

### DIFF
--- a/bin/sidekiq_health_check
+++ b/bin/sidekiq_health_check
@@ -1,0 +1,9 @@
+#!/usr/bin/env ruby
+
+module Worker
+  def self.healthcheck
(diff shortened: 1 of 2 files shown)


jsugarman:
closing as we are using [sidekiq_alive](https://github.com/arturictus/sidekiq_alive), which enables the use of httpGet over a specified port

PR #647, opened 2024-03-19
Automate deployment to prod for main

jsugarman:
## Description of change
Automate deployment to prod for main

Until we go to private beta, and perhaps even after that,
we might as well deploy to UAT and then production automatically.

This will:
- expose any deployment issues immediatley on a per merge basis
- not incur overhead of devs remebering to deploy


jogilder (approved):
LGTM

PR #655, opened 2024-03-21
Crm457 1192 further info

ivanELEC:
## Description of change

- Add further_information_explanation and incorrect_information_explanation fields
- Sync these fields when status is "sent_back"

Generally for tickets where these fields are needed from the caseworker app


app/services/prior_authority/assessment_syncer.rb
```diff
@@ -53,6 +57,14 @@ def sync_additional_cost(cost, cost_data, application)
         )
       end
 
+      def sync_further_info_request(application, app_store_record)
+        data = app_store_record['application']
+        application.update(
+          further_information_explanation: data['further_information_explanation'],
```
patrick-laa, line 63: So there's a slight complication here: If the caseworker checks 'further information', fills in some further information in the box that appears, then changes their mind and _unchecks_ further information, `data['further_information_explanation']` will still be populated, even though no further information is needed. As a safeguard, we should look at `data['updates_needed']` to see what fields are explicitly asked for, rather than just inferring it from what explanation texts have been populated.

jogilder (commented):
@ivanELEC can we split 'update needed' into two sections? Then have two sync functions
One for further info and a separate one for incorrect information?

app/services/prior_authority/assessment_syncer.rb
```diff
@@ -5,9 +5,13 @@ def call(application)
         app_store_record = AppStoreClient.new.get(application.id)
 
         sync_overall_comment(application, app_store_record)
```
patrick-laa, line 7: So Joel [raised the point](https://mojdt.slack.com/archives/C05212WRK5J/p1711041737323699?thread_ts=1711041195.429809&cid=C05212WRK5J) that this method should be called for `rejected`/`granted` and `part_grant` applications, but _not_ for `sent_back` applications, because (a) when an application is sent back there will not be a single overall comment, and (b) even if there was, this method assumes that there will be an `Event::Decision` event. (I think it's the absence of that that is making your specs fail)

ivanELEC:
> @ivanELEC can we split 'update needed' into two sections? Then have two sync functions One for further info and a separate one for incorrect information?

as in have update_needed as two different fields instead of an array and then 2 sync functions? The reason I kept it as one sync function is that these values need to be reset with each RFI, therefore the function basically stays the same if you split them up so feels like adding redundant code + they come from the same form

ivanELEC:
> @ivanELEC can we split 'update needed' into two sections? Then have two sync functions One for further info and a separate one for incorrect information?

but also I just realised I hadn't saved the file before sending the commit off so this review is based on incorrect code I sent!

patrick-laa (approved):

PR #656, opened 2024-03-21
Do not inline sidekiq for bin/dev

jsugarman:
## Description of change
Do not inline sidekiq for bin/dev

So it behaves as a hosted environments sidekiq worker
as expected and stats are updated.

without this the job will be processed inline with the side
effect that stats on the webui are not updated.

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,4 +1,4 @@
-web: bin/rails server -p 3001
+web: RUN_SIDEKIQ_IN_TEST_MODE=false bin/rails server -p 3001
 js: yarn build --watch
 css: yarn build:css --watch
 sidekiq: DISABLE_SIDEKIQ_ALIVE=true bundle exec sidekiq
--- a/.env.development
+++ b/.env.development
@@ -35,3 +35,5 @@ CLAMBY_ENABLED=true
 GOVUK_NOTIFY_API_KEY=not-a-real-key
 
 EOL_URL=https://apply-for-extension-of-upper-limits.dev.form.service.justice.gov.uk/
+
+RUN_SIDEKIQ_IN_TEST_MODE=true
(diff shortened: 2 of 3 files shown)


Procfile.dev
```diff
@@ -1,4 +1,4 @@
-web: bin/rails server -p 3001
+web: RUN_SIDEKIQ_IN_TEST_MODE=false bin/rails server -p 3001
```
patrick-laa, line 1: Think this is unnecessary - by default, RUN_SIDEKIQ_IN_TEST_MODE is not set, and in your initializer, if it's not set, it defaults to false. So we'd only want to set this like this if we'd set RUN_SIDEKIQ_IN_TEST_MODE to true somewhere.

jsugarman: could set it to nothing or anything other than 'true', but this was the only thing that seemed to get my sidekiq stats actually showing up and not running inline using `bin/dev`

patrick-laa: I guess it's just the inconsistency that .env.development in caseworker has `RUN_SIDEKIQ_IN_TEST_MODE=true`, so that needs to be overridden in the procfile to stop it happening, whereas in the provider app that line isn't there, so by default in develop mode sidekiq won't run in test mode, so overriding it from nil to "false" won't affect the logic in the initialiser either way.

jsugarman: i wonder if there is some shell variable leakage going between the two as well, not too mention sidekiq job processing leakage, as fixing it for caseworker seemed to get stats showing up on provider...i think.

Anyway, I have amended the commit so it defaults to inline true for  typical `rails s` but then this switches it of for `bin/dev`  - this is at least consistent behaviour with caseworker (following my PR on that repo), and gets bin/dev doing what i would hope it does.

patrick-laa (approved):

PR #666, opened 2024-03-26
Crm457 1192 further info form

ivanELEC:
## Description of change

 [CRM4 - Provider Create Further Information Form](https://dsdmoj.atlassian.net/browse/CRM457-1192)

## Notes for reviewer

- Complete syncing for further information and incorrect info
- Add further_informations to model to store requests for further information
- Add form for further information and associated documents and validation
- [Add further information form to tasklist and decision tree](https://mojdt.slack.com/archives/C05212WRK5J/p1711377812065099)


db/migrate/20240325121503_add_columns_to_further_information.rb
```diff
@@ -0,0 +1,9 @@
+class AddColumnsToFurtherInformation < ActiveRecord::Migration[7.1]
+  def change
+    add_column :further_informations, :caseworker_id, :uuid
+    add_column :further_informations, :requested_at, :datetime
+    add_column :further_informations, :expired_at, :datetime
+    remove_column :prior_authority_applications, :further_information_explanation
+    remove_column :prior_authority_applications, :incorrect_information_explanation
```
jsugarman, line 7: You should be able to use a single block

```
    change_table :further_informations, bulk: true do |t|
      t.uuid :caseworker_id
      ...
      t.remove :further_information_explanation
      ...
    end
```

app/controllers/prior_authority/steps/further_information_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+    class FurtherInformationController < BaseController
+      skip_before_action :verify_authenticity_token, only: [:create, :destroy]
+
+      def edit
+        @form_object = FurtherInformationForm.build(record, application: current_application)
+        @explanation = record.information_requested
+        @supporting_documents = record.supporting_documents
+      end
+
+      def create
```
patrick-laa, line 12: nit: A loooot of this class is a copy paste from another controller. It'd be nice to maybe move the shared stuff into a shared `MultiFileUploadable` concern or something.

app/models/supported_file_types.rb
```diff
@@ -11,4 +11,8 @@ class SupportedFileTypes
     application/msword application/vnd.openxmlformats-officedocument.wordprocessingml.document
     application/rtf image/jpeg image/bmp image/png image/tiff application/pdf
   ].freeze
+  FURTHER_INFORMATION = %w[
```
patrick-laa, line 14: nit: All these constants are identical. And I think they're always going to be identical - we have one set of rules about what files we accept, and we apply it everywhere. It might be we just need one `SUPPORTED_FILE_TYPES` constant.

app/controllers/prior_authority/steps/start_page_controller.rb
```diff
@@ -8,15 +8,34 @@ def show
...
+            view_context, application: current_application
+          )
+        end
+      end
+
+      def further_information_needed
+        if current_application.further_informations.empty?
+          false
+        else
+          last_further_info_request = current_application.further_informations.last.created_at
```
patrick-laa, line 35: I think we need to be explicit that we want the most recent further_information. So that would be: `further_informations.order(:created_at).last.created_at`, but we can do even better and just do `current_application.further_informations.maximum(:created_at)` and avoid loading a whole model object.

app/services/prior_authority/assessment_syncer.rb
```diff
@@ -1,82 +1,118 @@
...
+
+    def sync_further_info_request
+      latest_further_info = data['further_information'].last
+
+      application.further_informations.create(
+        {
+          caseworker_id: latest_further_info['caseworker_id'],
+          information_requested: latest_further_info['information_requested'],
+          requested_at: latest_further_info['requested_at'],
+          expired_at: latest_further_info['expires_at']
```
patrick-laa, line 103: I think this field is redundant, as we only care about the current deadline which we've already stored in `resubmission_deadline`. But if we do keep this field, it should be `expires_at`, not `expired_at`, because it's the date that it _will_ happen, not the date that it _has_ happened.

app/models/supported_file_types.rb
ivanELEC: yeah I kept making new ones purely if they changed in the future but I think you're right

app/controllers/concerns/multi_file_uploadable.rb
```diff
@@ -0,0 +1,79 @@
...
+    return_success({ evidence_id: evidence.id, file_name: params[:documents].original_filename })
+  rescue FileUpload::FileUploader::PotentialMalwareError => e
+    return_error(e, { message: t('activemodel.errors.messages.suspected_malware') })
+  rescue StandardError => e
+    return_error(e, { message: t('activemodel.errors.messages.upload_failed') })
+  end
+
+  # :nocov:
+  def destroy
+    raise 'Implement in subclass'
```
patrick-laa, line 24: I don't think there's much value in these. First, this is a module, so there are no subclasses. Second, these aren't methods that get called by other methods in this module, so these raises will never get triggered by someone forgetting to implement them. I'd just delete them.

app/controllers/prior_authority/steps/further_information_controller.rb
```diff
@@ -0,0 +1,38 @@
+module PriorAuthority
+  module Steps
+    class FurtherInformationController < BaseController
+      include MultiFileUploadable
+
+      def edit
+        @form_object = FurtherInformationForm.build(record, application: current_application)
+        @explanation = record.information_requested
```
patrick-laa, line 8: This might be neater if we made it available as a method on @form_object rather than its own variable. Same for `@supporting_documents`.

app/helpers/application_helper.rb
```diff
@@ -18,6 +18,15 @@ def multiline_text(string)
     ApplicationController.helpers.sanitize(string.gsub("\n", '<br>'), tags: %w[br])
   end
 
+  def further_information_needed
+    if current_application.further_informations.empty?
+      false
+    else
+      last_further_info = current_application.further_informations.order(:created_at).last.created_at
+      (current_application.status == 'sent_back') && (last_further_info >= current_application.app_store_updated_at)
```
patrick-laa, line 26: Handily the `enum` magic means (I learned recently) we can do `current_application.sent_back?` instead of `current_application.status == 'sent_back'`

app/presenters/prior_authority/tasks/further_information.rb
```diff
@@ -0,0 +1,35 @@
...
+      def path
+        edit_prior_authority_steps_further_information_path
+      end
+
+      def completed?
+        FORM.build(record, application:).valid?
+      end
+
+      def can_start?
+        PREVIOUS_TASKS.all? { |task| fulfilled?(task) }
```
patrick-laa, line 24: I _think_ this could just be `true`, as you'll never see this card if you're not allowed to start it.

app/services/prior_authority/assessment_syncer.rb
```diff
@@ -1,82 +1,117 @@
...
+    def info_correction_required?
+      data['updates_needed'].include?('incorrect_information')
+    end
+
+    def resubmission_deadline
+      data['resubmission_deadline']
+    end
+
+    def sync_further_info_request
+      latest_further_info = data['further_information'].last
```
patrick-laa, line 96: For safety (there are no promises about what order this array might be in), might be worth doing `data['further_information'].max_by { _1['requested_at'] }`

spec/controllers/prior_authority/reason_why_controller_spec.rb
```diff
@@ -1,90 +1,6 @@
 require 'rails_helper'
 
 RSpec.describe PriorAuthority::Steps::ReasonWhyController, type: :controller do
-  before do
-    allow(controller).to receive(:current_application).and_return(current_application)
-  end
-
-  describe '#create' do
```
patrick-laa, line 8: How come these specs have all been removed? Did they break? I wouldn't have expected reason why controller's behaviour to have changed.

spec/jobs/pull_updates_spec.rb
```diff
@@ -125,128 +119,6 @@
         expect(PriorAuthority::AssessmentSyncer).to have_received(:call).with(application)
       end
     end
-
-    context 'with multiple updates' do
```
patrick-laa, line 129: These also feel like valuable specs. How come they've been removed?

spec/controllers/prior_authority/reason_why_controller_spec.rb
ivanELEC: sorry I was under the impression the code I was bringing in from Joel's branch was an update to this, will look into this now

app/presenters/prior_authority/tasks/further_information.rb
ivanELEC: I think you're right now that I think about it

spec/controllers/prior_authority/reason_why_controller_spec.rb
ivanELEC: sorry this comment was relative to the pull_updates_spec tests. These tests haven't been removed, they are run in the shared example

patrick-laa (approved):

PR #674, opened 2024-03-27
CRM457-1277: Handle unrecognised office codes

patrick-laa:
## Description of change
We weren't handling unrecognised office codes when a particular service was specified.
 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1277)


jsugarman (approved):
i was going to safe in slack, do we just need a safe navigator here

PR #677, opened 2024-03-28
Stop restricting hosts in development mode

patrick-laa:
So that e2e tests, which specify a hostname, can use development mode easily.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -59,8 +59,6 @@ Rails.application.configure do
   # Highlight code that triggered database queries in logs.
   config.active_record.verbose_query_logs = true
 
-  config.hosts = ['crm7dev.apps.live.cloud-platform.service.justice.gov.uk', 'localhost', '127.0.0.1', '.ngrok-free.app']
-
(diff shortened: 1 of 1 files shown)


jogilder (approved):
LGTM

config/environments/development.rb
```diff
@@ -59,8 +59,6 @@
   # Highlight code that triggered database queries in logs.
   config.active_record.verbose_query_logs = true
 
-  config.hosts = ['crm7dev.apps.live.cloud-platform.service.justice.gov.uk', 'localhost', '127.0.0.1', '.ngrok-free.app']
```
dwhenry, line 62: as per my slack comment:

https://guides.rubyonrails.org/configuring.html#config-hosts

we open a potential security hole by removing this. So unless we are confident we guard this elsewhere I would prefer to have a host value being set (add to assess if it is missing as well)

patrick-laa: As mentioned on Slack, this value is env-var-configurable, and this is _just_ development mode.

PR #680, opened 2024-03-28
CRM457-1281 - build test data

dwhenry:
## Description of change
Build NSM data for performance testing


lib/test_data/nsm_builder.rb
```diff
@@ -0,0 +1,81 @@
...
+        args, kwargs = *options(**options).values.sample
+        claim = FactoryBot.create(*args, kwargs.call)
+
+        invalid_tasks = check_tasks(claim)
+
+        if invalid_tasks.any?
+          Rails.logger.debug { "Invalid for #{invalid_tasks.map(&:last).join(', ')}" }
+
+          # rubocop:disable Lint/Debugger
+          debugger if ENV['DEBUG_TEST_DATA']
```
dwhenry, line 22: should I remove this as was super useful when debugging why create failed?

spec/factories/disbursements.rb
```diff
@@ -8,7 +8,7 @@
     trait :valid do
       disbursement_date { age.days.ago.to_date }
       disbursement_type { DisbursementTypes.values.reject(&:other?).sample.to_s }
-      miles { 50 + (200**Random.new.rand(1.0).to_i) } # range: 50 -> 250
+      miles { 50 + (200**Random.new.rand(1.0).to_i) } # range: 50 -> 250 with logorithmic distribution
```
patrick-laa, line 11: I'm not entirely sure what you mean by logarithmic distribution here, but I don't think this code does what you think it does. `Random.new.rand(1.0)` is always less than 1, so `Random.new.rand(1.0).to_i` is always 0, so `(200**Random.new.rand(1.0).to_i)` is always 1, so this always sets miles to 51

patrick-laa: Ah, ok, I see what you mean by logarithmic distribution. You just need to move the `)` that's currently to the right of `to_i` the left of it.

dwhenry: I don't understand why the wrote the rand function like this... I realised halfway through that the 1.0 broke things and thought I had removed it from everywhere... :(

fixed now

also for reference:
```
irb(main):001> Random.new.rand(1.0)
=> 0.9108281877297423
irb(main):002> rand(1.0)
=> 0
```

lib/test_data/nsm_builder.rb
```diff
@@ -0,0 +1,86 @@
+module TestData
+  class NsmBuilder
+    def build_many(bulk: 100, large: 4, year: 2023)
+      raise 'Do not run on production' if HostEnv.production?
+
+      bulk.times do
+        build(year:)
+
+        # avoid issues with large number of applications with the same last_updated_at time
+        sleep 0.1 unless HostEnv.local?
```
patrick-laa, line 10: Why is this not an issue locally?

lib/test_data/nsm_builder.rb
```diff
@@ -0,0 +1,86 @@
...
+      end
+
+      Rails.logger.info "Created large examples: #{large_ids.to_sentence}"
+    end
+
+    def build(**options)
+      ActiveRecord::Base.transaction do
+        args, kwargs = *options(**options).values.sample
+        claim = FactoryBot.create(*args, kwargs.call)
+        # Nsm::Steps::SolicitorDeclarationForm.new(application: claim, record: claim, signatory_name: Faker::Name.name).save
```
patrick-laa, line 27: Guessing this can be deleted?

lib/test_data/nsm_builder.rb
```diff
@@ -0,0 +1,86 @@
...
+          end
+        ],
+        breach: [
+          [:claim, :complete, :case_type_breach, :build_associates],
+          proc do
+            date = date_for(year)
+            { date: date, disbursements_count: rand(max / 2), work_items_count: rand(min..max), updated_at: date }
+          end
+        ],
+        no_disburesments: [
```
patrick-laa, line 64: typo `disbursements`

patrick-laa (approved):

lib/test_data/nsm_builder.rb
dwhenry: so this is mostly to avoid the delay in the tests. I will change it to `Rails.env.test?` so this is more clear

PR #681, opened 2024-03-28
CRM457-1173: Stop casting date inputs to integers automatically

patrick-laa:
## Description of change
If someone puts 'banana' into the month part of a date field, we shouldn't pre-populate the field with `0` when we re-render the form - we should pre-populate it with 'banana'. Same as we do with `gbp` and `fully_validatable_integer` fields.

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1173)

### DIFF
--- a/gems/laa_multi_step_forms/app/lib/govuk_design_system_formbuilder/elements/fully_validatable_date.rb
+++ b/gems/laa_multi_step_forms/app/lib/govuk_design_system_formbuilder/elements/fully_validatable_date.rb
@@ -1,0 +1,28 @@
+module GOVUKDesignSystemFormBuilder
+  module Elements
+    class FullyValidatableDate < ::GOVUKDesignSystemFormBuilder::Elements::Date
+      # Unlike Date, this does not append an `i` to the values of SEGMENTS,
+      # meaning that Rails does not automatically cast values to integers,
+      # allowing for more expressive error messages if a non-numerical string
+      # is entered.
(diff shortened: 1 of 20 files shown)


gems/laa_multi_step_forms/app/lib/govuk_design_system_formbuilder/elements/fully_validatable_date.rb
```diff
@@ -0,0 +1,28 @@
...
+        format(
+          '%<object_name>s[%<input_name>s(%<segment>s)]',
+          object_name: @object_name,
+          input_name: @attribute_name,
+          segment: SEGMENTS.fetch(segment)
+        )
+      end
+    end
+  end
+end
```
dwhenry, line 28: have we looked at pushing this upstream - ideally with it being an override option to avoid break existing usage?

dwhenry (approved):

gems/laa_multi_step_forms/app/lib/govuk_design_system_formbuilder/elements/fully_validatable_date.rb
patrick-laa: I've raised it as an issue, and looks like they've started work on doing just that: https://github.com/x-govuk/govuk-form-builder/compare/make-date-segments-configurable

PR #686, opened 2024-04-02
Crm457 1281/randomise test data

dwhenry:
## Description of change
Split this to a separate PR to merge quicker

Randomise the data that is produced for the tests, this will allow reuse of the Factories to create the performance testing data

### DIFF
--- a/spec/presenters/nsm/check_answers/case_details_card_spec.rb
+++ b/spec/presenters/nsm/check_answers/case_details_card_spec.rb
@@ -15,7 +15,7 @@ RSpec.describe Nsm::CheckAnswers::CaseDetailsCard do
 
   describe '#row_data' do
     it 'generates case detail rows with no remittal' do
-      expect(subject.row_data).to eq(
+      expect(subject.row_data).to match(
         [
           {
             head_key: 'main_offence',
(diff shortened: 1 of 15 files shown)


spec/presenters/nsm/check_answers/case_details_card_spec.rb
```diff
@@ -15,15 +15,15 @@
...
+      expect(subject.row_data).to match(
         [
           {
             head_key: 'main_offence',
             text: 'Theft'
           },
           {
             head_key: 'main_offence_date',
-            text: '1 January 2023'
+            text: /\A\d{1,2} \w+ \d{4}\z/
```
patrick-laa, line 26: This feels to me like it weakens the test. "0 Banana 0000" would pass. Since we're here testing the precise formatting of dates (and tests like this allowed us to spot where there was a formatting error with leading zeroes being present in the day numbers), I think it's reasonable that these tests would be premised on having an explicit date provided in the test setup, and a matching string representation of that specific date being expected in the test details.

dwhenry: yeah, it was find with `yyyy-mm-dd` but these felt more flaky... I will set a explicit date for these tests.

patrick-laa (approved):

PR #691, opened 2024-04-04
CRM457-1181: Stop using combined name fields

patrick-laa:
## Description of change
- Split full_name fields to first_name and last_name fields
- Due to shared models, it was easier to update NSM at the same time than leave it to tidy up later
- I also tweaked the UI field widths, label texts and header sizes to match the designs for the affected screens

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1181)

### DIFF
--- a/app/services/submit_to_app_store/prior_authority/solicitor_payload_builder.rb
+++ b/app/services/submit_to_app_store/prior_authority/solicitor_payload_builder.rb
@@ -7,9 +7,9 @@ class SubmitToAppStore
 
       def payload
         @application.solicitor.as_json(only: %i[
-                                         full_name
                                          reference_number
-                                         contact_full_name
+                                         contact_first_name
+                                         contact_last_name
                                          contact_email
                                        ])
       end
--- a/spec/services/submit_to_app_store/prior_authority_payload_builder_spec.rb
+++ b/spec/services/submit_to_app_store/prior_authority_payload_builder_spec.rb
@@ -53,8 +53,8 @@ RSpec.describe SubmitToAppStore::PriorAuthorityPayloadBuilder do
         },
         solicitor: {
           contact_email: 'james@email.com',
-          contact_full_name: 'James Blake',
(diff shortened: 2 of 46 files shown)


app/services/submit_to_app_store/prior_authority/solicitor_payload_builder.rb
```diff
@@ -7,9 +7,9 @@ def initialize(application)
 
       def payload
         @application.solicitor.as_json(only: %i[
-                                         full_name
```
dwhenry, line 10: are we intentionally dropping this field?

spec/services/submit_to_app_store/prior_authority_payload_builder_spec.rb
```diff
@@ -53,8 +53,8 @@
         },
         solicitor: {
           contact_email: 'james@email.com',
-          contact_full_name: 'James Blake',
-          full_name: 'Richard Jenkins',
```
dwhenry, line 57: we have dropped the solicitor name here

dwhenry (changes requested):
Think you have dropped a field on the solicitor in the submit payload. :)

app/services/submit_to_app_store/prior_authority/solicitor_payload_builder.rb
patrick-laa: Yes, it's intentional - we don't use the solicitor name field in PA, we use the solicitor contact name.

spec/services/submit_to_app_store/prior_authority_payload_builder_spec.rb
patrick-laa: As above - we only collect one solicitor name in PA, which we call the contact name. The solicitor model has 2 name fields, one prefixed with 'contact'. There's no point sending the other one over, as it will always be nil

PR #692, opened 2024-04-04
Fix assessment syncer error handling

patrick-laa:
## Description of change

### DIFF
--- a/app/services/prior_authority/assessment_syncer.rb
+++ b/app/services/prior_authority/assessment_syncer.rb
@@ -22,7 +22,7 @@ module PriorAuthority
         sync_sent_back_request
       end
     rescue StandardError => e
-      Sentry.capture_exception("#{self.class.name} encountered error '#{e}' for application '#{application.id}'")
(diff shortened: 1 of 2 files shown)


app/services/prior_authority/assessment_syncer.rb
```diff
@@ -22,7 +22,7 @@ def call
         sync_sent_back_request
       end
     rescue StandardError => e
-      Sentry.capture_exception("#{self.class.name} encountered error '#{e}' for application '#{application.id}'")
+      Sentry.capture_message("#{self.class.name} encountered error '#{e}' for application '#{application.id}'")
```
dwhenry, line 25: any reason we don;t capture the error instead of a custom message... as that would give us a backtrace on the issue which would help with debug (maybe)

dwhenry (approved):
maybe

app/services/prior_authority/assessment_syncer.rb
patrick-laa: My guess is that Joel/Ivan wanted to ensure they captured the application ID. I'm not _sure_ if it's possible to do that with a `capture_exception`

PR #698, opened 2024-04-09
CRM4 - Provider Application - Notify App Store When Provider Updates Application

ivanELEC:
## Description of change

 [CRM4 - Provider Application - Notify App Store When Provider Updates Application](https://dsdmoj.atlassian.net/jira/software/c/projects/CRM457/boards/1192?selectedIssue=CRM457-1275)

- Add PUT request to app store client
- Add logic to submit to app store job to use PUT request when:
  - app store last updated at is present
  - submission is for prior authority
- add tests


app/services/submit_to_app_store.rb
```diff
@@ -31,7 +31,12 @@ def submit(submission)
     # able to support re-sending/submitting an appplication so we can ignore
     # put requests
     post_manager = AppStoreClient.new
-    post_manager.post(payload)
+
+    if submission.is_a?(PriorAuthorityApplication)
+      submission.app_store_updated_at.present? ? post_manager.put(payload) : post_manager.post(payload)
```
patrick-laa, line 36: Given it's not `post`ing, maybe we stop calling this variable `post_manager`. Maybe just `client`?

spec/services/app_store_client_spec.rb
```diff
@@ -88,6 +88,96 @@
...
+        expect(described_class).to receive(:put)
+          .with('http://localhost:8000/v1/application/',
+                body: message.to_json,
+                headers: { authorization: 'Bearer test-bearer-token' })
+
+        subject.put(message)
+      end
+    end
+
+    context 'when APP_STORE_USERNAME is present' do
```
patrick-laa, line 146: This test is junk - there is no APP_STORE_USERNAME or APP_STORE_PASSWORD - and I've a horrible feeling that if you `git blame` this file you'll see a context like this with my name on it, cos when I was refactoring the AppStoreClient I didn't spot that I was copy-pasting junk! But we can delete this whole context, I think.

app/services/submit_to_app_store.rb
```diff
@@ -31,7 +31,12 @@ def submit(submission)
     # able to support re-sending/submitting an appplication so we can ignore
     # put requests
     post_manager = AppStoreClient.new
-    post_manager.post(payload)
+
+    if submission.is_a?(PriorAuthorityApplication)
```
patrick-laa, line 35: You'll also need to look in SubmitToAppStore::PriorAuthority::EventBuilder because there are some TODO's to actually populate the event with the latest further_information data.

app/services/submit_to_app_store.rb
ivanELEC: yeah agreed

spec/services/app_store_client_spec.rb
ivanELEC: oh yeah this is using the old app store credentials. Yeah I felt weird about copying a load of tests and putting "put" there so reducing the copied stuff is a relief

app/services/submit_to_app_store.rb
```diff
@@ -30,8 +30,13 @@ def submit(submission)
     # TODO: we only do post requests here as the system is not currently
     # able to support re-sending/submitting an appplication so we can ignore
     # put requests
-    post_manager = AppStoreClient.new
-    post_manager.post(payload)
+    client = AppStoreClient.new
+
+    if submission.is_a?(PriorAuthorityApplication)
+      submission.app_store_updated_at.present? ? client.put(payload) : client.post(payload)
```
patrick-laa, line 36: Oh, and while I remember, you'll probably need to update PriorAuthorityPayloadBuilder to include the further_informations payload (the EventBuilder contains info about the most recent FI, but I think we agreed the payload would contain a history of _all_ FIs

app/services/app_store_client.rb
```diff
@@ -17,6 +17,17 @@ def post(message)
     end
   end
 
+  def put(message)
+    response = self.class.put("#{host}/v1/application/", **options(message))
```
dwhenry, line 21: shouldn't the URL have the application ID in. it?

dwhenry, line 21: shouldn't the URL have the application ID in. it?

app/services/submit_to_app_store.rb
ivanELEC: ooh sorry glossed over this one and sorted the event part only - will add this in shortly

app/services/submit_to_app_store/prior_authority/event_builder.rb
```diff
@@ -18,14 +18,41 @@ def payload
...
           }
         ]
       end
 
+      def further_information_comment
+        if further_information_supplied(@application)
+          @application.further_informations
+                      .order(:created_at).last.information_supplied
+        else
+          ''
```
dwhenry, line 34: any reason this is not nil?

spec/services/submit_to_app_store/prior_authority/event_builder_spec.rb
```diff
@@ -376,6 +402,8 @@
           expect(listed_corrections).to eq ['alternative_quote_2']
         end
       end
+
+      context 'when further info'
```
dwhenry, line 406: ??? can this be removed or are you expecting additional tests

spec/factories/prior_authority_application.rb
```diff
@@ -122,7 +123,9 @@
     end
 
     trait :with_further_information_supplied do
-      further_informations { [build(:further_information, :with_response, :with_supporting_documents)] }
+      after(:create) do |paa|
```
dwhenry, line 126: not sure what the advantages of the changes here are??

why the `after(:create)`

dwhenry (approved):
fix the empty context and seems to do the job

app/services/submit_to_app_store/prior_authority/event_builder.rb
ivanELEC: no, actually this should be nil, updating now

spec/services/submit_to_app_store/prior_authority/event_builder_spec.rb
ivanELEC: i think I wrote a whole test and didn't save the file :S

spec/factories/prior_authority_application.rb
ivanELEC: needed to do this to use the further_information objects in another let

PR #702, opened 2024-04-10
CRM457-1281 - Fix bug caused by random data

dwhenry:
## Description of change
Fix bug caused by random data


spec/services/submit_to_app_store/nsm_payload_builder_spec.rb
```diff
@@ -49,7 +49,7 @@
             'miles' => disbursement.miles.to_s,
             'other_type' => { en: nil, value: nil },
             'pricing' => pricing[disbursement.disbursement_type],
-            'prior_authority' => nil,
+            'prior_authority' => (disbursement.total_cost_without_vat >= 100.0 ? 'yes' : nil),
```
patrick-laa, line 52: It feels like the payload_builder_spec shouldn't have to know that there's this arbitrary logic around the magic number 100 in the factory. It just cares that the value of `prior_authority` in the payload matches the value of prior_authority in the disbursement. So could we just do:

```ruby
'prior_authority' => disbursement.prior_authority,
```

?

PR #703, opened 2024-04-11
CRM457-1327: Sub/Pub instead of cron

patrick-laa:
## Description of change

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1327)

### DIFF
--- a/app/services/app_store_update_processor.rb
+++ b/app/services/app_store_update_processor.rb
@@ -1,0 +1,35 @@
+class AppStoreUpdateProcessor
+  class << self
+    def call(record, is_full: false)
+      case record['application_type']
+      when 'crm7'
+        update_claim(record['application_id'], convert_params(record))
+      when 'crm4'
+        update_prior_authority_application(record['application_id'],
+                                           convert_params(record),
+                                           (record if is_full))
+      end
+    end
+
+    def convert_params(record)
+      {
+        status: record['application_state'],
+        app_store_updated_at: record['updated_at']
+      }
+    end
+
+    def update_claim(claim_id, params)
+      claim = Claim.find_by(id: claim_id)
+
+      claim&.update!(params)
+    end
+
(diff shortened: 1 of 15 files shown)


app/services/app_store_update_processor.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    def update_claim(claim_id, params)
+      claim = Claim.find_by(id: claim_id)
+
+      claim&.update!(params)
+    end
+
+    def update_prior_authority_application(application_id, params, full_record_or_nil)
+      application = PriorAuthorityApplication.find_by(id: application_id)
+      return unless application
```
dwhenry, line 29: I don;t understand... why can't this create new records?

helm_deploy/values-dev.yaml
```diff
@@ -74,7 +74,8 @@ variables:
   environment: development
   portalIDPMetadataFile: config/laa_portal/metadata/dev.xml
   spEntityId: LAA_Portal_CRM7_Dev
-  appStore: https://laa-crime-application-store-dev.apps.live.cloud-platform.service.justice.gov.uk
+  # TEMPORARILY POINT THIS BRANCH AT APP STORE BRANCH. TODO: REVERT BEFORE MERGE
+  appStore: https://spike-ruby-nscc-store-dev.cloud-platform.service.justice.gov.uk/
```
dwhenry, line 78: ....

dwhenry (commented):
I like the idea but I have one thing to say...

we currently load records in a 2 stage process - this is a terrible idea and should be removed - it was based on the idea on receiving the message, then creating a record that the data is saved to - it should just create the job to sync the record... this would be much nicer. :)

dwhenry (approved):
Ignore my last comment, I thought this was in the assess application because I';m an idiot...

PR #704, opened 2024-04-15
Fix References to Old Namespaces and Github Teams

ivanELEC:
## Description of change

- Reference correct Snyk Project
- Reference correct github team for codeowners
- Use laa-submit-crime-forms to name logstasher source


dwhenry (approved):
assuming this won;t break our assess with the owners change

PR #707, opened 2024-04-16
CRM457-1327: Allow bypassing sync authentication to enable e2e tests

patrick-laa:
## Description of change
- It's ok if we're in a non-auth environment if the app store webhook URL doesn't have a bearer token
- If we don't provide a bearer token, use a fallback header to tell the app store our role
- Allow launching the app in puma cluster mode so that if needed we can have multiple workers to avoid logjams where a request triggers a synchronous webhook callback that can't be processed.

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1327)

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -1,3 +1,7 @@
+# If this is set to a positive number, Puma will launch in cluster mode.
+# Note that on Mac you may need to also have env var `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES`
+workers ENV.fetch("WORKERS", 0)
+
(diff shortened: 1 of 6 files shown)


config/puma.rb
```diff
@@ -1,3 +1,7 @@
+# If this is set to a positive number, Puma will launch in cluster mode.
+# Note that on Mac you may need to also have env var `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES`
```
dwhenry, line 2: same comment for the .env.sample

dwhenry (approved):

PR #715, opened 2024-04-18
User local domain for production app store

ivanELEC:
## Description of change
Use local app store URL for production app store

[CRM457-1288](https://dsdmoj.atlassian.net/browse/CRM457-1288)

Confirmation of of successful call from to Production app store from Production provider app pod:
image

[CRM457-1288]: https://dsdmoj.atlassian.net/browse/CRM457-1288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/helm_deploy/values-prod.yaml
+++ b/helm_deploy/values-prod.yaml
@@ -74,7 +74,7 @@ variables:
   environment: production
   portalIDPMetadataFile: config/laa_portal/metadata/prod.xml
   spEntityId: LAA_Portal_CRM7_Prod
-  appStore: https://laa-crime-application-store-prod.apps.live.cloud-platform.service.justice.gov.uk
+  appStore: http://laa-crime-application-store-app.laa-crime-application-store-production.svc.cluster.local
   maxFileUploadBytes: 10485760
   eolUrl: https://apply-for-extension-of-upper-limits.form.service.justice.gov.uk/
   enableSyncTriggerEndpoint: 'false'


patrick-laa (approved):
Would it make sense for us to use the local domain for UAT too (even though the app store will _also_ be available via its public ingresses in that namespace), just to keep our pre-prod environment as similar to prod as possible? Just in case, say, the format of internal domain urls changes or some nonsense like that, UAT failing might give us an early warning.

ivanELEC:
> Would it make sense for us to use the local domain for UAT too (even though the app store will _also_ be available via its public ingresses in that namespace), just to keep our pre-prod environment as similar to prod as possible? Just in case, say, the format of internal domain urls changes or some nonsense like that, UAT failing might give us an early warning.

yeah I think this is a good shout - I was initially going to say Dev as I'd already setup the network policy stuff on dev but UAT makes more sense as the "closest thing to production" environment

helm_deploy/values-prod.yaml
```diff
@@ -74,7 +74,7 @@ variables:
   environment: production
   portalIDPMetadataFile: config/laa_portal/metadata/prod.xml
   spEntityId: LAA_Portal_CRM7_Prod
-  appStore: https://laa-crime-application-store-prod.apps.live.cloud-platform.service.justice.gov.uk
+  appStore: http://laa-crime-application-store-app.laa-crime-application-store-production.svc.cluster.local
```
jogilder, line 77: @ivanELEC does it matter that this is not https?

ivanELEC: So my assumption is that because this can only ever possibly be accessed internally that we are secured from outside noseying - this is also how LGA have it setup (all http not https)

PR #717, opened 2024-04-22
CRM457-1093: Modsec

patrick-laa:
## Description of change
DRY up modsec config, switch to standard settings, and add correct tagging
(Note we're _not_ adding the custom modsec rules from the security configuration review because they're... wrong.
 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1093)


dwhenry (approved):
I have no idea... so just assuming tested and correct

PR #723, opened 2024-04-23
Provider - specify environment for sentry

jogilder:
## Description of change
Change environment so that it doesn't automatically assume it is production


config/initializers/sentry.rb
```diff
@@ -1,5 +1,6 @@
 if ENV.fetch('SENTRY_DSN', nil).present?
   Sentry.init do |config|
+    config.environment = HostEnv.env_name
```
patrick-laa, line 3: This might be erroring out due to HostEnv not being loaded at the point Sentry is inited. You could try `ENV.fetch('ENV')`

PR #729, opened 2024-04-24
Crm457 974 fix last update finder

ivanELEC:
## Description of change

- Move corrected_info logic from PA payload builder to it's own class, ChangeLister
- Use ChangeLister to determine whether anything has changed in CheckAnswersForm validation
- Update tests 

 [CRM457-974](https://dsdmoj.atlassian.net/browse/CRM457-974)

[CRM457-974]: https://dsdmoj.atlassian.net/browse/CRM457-974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/submit_to_app_store/prior_authority_payload_builder.rb
```diff
@@ -5,20 +5,15 @@ def initialize(application:)
     end
 
     def payload
-      content = data
```
patrick-laa, line 8: By deleting this you call `data` twice, which means that the payload has to be fully constructed from scratch twice. If you don't like stashing it in a variable here, can I suggest you memoise the `data` method by starting it off with `@data ||=`?

app/forms/prior_authority/steps/check_answers_form.rb
```diff
@@ -23,11 +23,8 @@ def application_corrected
       end
 
       def application_changed_since_request?
-        last_updated_at > application.resubmission_requested
-      end
-
-      def last_updated_at
-        LastUpdateFinder.call(application)
```
patrick-laa, line 30: This is the only place where LastUpdateFinder is called, so if we're removing it here we should probably delete it entirely.

patrick-laa (approved):

app/forms/prior_authority/steps/check_answers_form.rb
```diff
@@ -23,11 +23,8 @@ def application_corrected
...
+        content = SubmitToAppStore::PriorAuthorityPayloadBuilder.new(application:).data
```
patrick-laa, line 26: Oh, something to look out for - at the moment EventBuilder is calling `deep_stringify_keys` on this hash in its initializer, before passing it in the ChangeLister. You might find that without doing that here, ChangeLister produces unexpected results. So might be work doing a deep_stringify_keys _inside_ change_lister so you don't need to call it before passing the data in.

PR #732, opened 2024-04-25
CRM457-750: Add validation to uplift 'Update calculation' buttons

patrick-laa:
## Description of change
- Switch from a 'save_and_refresh' to a 'reload' on click on 'Update calculation' for pages with uplift values
- Change the 'reload' behaviour to validate the form so that when the page is re-rendered it displays any validation errors (but will still update the calculation if it can)
- Add the missing error messaging
- Adjust the minimum uplift threshold from 0 to 1 to match the error messaging

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-750)

### DIFF
--- a/gems/laa_multi_step_forms/app/controllers/steps/base_step_controller.rb
+++ b/gems/laa_multi_step_forms/app/controllers/steps/base_step_controller.rb
@@ -16,12 +16,12 @@ module Steps
     def update
       raise 'implement this action, if needed, in subclasses'
     end
+    # :nocov:
 
     def reload
       @form_object.valid?
       render :edit
     end
-    # :nocov:
 
     private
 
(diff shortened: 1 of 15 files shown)


gems/laa_multi_step_forms/app/controllers/steps/base_step_controller.rb
```diff
@@ -18,7 +18,8 @@ def update
     end
 
     def reload
-      raise 'implement this action, if needed, in subclasses'
+      @form_object.valid?
+      render :edit
     end
     # :nocov:
```
jsugarman, line 24: should the :nocov: be moved up to line 19 given that it is now actual functionality

jsugarman (approved):
looks ok to me. are you confident that the new default `reload` functionality is application for all step controllers?

PR #735, opened 2024-04-26
CRM457-1370: New office code selection journey for PA

patrick-laa:
## Description of change
- Switch to new office code journey for PA
- Delete old office code stuff now that neither PA nor NSM are using it.

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1370)


app/models/prior_authority_application.rb
```diff
@@ -50,7 +42,7 @@ class PriorAuthorityApplication < ApplicationRecord
   scope :reviewed, -> { where(status: %i[granted part_grant rejected auto_grant sent_back expired]) }
   scope :submitted_or_resubmitted, -> { where(status: %i[submitted provider_updated]) }
 
-  scope :for, ->(provider) { where(office_code: provider.selected_office_code) }
+  scope :for, ->(provider) { where(office_code: provider.office_codes).or(where(provider:)) }
```
dwhenry, line 45: not 100% sure on this - if a provider has an office code removed from their list then they should no longer be able to see details for that office code - surely???

at least that is my assumption?

app/services/submit_to_app_store/prior_authority/firm_office_payload_builder.rb
```diff
@@ -8,13 +8,12 @@ def initialize(application)
...
         @application.firm_office.as_json(only: %i[
                                            name
-                                           account_number
                                            address_line_1
                                            address_line_2
                                            town
                                            postcode
                                            vat_registered
-                                         ])
+                                         ]).merge(account_number: @application.office_code)
```
dwhenry, line 16: this is fine if the user only has a single office code.. but if they have multiple wouldn't this pre-select the first one on the next screen?

dwhenry (commented):
So I think we need to step back slightly here there is a fundamental problem in that:
* we need office code populate when the application is created as it is what we filter on to determine what the user can see - you got around this by also including when the provider created the submission, but that means those submission are only visible to the creator until the office code if selected. :(
* we want the user to be able to select the office code during the process.
* point 1 means office code is pre-set before the selection screen is we don't want

what we really want is almost a fake firm object that represents the group of office codes, this could then be applied to the submission on creation with the office ocde selected in teh normal flow....... maybe

app/services/submit_to_app_store/prior_authority/firm_office_payload_builder.rb
patrick-laa: So this is the payload on submission to the app store, once the office code has either explicitly or implicitly been chosen. There is no next screen.

patrick-laa: (Note that at this point we're no longer referring to the user's office codes, we're just putting into the appropriate part of the payload the _application's_ office code)

app/models/prior_authority_application.rb
patrick-laa: Ok, so re this bit

> those submission are only visible to the creator until the office code if selected.

Aimee has confirmed that she's OK with that for now.

Re this bit:

> if a provider has an office code removed from their list then they should no longer be able to see details for that office code

What I've done is weaken this clause so that the second half of the 'or' only applies if the application doesn't have a selected office code.

dwhenry (approved):
based on agree comment frim aimee...

PR #741, opened 2024-04-29
Stop treating work items as invalid if uplift allowed but not applied

patrick-laa:
## Description of change
If you save a work item that is eligible for uplift, but don't apply an uplift, it gets saved to the database with a zero uplift. But then the dodgy form logic decides to say that that uplift _has_ been applied - an uplift of zero. And that's now invalid.

This PR changes the form logic so that if a work item is eligible for uplift and what has been saved is zero, it now correctly interprets that as uplift _not_ having been applied, which is a valid state.

### DIFF
--- a/app/forms/nsm/steps/work_item_form.rb
+++ b/app/forms/nsm/steps/work_item_form.rb
@@ -27,7 +27,7 @@ module Nsm
 
       def apply_uplift
         allow_uplift? &&
-          (@apply_uplift.nil? ? (uplift.present? && uplift.positive?) : @apply_uplift == 'true')
+          (@apply_uplift.nil? ? uplift.to_f.positive? : @apply_uplift == 'true')
       end
 
       def pricing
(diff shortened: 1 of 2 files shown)


app/forms/nsm/steps/work_item_form.rb
```diff
@@ -27,7 +27,7 @@ def allow_uplift?
 
       def apply_uplift
         allow_uplift? &&
-          (@apply_uplift.nil? ? uplift.present? : @apply_uplift == 'true')
+          (@apply_uplift.nil? ? (uplift.present? && uplift.positive?) : @apply_uplift == 'true')
```
jsugarman, line 30: Could shorten a bit:
```suggestion
          (@apply_uplift.nil? ? (uplift.to_f.positive?) : @apply_uplift == 'true')
```

nil.to_f.positive? => false
0.0001.to_f.positive? => true

jsugarman (approved):

PR #746, opened 2024-05-01
Crm457 1394/status card

dwhenry:
## Description of change
Status screen based on status

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1394)

## Notes for reviewer
The locations that rely on adjusted amounts have been left with a placeholder for now


app/presenters/nsm/check_answers/application_status_card.rb
```diff
@@ -3,42 +3,124 @@
...
+        if status == 'submitted'
+          join_strings(status_tag(status), submitted_data, translate(:awaiting_review), tag.br, claimed_amount)
         else
-          'returned'
+          join_strings(status_tag(status), submitted_data, tag.br, claimed_amount, allowed_amount(status))
+        end
+      end
+
+      def join_strings(*strings)
+        strings.map { |str| str == tag.br ? str : "<p>#{str}</p>" }.join
```
patrick-laa, line 51: This seems really weird to me. If I call `join_strings("foo", "bar", tag.br, "baz")` this will return:

```
<p>foo</p>
<p>bar</p>
<br>
<p>baz</p>
```

But there are already line breaks between bar and baz because they're in separate paragraphs. The br increases the gap between bar and baz but by a very non-standard amount - the design system is very specific about the gaps between paragraphs, and adding extra `br` elements between paragraphs just messes with that.

patrick-laa: Also, I'm really not a fan of manually building HTML strings that you then have to sanitize. Much better to create sanitized strings in the first place, i.e. replace `"<p>#{str}</p>"` with `tag.p(str)`

app/presenters/nsm/check_answers/application_status_card.rb
```diff
@@ -3,42 +3,124 @@
...
+
+        'Pending Data'
+      end
+
+      def translate(key)
+        I18n.t("nsm.steps.check_answers.groups.#{group}.#{section}.#{key}")
+      end
+
+      def status_tag(_status)
+        "<strong class=\"govuk-tag #{I18n.t("nsm.claims.index.status_colour.#{claim.status}")}\">" \
```
patrick-laa, line 75: Likewise here, this could be:

```ruby
tag.strong(I18n.t("nsm.claims.index.status.#{claim.status}"), 
                   class: "govuk-tag #{I18n.t("nsm.claims.index.status_colour.#{claim.status}")}"
```

app/presenters/nsm/check_answers/application_status_card.rb
```diff
@@ -3,42 +3,124 @@
...
+              helper.url_for(controller: 'nsm/steps/view_claim', action: :show, id: claim.id, section: 'adjustments',
+                             anchor: type, only_path: true),
+              class: 'govuk-link--no-visited-state'
+            )
+          end
+        end
+        tag.ul sanitize(li_elements.join, tags: %w[a li]), class: 'govuk-list govuk-list--bullet'
+      end
+
+      # TODO: implement logic here to check for changes
```
patrick-laa, line 115: So should this method be called `any_changes?`? If not, why are we checking for changes here?

app/presenters/nsm/check_answers/application_status_card.rb
dwhenry: to the first point... take that up with @mary as that is the design.. :)

ivanELEC (approved):

PR #750, opened 2024-05-02
6 is the magic number

patrick-laa:
Internal testers routinely hit anomaly scores of 5, whatever that means. Let's set the anomaly score threshold to 6, which is what the caseworker app has.

### DIFF
--- a/helm_deploy/templates/ingress.yaml
+++ b/helm_deploy/templates/ingress.yaml
@@ -25,6 +25,7 @@ metadata:
     nginx.ingress.kubernetes.io/modsecurity-snippet: |
       SecRuleEngine On
       SecDefaultAction "phase:2,pass,log,tag:github_team=ministryofjustice/laa-crime-forms-team"
+      SecAction "id:900110,phase:1,nolog,pass,t:none,setvar:tx.inbound_anomaly_score_threshold=6"
     external-dns.alpha.kubernetes.io/aws-weight: "100"
     {{- toYaml . | nindent 4 }}
   {{- end }}


jogilder:
laa-assess-crime-forms is one of only two teams setting this for modsec
https://github.com/search?q=setvar%3Atx.inbound_anomaly_score_threshold%3D+path%3Ahelm_deploy&type=code

Not sure this is the way we should be doing this. Maybe instead we put this into detect only mode and see what comes up?
```
nginx.ingress.kubernetes.io/modsecurity-snippet: |
  SecAuditEngine On
  SecRuleEngine DetectionOnly
  SecDefaultAction "phase:2,pass,log,tag:github_team=<my-great-team-name>,tag:<your_random_tag_name>=<your_random_identifier>"
```
https://user-guide.cloud-platform.service.justice.gov.uk/documentation/networking/modsecurity.html#ingress-annotations

helm_deploy/templates/ingress.yaml
```diff
@@ -25,6 +25,7 @@ metadata:
     nginx.ingress.kubernetes.io/modsecurity-snippet: |
       SecRuleEngine On
       SecDefaultAction "phase:2,pass,log,tag:github_team=ministryofjustice/laa-crime-forms-team"
+      SecAction "id:900110,phase:1,nolog,pass,t:none,setvar:tx.inbound_anomaly_score_threshold=6"
```
jogilder, line 28: Not sure this is the way we should be doing this. Maybe instead we put this into detect only mode and see what comes up?

jogilder: Ok to go ahead with this but we should understand what it means.

jogilder (approved):
LGTM

PR #752, opened 2024-05-03
CRM457-1309: Partial postcodes

patrick-laa:
## Description of change
Allow partial postcodes but require a town for primary quote
 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1309)


jsugarman (commented):
This looks fine but wondering whether there was a decision to always enforce town presence. 

From a business PoV it feels like we should only enforce the need for a town if we do not have a full postcode. But i guess it may be possible for a full postcode to cover areas inside and outside of a "town"??!

As an aside this will invalidate existing claims/applications i imagine too?! Do we need a data migration for UAT/dev perhaps

spec/forms/prior_authority/steps/primary_quote_form_spec.rb
```diff
@@ -24,7 +25,8 @@
   let(:contact_first_name) { 'Joe' }
   let(:contact_last_name) { 'Bloggs' }
   let(:organisation) { 'LAA' }
-  let(:postcode) { 'CR0 1RE' }
+  let(:town) { 'Townville' }
```
dwhenry, line 28: it's https://en.wikipedia.org/wiki/Townsville

jsugarman (approved):

PR #769, opened 2024-05-10
CRM457-1455: Amend autoscaler config to use boolean

jsugarman:
## Description of change
Amend autoscaler config to use boolean

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1455)

inline with other apps. not sure how use of "enabled" string sneaked
in.


jogilder (approved):
LGTM

PR #771, opened 2024-05-10
CRM457-1471: Avoid cookie overflow errors

patrick-laa:
## Description of change

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1471)


jsugarman (commented):
I know this the same as the assess PR but after looking up related blogs/docs I was wondering whether we need to us the redis-rails gem for this. does rails not include this in any event

https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-rediscachestore

There are also warnings about using the same redis database for volatile and non-volatile jobs.

patrick-laa:
@jsugarman I've switched to `redis` from `redis-rails` are you're right, that added a bunch of dependencies we didn't need. Re the volatile/non-volatile, from the Slack discussion it sounds like the plan is to prioritise monitoring then suck it and see.

jsugarman (approved):
is there a ticket to look at incremental improvement to isolate volatile/cache data. otherwise 👍

PR #772, opened 2024-05-10
Account for RFIs when pulling in comments

ivanELEC:
## Description of change

 To hopefully fix[ this bug](https://mojdt.slack.com/archives/C05212WRK5J/p1715342111450759)

### DIFF
--- a/app/services/nsm/assessment_syncer.rb
+++ b/app/services/nsm/assessment_syncer.rb
@@ -20,11 +20,11 @@ module Nsm
     private
 
     def sync_overall_comment
-      comment_event = app_store_record['events'].select { _1['public'] && _1['event_type'] == 'decision' }
-                                                .max_by { DateTime.parse(_1['created_at']) }
-
       return if claim.status.in?(%w[submitted granted])
 
+      event_type = claim.send_back? ? 'send_back' : 'decision'
+      comment_event = app_store_record['events'].select { _1['public'] && _1['event_type'] == event_type }
+                                                .max_by { DateTime.parse(_1['created_at']) }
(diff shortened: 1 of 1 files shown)


app/services/nsm/assessment_syncer.rb
```diff
@@ -20,11 +20,11 @@ def call
     private
 
     def sync_overall_comment
-      comment_event = app_store_record['events'].select { _1['public'] && _1['event_type'] == 'decision' }
-                                                .max_by { DateTime.parse(_1['created_at']) }
-
       return if claim.status.in?(%w[submitted granted])
 
+      event_type = claim.status == 'provider_requested' ? 'send_back' : 'decision'
```
patrick-laa, line 25: Two things here. First, because status is an enum, you can call `claim.provider_requested?` instead of doing a string comparison. Second, I think _both_ send-back states use the `send_back` event type, i.e. "further_info" as well as "provider_requested"

ivanELEC: updated the above however blocked by SNYK due to braces

patrick-laa (approved):

spec/services/nsm/assessment_syncer_spec.rb
```diff
@@ -44,14 +44,35 @@
...
+              public: true,
+              details: { comment: 'Decision comment' }
+            },
+          ],
+          application: {}
+        }.deep_stringify_keys
+      end
+
+      it 'syncs the assessment_comment' do
+        expect(claim.assessment_comment).to eq 'Decision comment'
```
dwhenry, line 72: would have preferred this test to have a different comment here

dwhenry (approved):

PR #784, opened 2024-05-14
CRM457-1450: RBBA Bug Fix

imtiazAhmd:
## Description of change
The `uplift_applied?` method checks if any work item has a positive uplift.
currently it uses `filter_map` to extract the `uplift` of each work item, and checks if any of them are positive. 

Issue is if uplift is 0, `any?` returns true and all of our application was being assessed as High Risk.

Added `positive?` method to check if a number is greater than 0. So, method is only returns true if uplift is more than 0.

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-752)

## Notes for reviewer
Tested with Paper form PP2 that had initial assessment of Low risk from Caseworker.

### Before changes:
**Current Dev env - Wrong assessment**
Screenshot 2024-05-14 at 15 54 25
### After changes:
Current fixed env - Correct assessment
Screenshot 2024-05-14 at 15 32 33

### DIFF
--- a/app/services/risk_assessment/high_risk_assessment.rb
+++ b/app/services/risk_assessment/high_risk_assessment.rb
@@ -29,7 +29,7 @@ module RiskAssessment
     end
 
     def uplift_applied?
-      @items[:work_items].work_item_forms.any? { _1.uplift&.positive? }
+      @items[:work_items].work_item_forms.any?(&:apply_uplift)
     end
 
     def extradition?
--- a/app/services/risk_assessment/high_risk_assessment.rb
+++ b/app/services/risk_assessment/high_risk_assessment.rb
@@ -29,7 +29,7 @@ module RiskAssessment
     end
 
     def uplift_applied?
-      @items[:work_items].work_item_forms.filter_map(&:uplift).any?
+      @items[:work_items].work_item_forms.filter_map(&:uplift).any?(&:positive?)
     end
 
     def extradition?
--- a/app/services/risk_assessment/high_risk_assessment.rb
+++ b/app/services/risk_assessment/high_risk_assessment.rb
@@ -29,7 +29,7 @@ module RiskAssessment
     end
 
     def uplift_applied?
-      @items[:work_items].work_item_forms.filter_map(&:uplift).any?(&:positive?)
+      @items[:work_items].work_item_forms.any? { _1.uplift&.positive? }
     end
 
     def extradition?
--- a/spec/services/risk_assessment/high_risk_assessment_spec.rb
+++ b/spec/services/risk_assessment/high_risk_assessment_spec.rb
@@ -41,6 +41,15 @@ RSpec.describe RiskAssessment::HighRiskAssessment do
         end
       end
 
+      context 'when there is workitem without an uplift' do
+        let(:uplifted_claim) { build(:claim, :one_work_item) }
+
+        it do
+          expect(described_class.new(uplifted_claim)).not_to be_uplift_applied
+          expect(described_class.new(uplifted_claim).assess).not_to be_truthy
+        end
+      end
+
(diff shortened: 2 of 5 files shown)


app/services/risk_assessment/high_risk_assessment.rb
```diff
@@ -29,7 +29,7 @@ def enhancement_applied?
     end
 
     def uplift_applied?
-      @items[:work_items].work_item_forms.filter_map(&:uplift).any?
+      @items[:work_items].work_item_forms.filter_map(&:uplift).any?(&:positive?)
```
patrick-laa, line 32: I think this would work better as:

```ruby
 @items[:work_items].work_item_forms.any? { _1.uplift.positive? }
 ```
 
 That saves us iterating through the work array twice.

File: spec/services/risk_assessment/high_risk_assessment_spec.rb [hunk 2]
```diff
@@ -41,6 +41,15 @@
         end
       end
 
+      context 'when there is workitem without an uplift' do
```
patrick-laa, line 44: This spec is probably in the wrong place. This file is laid out rather unusually, but at the moment you'll see that you're inside the `context '#assess'`/`context 'returns true'` block, which shows a bunch of examples where the high risk assessment returns true. But this is an example where it returns false. So we probably want this inside the block that starts on line 72.

patrick-laa (commented):
What do you think about @jogilder's idea to add some high-level specs that use some real data to check that the risk is scored correctly? As we've seen, these contrived factory examples don't really verify anything. We'd probably want to put them in risk_assessment_scorer_spec.

app/services/risk_assessment/high_risk_assessment.rb
imtiazAhmd: Thanks @patrick-laa , I had to change a bit from what you suggested. as uplift can be nil, and I was getting 'positive?' doesn't have a method nil in my rspec. to fix it I have used lonely operator '&.positive?' so 'positive?' is only called if '_1.uplift' is not nil.

`@items[:work_items].work_item_forms.any? { _1.uplift&.positive? }`

spec/services/risk_assessment/high_risk_assessment_spec.rb
imtiazAhmd: Placed the spec in the `context '#assess'/ context 'returns false'` block. Updated the context claim with work item and existing spec for checking no uplift.

app/services/risk_assessment/high_risk_assessment.rb
```diff
@@ -29,7 +29,7 @@ def enhancement_applied?
     end
 
     def uplift_applied?
-      @items[:work_items].work_item_forms.filter_map(&:uplift).any?
+      @items[:work_items].work_item_forms.any? { _1.uplift&.positive? }
```
patrick-laa, line 32: Just thought of a bug in this. Repro:

1. create an NSM claim and check the 'enhanced rates claimed' checkbox on the 'reason for claim' screen
2. add a work item with an uplift applied
3. From the check answers screen go back to the 'reason for claim' screen and uncheck 'enhanced rates claimed'
4. Re-fill the form without editing your original work item
5. Submit the form.

As the uplift stored in the database is a positive number, this line of code will flag the claim as high risk even though, since uplift cannot be applied as enhanced rates are not claimed, it should not be.

To get around this, we can use the `apply_uplift` method on the work item form, like so:

```ruby
 @items[:work_items].work_item_forms.any?(&:apply_uplift)
 ```

This method already contains the logic we need to prevent false positives while also dealing with uplift being nil.

patrick-laa (approved):

PR #785, opened 2024-05-14
CRM7: Pull In Adjustment Data From Payload

ivanELEC:
## Description of change

 [CRM7: Pull In Adjustment Data From Payload](https://dsdmoj.atlassian.net/browse/CRM457-1475)

- [x] Pull in letters adjustment data
- [x] Pull in calls adjustment data
- [x] Pull in disbursments adjustment data
- [x] Pull in work item adjustment data
- [x] Update testing


app/services/nsm/assessment_syncer.rb
```diff
@@ -12,20 +12,79 @@ def initialize(claim, record:)
...
 
     def call
-      sync_overall_comment
+      case claim.status
+      when 'part_grant'
+        sync_letter_adjustments
+        sync_call_adjustments
+        sync_overall_comment
+        sync_work_items
+        sync_disbursements
```
dwhenry, line 21: given there is a lot of updates in this would be good to see it being done in a transaction

app/services/nsm/assessment_syncer.rb
```diff
@@ -12,20 +12,79 @@ def initialize(claim, record:)
...
+      claim.update(calls_adjustment_comment: record['adjustment_comment']) if record['adjustment_comment'].present?
+    end
+
+    def sync_work_items
+      work_items.each do |work_item|
+        record = WorkItem.find(work_item['id'])
+        record.update(allowed_time_spent: work_item['time_spent']) if work_item['time_spent_original'].present?
+        record.update(allowed_uplift: work_item['uplift']) if work_item['uplift_original'].present?
+        record.update(adjustment_comment: work_item['adjustment_comment']) if work_item['adjustment_comment'].present?
+      end
```
dwhenry, line 58: feel like these could be simplified, but good for now.

app/services/nsm/assessment_syncer.rb
```diff
@@ -12,20 +12,79 @@ def initialize(claim, record:)
...
+    end
+
+    def letters
+      app_store_record['application']['letters_and_calls'].select { _1['type']['value'] == 'letters' }
+                                                          .first
+    end
+
+    def calls
+      app_store_record['application']['letters_and_calls'].select { _1['type']['value'] == 'calls' }
+                                                          .first
```
dwhenry, line 79:
```
app_store_record['application']['letters_and_calls'].detect { _1['type']['value'] == 'calls' }
```

dwhenry (approved):
good starting point IMO

app/services/nsm/assessment_syncer.rb
ivanELEC: done this, not sure how transaction works under the hood when it comes to updating associated records so I've ensured the transaction is happening relative to the claim by selecting work items and disbursements via the claim.

dwhenry (approved):

PR #789, opened 2024-05-15
Modsec change to allow big data chunck

dwhenry:
## Description of change
Modsec changes


jogilder (changes requested):
Can we remove `SecRequestBodyLimit` and check what we actually need for `SecRequestBodyNoFilesLimit`?

patrick-laa (approved):

PR #803, opened 2024-05-17
Differentiate between app and worker deployments in selector labels

ivanELEC:
## Description of change

 [Unexpected Scaling of Pods](https://dsdmoj.atlassian.net/browse/CRM457-1508)

Trying to add an extra selector label: app - which will be used to differentiate worker deployment pods from app deployment pods - in an attempt to fix the issue with our autoscaler scaling our pods up when we seemingly haven't breached limitations


jogilder (changes requested):
I would be tempted to keep the labels and just either remove the selectorlabels in the deployments or add in something that disambiguates the selectorlabels?
https://github.com/kubernetes/kubernetes/issues/120875#issuecomment-2062615449

PR #808, opened 2024-05-20
Log the auth hash

ivanELEC:
[This is to understand why we are getting errors trying to login with credentials that should be valid
](https://mojdt.slack.com/archives/C060NFMUGP5/p1716197228141829?thread_ts=1713270186.093129&cid=C060NFMUGP5)

REVERT THIS COMMIT TO PREVENT PII IN LOGS

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -419,10 +419,6 @@ workflows:
             branches:
               ignore:
                 - main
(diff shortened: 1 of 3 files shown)


dwhenry (approved):
Are there any security concerns with this??? do we want to limit this to dev servers?

PR #813, opened 2024-05-20
Increase Clam AV and Application Container Memory Limits to 4GB

ivanELEC:
## Description of change

 [For testing pod resource usage](https://dsdmoj.atlassian.net/browse/CRM457-1508)


helm_deploy/templates/deployment.yaml
```diff
@@ -38,12 +38,7 @@ spec:
               containerPort: 3310
               protocol: TCP
           resources:
-            limits:
-              cpu: 500m
-              memory: 3Gi
-            requests:
-              cpu: 10m
-              memory: 1Gi
+            {{- toYaml .Values.resources | nindent 12 }}
```
jogilder, line 41: would we want a separate resources section for clamav in case we wanted them to be different?

ivanELEC: yeah probably a good shout - even though want them to be the same for now to try and get some form of predictablility with the threshold calculations, I assume this is the type of thing we would want to be able to easily edit

PR #816, opened 2024-05-21
Crm457 1459/adjustments calculations

dwhenry:
## Description of change
Refactor existing code to extract cost calculations into a concern which can then be used directly in the various models (as well as forms) and allow for cleaner calcultions of costs where required (i.e. no need to load the forms)

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1459)


ivanELEC (approved):
LGTM

PR #822, opened 2024-05-22
Crm457 1360 - MONITORING - Exporting metrics for Redis Queues and Sidekiq

ivanELEC:
## Description of change

 [MONITORING - Exporting metrics for Redis Queues and Sidekiq](https://dsdmoj.atlassian.net/browse/CRM457-1360)

- [x] Add Prometheus Client and Metrics Endpoint
- [x] Configure Service to be able to link to ServiceMonitor
- [x] Add ServiceMonitor to helm deploy 

- [Service Monitor and Network Policy to allow out Service to connect to Service Monitor](https://github.com/ministryofjustice/cloud-platform-environments/pull/23143)
- The fullName value from the helper template is used, this takes it's value from the fullNameOverride in our values files - for this reason we don't have to worry about branch builder services linking to Prometheus


helm_deploy/templates/deployment.yaml
```diff
@@ -11,6 +11,7 @@ spec:
   selector:
     matchLabels:
       {{- include "laa-submit-crime-forms.selectorLabels" . | nindent 6 }}
+      app: {{ include "laa-submit-crime-forms.fullname" . }}
```
jsugarman, line 14: 👍 especially as worker deployment is already doing this but with `-worker` appended

jogilder: Maybe suffix with web here as well?

helm_deploy/templates/service.yaml
```diff
@@ -13,3 +13,4 @@ spec:
       name: http
   selector:
     {{- include "laa-submit-crime-forms.selectorLabels" . | nindent 4 }}
+    app: {{ include "laa-submit-crime-forms.fullname" . }}
```
jogilder, line 16: I am not sure how this one works - whether it needs the `-web` or not? Prior to this it worked when we had `app:web`.
@jsugarman any ideas?

ivanELEC: this label renders to "laa-submit-crime-forms-app" already so this would turn to laa-submit-crime-forms-app-web
which feels more confusing to me. But yes, if we add "-web" to the deployment, we'd need to add "-web" to the service label (I tried not matching them and I got errors). I think the reason it worked before was that the service labels were still a subset of the deployment labels

ivanELEC:
> See comments. Will you do a separate PR to add in the specific metrics for sidekiq?

Yes, this PR is just to get the first bits setup so I can ensure there is connectivity to prometheus

helm_deploy/templates/deployment.yaml
```diff
@@ -19,7 +20,7 @@ spec:
       {{- end }}
       labels:
         {{- include "laa-submit-crime-forms.selectorLabels" . | nindent 8 }}
-        app: web
+        app: {{ include "laa-submit-crime-forms.fullname" . }}
```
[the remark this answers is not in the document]
ivanELEC: my preference would be to use the file name instead of the label to differentiate as the names behind these variables already communicate what they are: 
laa-submit-crime-forms-app
laa-submit-crime-forms-app-worker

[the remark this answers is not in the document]
ivanELEC: we could maybe name the deployment.yaml file deployment-app or deployment-web?

helm_deploy/templates/service.yaml
jsugarman: for the deployments we never bothered with a `-web` or `-app` in apply. just ` `-worker` suffix in the other deployment.

For the service Apply we have a service-metrics file with  `-metrics` suffix on the label
https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/main/helm_deploy/apply-for-legal-aid/templates/service-metrics.yaml#L17C6-L17C69

plus a service-monitor file that uses matches it

https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/main/helm_deploy/apply-for-legal-aid/templates/service-monitor.yaml

BUT the prometheus rules in CP do not explicitly use these so am not sure how it all wires together and we use a deprecated prometheus gem so not necessarily a good model to follow

jsugarman: from the CP docs and Apply's implementation it sounds like we should create a service-metrics.yml separate from service.yml and use the `-metrics` label for that. Then add a service-monitor.yml that matches on the `-metrics` service label. Then we also need a networks-policy.yml to allow prometheus to scrap the `/metrics` endpoint of the app.

ivanELEC: yeah like Jo said in Slack - Apply is using metrics in a different way. They have Prometheus Exporter and [deploy a separate container with prometheus exporter running](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/main/helm_deploy/apply-for-legal-aid/templates/deployment_metrics.yaml). Whereas we use promethues client and add that client as middleware to our existing application so the Service of the application and metrics is shared. 

I think it depends on whether we see benefits in doing things this way or that way

helm_deploy/templates/deployment.yaml
jogilder: Agreed to leave this out.

helm_deploy/templates/deployment.yaml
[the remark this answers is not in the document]
jogilder: No need to do this

helm_deploy/templates/service.yaml
jogilder: Discussed and resolved

jsugarman:
see https://github.com/ministryofjustice/cloud-platform-environments/pull/23269 for privs needed by CircleCI

dwhenry (commented):
I have no idea

PR #823, opened 2024-05-22
Crm457 1459/adjustments calculations 2

dwhenry:
## Description of change
Add Work item adjustments screen

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1459)

Screenshot 2024-05-22 at 16 17 40

screen shot (slightly different from prototype) confirmed with @mary-bowden


app/views/nsm/steps/view_claim/show.html.erb
```diff
@@ -21,6 +21,11 @@
         <li class="moj-sub-navigation__item">
           <%= link_to t('.tabs.claimed_costs'), nsm_steps_view_claim_path(@claim, section: 'claimed_costs'), class: 'moj-sub-navigation__link', 'aria-current': (@section == :claimed_costs ? "page" : nil) %>
         </li>
+        <% if @claim.status == 'part_grant' %>
```
jsugarman, line 24: the statuses have enums you could use but either way

```suggestion
        <% if @claim.part_grant? %>
```

spec/system/nsm/view_claim_spec.rb
```diff
@@ -51,6 +51,20 @@
     end
   end
 
+  # TODO: update this to have adjusted values as well...
```
jsugarman, line 54: **nit** Is there a ticket for this todo

jsugarman (approved):
lots going on in the views but maybe could be refactored another time 👍

app/views/nsm/steps/view_claim/show.html.erb
dwhenry: let me look at this as I tried it and it didn't work... :(

PR #824, opened 2024-05-23
bump braces 3.0.3

jsugarman:
## Description of change
Bump braces 3.0.3

- Bump braces 3.0.2 to 3.0.3
- Remove snyk ignore

This should fix the CVE we currently ignore

## Notes for reviewer

If Snyk scan passes when the specific CVE is unignored then it is good to go

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,10 +1,11 @@
 # Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
 version: v1.25.0
 # ignores vulnerabilities until expiry date; change duration by modifying expiry date
-ignore:
-  SNYK-JS-BRACES-6838727:
-    - '*':
-        reason: no fix; no-impact
-        expires: 2099-01-01T00:00:00.000Z
-        created: 2024-05-13T21:14:53.339Z
+ignore: {}
+  # FIXED by bumping to braces@3.0.3
+  # SNYK-JS-BRACES-6838727:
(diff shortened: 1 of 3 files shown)


.snyk
```diff
@@ -1,10 +1,11 @@
...
 version: v1.25.0
 # ignores vulnerabilities until expiry date; change duration by modifying expiry date
-ignore:
-  SNYK-JS-BRACES-6838727:
-    - '*':
-        reason: no fix; no-impact
-        expires: 2099-01-01T00:00:00.000Z
-        created: 2024-05-13T21:14:53.339Z
+ignore: {}
+  # FIXED by bumping to braces@3.0.3
```
patrick-laa, line 5: Is it worth leaving this commented out as an example of the syntax for ignoring CVEs for people without the snyk CLI?

patrick-laa (approved):

.snyk
patrick-laa: (That may have been what you were planning anyway - I got half way through typing a comment saying we should delete this as it's no longer useful then changed my mind)

jsugarman: @patrick-laa To leave as example was my thinking, but happy either way?

jsugarman: and the CLI took me over half hour to install due to some aws-cpp library bump (possibly a one off) and so was a pain to get up and running locally 😄

PR #846, opened 2024-05-29
Increase Clamav limits

jsugarman:
## Description of change
Increase CLAMAV container resource limits

To prevent HPA from spinning up more pods
to handle usage.


jsugarman:
hpa did scale up regardless so lets not do this

PR #849, opened 2024-05-30
Crm457 1541/missing travel reason

dwhenry:
## Description of change
Fix flow when changing from detained to not detained requiring a travel_cost_reason

* Primary quote marked as in process in task list
* click primary quote link will direct user to travel cost screen when it is invalid

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1541)

### DIFF
--- a/app/presenters/prior_authority/tasks/primary_quote.rb
+++ b/app/presenters/prior_authority/tasks/primary_quote.rb
@@ -41,7 +41,7 @@ module PriorAuthority
       end
 
       def travel_costs_complete?
-        # return true here as no record is the same as no data wchih is true
+        # return true here as no record is the same as no data which is true
         return true unless record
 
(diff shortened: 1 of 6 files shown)


app/presenters/prior_authority/tasks/primary_quote.rb
```diff
@@ -37,6 +40,14 @@ def first_steps_completed?
           PriorAuthority::Steps::ServiceCostForm.build(record, application:).valid?
       end
 
+      def travel_costs_complete?
+        # return true here as no record is the same as no data wchih is true
```
jsugarman, line 44:
```suggestion
        # return true here as no record is the same as no data which is true
```

jsugarman (approved):

PR #870, opened 2024-06-05
Fix Alternative Quotes Incorrect Data

ivanELEC:
## Description of change

 [CRM4 Caseworker: Application cannot be opened potentially due to number of uploads](https://dsdmoj.atlassian.net/browse/CRM457-1567)

## Notes for reviewer

*Misleading bug, the real issue is that there are issues with the quotes data and the way that data is consumed by caseworker app 

This PR uses the primary quote item type and cost type for alternative quotes when constructing the payload because these are defined as such in the business process

This PR is to address[ this error ](https://ministryofjustice.sentry.io/issues/5453466200/?alert_rule_id=14899609&alert_type=issue&notification_uuid=56ce2154-7606-407b-91c5-e19f3f0e3438&project=4506020469473280&referrer=slack) which seems to be happening because the cost type isn't being set on alternative quotes

### DIFF
--- a/app/services/submit_to_app_store/prior_authority/quote_payload_builder.rb
+++ b/app/services/submit_to_app_store/prior_authority/quote_payload_builder.rb
@@ -22,7 +22,7 @@ class SubmitToAppStore
 
       def primary_item_type
         form = ::PriorAuthority::Steps::ServiceCostForm.build(@application.primary_quote, application: @application)
-        form.cost_type
+        form.item_type
       end
 
       def document(quote)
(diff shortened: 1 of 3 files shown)


app/services/submit_to_app_store/prior_authority/quote_payload_builder.rb
```diff
@@ -3,20 +3,25 @@ module PriorAuthority
...
-            cost_type: form.cost_type,
-            item_type: form.item_type,
             document: document(quote)
-          )
+          ).merge(primary_quote_attributes)
         end
       end
 
+      def primary_quote_attributes
+        form = ::PriorAuthority::Steps::ServiceCostForm.build(@primary_quote, application: @application)
```
patrick-laa, line 18: It might make sense to memoize this method, otherwise you're having to recalculate the cost type and item type for every quote.

patrick-laa (approved):

PR #875, opened 2024-06-06
Crm457 1505: Provider login: deactivating office causes ERR_TOO_MANY_REDIRECTS

ivanELEC:
## Description of change

 [Provider login: deactivating office causes ERR_TOO_MANY_REDIRECTS](https://dsdmoj.atlassian.net/browse/CRM457-1505)

### DIFF
--- a/app/controllers/providers/omniauth_callbacks_controller.rb
+++ b/app/controllers/providers/omniauth_callbacks_controller.rb
@@ -28,6 +28,7 @@ module Providers
 
     def check_provider_is_enrolled
       gatekeeper = Providers::Gatekeeper.new(auth_hash.info)
+      Rails.logger.info auth_hash.info.office_codes
       return if gatekeeper.provider_enrolled?
 
       Rails.logger.warn "Not enrolled provider access attempt, UID: #{auth_hash.uid}, " \
--- a/config/feature_flags.yml
+++ b/config/feature_flags.yml
@@ -1,7 +1,7 @@
 feature_flags:
   omniauth_test_mode:
     local: <%= ENV.fetch("OMNIAUTH_TEST_MODE", false) %>
-    development: true
+    development: false
     uat: false
     production: false # must never be enabled in the live service
   postal_evidence:


app/controllers/providers/omniauth_callbacks_controller.rb
```diff
@@ -5,11 +5,17 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
-      sign_in_and_redirect(
-        provider, event: :authentication
-      )
+      if Provider.active_offices?(auth_hash)
+        provider = Provider.from_omniauth(auth_hash)
+
+        sign_in_and_redirect(
+          provider, event: :authentication
+        )
+      elsif offices_in_list?
```
jsugarman, line 14: this feels a bit odd but i think i understand it

```
if they have any active office (or NOT any inactive offices)
  sign them in 👍 
elsif auth hash has enrolled offices (or email enrolled or ALL enabled)
  redirect to "you need an active office code" <--- This feels odd
else
  redirect to "not enrolled yet" 👍 
end 
```

Given
- inactive offices BBBBBB and CCCCCC
- enrolled offices AAAAAA, BBBBBB and CCCCCC

if user A has AAAAAA, BBBBBB and CCCCC then they get signed in
if user B has  BBBBBB and CCCCCC then they fall through but because they are enrolled with inactive office codes they get redirect to the new page.
if user D has DDDDDD they fall through and get not enrolled page

When we switch to ALL for crm5 and unenroll AAAAAA then this should still work as expected user A 👍 and should still redirect B 👍 , but not D 👍 

I think this will work ok therefore but perhaps change the name of `offices_in_list?` to `[inactive_]office_enrolled?`

app/controllers/providers/omniauth_callbacks_controller.rb
```diff
@@ -35,5 +41,9 @@ def check_provider_is_enrolled
 
       redirect_to laa_msf.not_enrolled_errors_path
     end
+
+    def offices_in_list?
+      true if Providers::Gatekeeper.new(auth_hash.info).provider_enrolled?
```
jsugarman, line 46:
```suggestion
      Providers::Gatekeeper.new(auth_hash.info).provider_enrolled?
```

app/controllers/providers/omniauth_callbacks_controller.rb
ivanELEC: yeah fair point, I did find that method name a bit awkward. There's probably a cleaner way to do all of this but I didn't want to redesign how it's currently done for no good reason. I've updated the method name

PR #876, opened 2024-06-06
CRM457-1590: Enable simple onboarding of ALL providers per service

jsugarman:
## Description of change
Enable simple onboarding of ALL providers per service

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1590)

Current onboarding approach requires adding all office codes to the provider app and specifying which service(s) they can use. This will become onerous as we onboard more and pointless when we go to public
beta and ALL crime providers (in LAA portal) will be allowed access.

This enables us to switch on a specific form type/service for all providers with the portal link, sa below (for crm5)

```
# gatekeeper.yml
production:
  office_codes:
    ALL: ['crm5']
```

## Notes for reviewer
You can test locally by adding below config and try to access with a test user mock login that normally would have no access.

```
# gatekeeper.yml
localhost:
  office_codes:
    ALL: ['crm5']
```

### DIFF
--- a/app/services/providers/gatekeeper.rb
+++ b/app/services/providers/gatekeeper.rb
@@ -13,7 +13,15 @@ module Providers
     end
 
     def provider_enrolled?(service: ANY_SERVICE)
-      email_enrolled? || office_enrolled?(service:)
+      email_enrolled? || all_enrolled?(service:) || office_enrolled?(service:)
+    end
+
+    def all_enrolled?(service: ANY_SERVICE)
+      if service == ANY_SERVICE
+        allowed_office_codes.fetch(:ALL, []).any?
+      else
+        allowed_office_codes.fetch(:ALL, []).include?(service.to_s)
+      end
     end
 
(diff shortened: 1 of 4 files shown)


app/services/providers/gatekeeper.rb
```diff
@@ -13,7 +13,15 @@ def initialize(auth_info)
     end
 
     def provider_enrolled?(service: ANY_SERVICE)
-      email_enrolled? || office_enrolled?(service:)
+      email_enrolled? || all_enrolled?(service:) || office_enrolled?(service:)
+    end
+
+    def all_enrolled?(service: ANY_SERVICE)
+      if service == ANY_SERVICE
+        allowed_office_codes.key?(:ALL)
```
dwhenry, line 21:
```
allowed_office_codes.fetch(:ALL, []).any?
```
this just guards against the situation when  the yaml has:
```
localhost:
  office_codes:
    ALL: []
```

PR #878, opened 2024-06-07
CRM457-1592: Separate gatekeeper config per service

jsugarman:
## Description of change
Separate gatekeeper config per service

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1592)

Current onboarding approach requires adding all office codes to the provider app and specifying what service/form-type they can use:

```yaml
1A11A1: ['crm5']
```

This mechanism does not scale well when we need to onboard hundreds/thousands for a specific form type and not others, then onboard some of those for a different form type. Dev discussion agreed to use separate lists for each form type so that it can be managed per form type in isolation via copy pasting of a tranche.

## Notes for reviewer
Can be tested locally or on the DEV branch. The existing tests/dummy users work as expected when logging in with each of them on DEV.


config/gatekeeper_crm5.yml
```diff
@@ -0,0 +1,79 @@
...
+    - 2N232W
+
+    # First CRM5 onboarding tranche - 29/05/2024
+    - 2L672A
+    - 2M259N
+
+    # Second CRM5 onboarding tranche - 06/06/2024
+    - 2Q553U
+
+    # Second CRM5 onboarding tranche - 06/06/2024
```
jsugarman, line 67: remove

dwhenry (approved):

PR #892, opened 2024-06-11
CRM457-1457: Export metrics to prometheus

patrick-laa:
## Description of change

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1457)


Gemfile
```diff
@@ -17,7 +17,7 @@ gem 'logstasher', '~> 2.1'
 gem 'oauth2', '~> 2.0'
 gem 'pagy', '~> 8.4.2'
 gem 'pg', '~> 1.5'
-gem 'prometheus-client'
+gem 'prometheus_exporter'
```
jogilder, line 20: why does this not have prometheus-client but the gem lock file does?

Gemfile.lock
```diff
@@ -312,6 +312,8 @@ GEM
       optimist (~> 3.0)
     pg (1.5.6)
     prometheus-client (2.1.0)
```
jogilder, line 314: still has prometheus-client - is this a dependency?

jogilder (commented):
see comment below

Gemfile.lock
patrick-laa: exactly - sidekiq_prometheus has a dependency on prometheus-client

Gemfile
jogilder: prometheus client for sidekiq work in another ticket hence needed here.

jogilder (approved):
LGTM

PR #896, opened 2024-06-11
Crm457 1612 fix translation

ivanELEC:
## Description of change

 [CRM4 - Content Change - Service Type - Translation (documents)](https://dsdmoj.atlassian.net/browse/CRM457-1612)
 

-  Add Cost Multiplier To Service Type Rules
-  Default Cost Multiplier To 1 If Not Provided In Initiailization
-  Set Cost Multiplier To 0.001 For Translation (documents)
-  Use Cost Multiplier In Service Cost Form
-  Add Cost Item to Service Type Rule For Alternate Qualifiers (such as thousand words instead of words)
- Update Tests

### DIFF
--- a/app/forms/prior_authority/quote_cost_validations.rb
+++ b/app/forms/prior_authority/quote_cost_validations.rb
@@ -11,7 +11,7 @@ module PriorAuthority
 
       with_options if: :per_item? do
         validates :items, item_type_dependant: { pluralize: true }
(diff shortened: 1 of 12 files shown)


patrick-laa:
For the sake of expediency it may be worth addressing this in a separate PR, but error messages will need to be updated to use "1000 words" here:
Screenshot 2024-06-12 at 12 35 24

PR #931, opened 2024-06-20
CRM457-1649: Only display prior authority for other type disbursements

jsugarman:
## Description of change
Only display prior authority for "other" type disbursements

Prior authority question is not asked for mileage disbursements

### Before changes:
:boom:

### After changes:

Screenshot 2024-06-20 at 14 49 45


dwhenry (commented):
Should we make a chnage so we only ask it then as well? as we currently always ask the question. :)

jsugarman:
> Should we make a chnage so we only ask it then as well? as we currently always ask the question. :)

That question was made conditional a while ago now. so we don't ask it for mileage disbursements but this change was missed then. And a similar change has gone in on assess already to handle nil prior_authority.

patrick-laa (approved):

PR #935, opened 2024-06-21
CRM457-1408: Custom Modsec errors attempt 2

patrick-laa:
## Description of change
THe problem with our initial approach is it caused our custom error page to be served for _all_ 403s, including ones produced by our app. This approach gives modsec errors status code 423, which is obscure enough we can be more confident our system will not return that code in any other circumstance. Therefore a blanket "show the modsec error page any time the ingress reports a 423 response" is unlikely to cause problems.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1408)

Note that to enable testing, I've temporarily enabled allowing branches to deploy the custom error service. Before merge I'll remove that so that it's only main that deploys it.


helm_deploy/templates/custom-error-pages.yaml
```diff
@@ -0,0 +1,107 @@
...
...1px solid rgba(0,0,0,0);border-radius:50%}@media(min-width: 61.25em){.app-swatch{display:inline-block;position:static;margin-top:6px;margin-right:10px;margin-bottom:7px;vertical-align:middle}}.app-swatch-border{border-color:#b1b4b6}.app-colour-list-column--name{font-weight:normal;text-align:left}@media(min-width: 61.25em){.app-colour-list-column--name{width:50%}}@media(min-width: 61.25em){.app-colour-list-column--colour{width:15%}}.app-colour-list-column--colour code{padding:0;color:#0b0c0c;background:none}@media(min-width: 61.25em){.app-colour-list-column--notes{width:35%}}pre{max-width:38em;margin:0;padding:20px;overflow:auto;border:1px solid #b1b4b6;background-color:#f8f8f8;font-size:16px;margin-bottom:15px}@media(min-width: 40.0625em){pre{margin-bottom:20px}}li code,td code,p code{padding:0 5px;color:#d13118;background-color:#f3f2f1}.app-prose-scope strong,.app-prose-scope b{font-weight:700}.app-prose-scope a:focus code{color:#0b0c0c;background:rgba(0,0,0,0)}.app-prose-scope pre+h2{padding-top:20px}.app-prose-scope pre+h3,.app-prose-scope pre+h4{padding-top:10px}.app-prose-scope>:last-child{margin-bottom:0}.app-whats-new{border-bottom:1px solid #b1b4b6;background-color:#f8f8f8}
```
dwhenry, line 20: I don't like this....

it will be difficult to maintain if/when the designs change....

options (bad):
* turning this into a erb and rendering the page as part of the deploy 
* something else

dwhenry (approved):
understand why, but still don;t like...

PR #938, opened 2024-06-24
Delete redis pvc on merge into main

jsugarman:
## Description of change
Delete redis pvc on merge into main

Currently have 260 of these sitting around. This should
help future tidy up

## Notes for reviewer
```sh
kubectl get pvc | grep redis | wc -l
```


patrick-laa (approved):

jsugarman:
the failure of the GHA on this branch was due to redis being in the name and the fact that when  the redis chart detects this it calls the pvc master-0, not redis-master-0. its pretty rare but we could emulate this logic but does not seem worth it

PR #941, opened 2024-06-25
CRM457-1682 - Provider: Incorrect Client Email Validation

ivanELEC:
## Description of change

[Provider: Incorrect Client Email Validation](https://dsdmoj.atlassian.net/browse/CRM457-1682)


jsugarman (commented):
We should probably amend, create/extend any email validation specs to regression test this.

jsugarman (approved):
💯

PR #964, opened 2024-07-02
CRM457-1705: Do not show links to documents on PDFs

patrick-laa:
## Description of change
- PDF download was breaking because link generation to uploaded documents was failing
- It shouldn't have been failing, but also the PDFs shouldn't have links inside them anyway (because those links look ugly and will not work as they link to authenticated URLs)
- So this removes links to documents from PDFs

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1705)


jsugarman (approved):
is there not any integration testing for this stuff 😬

PR #984, opened 2024-07-09
CRM457-1672: Try to avoid migration races

patrick-laa:
## Description of change
Sometimes 2 pods run migrations at the same time, one loses the race, and we get error messages on Slack even though nothing bad has happened. This should prevent those false positives.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1672)

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -1,0 +1,24 @@
+namespace :db do
+  namespace :preparation do
+    desc "Run db:prepare and retry if 2-pods-running-this-at-once issues encountered"
+    task run_with_retry: :environment do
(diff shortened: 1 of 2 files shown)


elken (approved):
Only remark is do we want to make the 3 an env var? Same for the other 2 so I won't duplicate the potential discussion lol

robinsk1:
can the pods not be built sequentially so that wouldn't happen?

patrick-laa:
> Only remark is do we want to make the 3 an env var? Same for the other 2 so I won't duplicate the potential discussion lol

PR #994, opened 2024-07-11
CRM457-1744: Allow downloading files with weird chars in them

patrick-laa:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1744)

### DIFF
--- a/spec/system/prior_authority/view_applications_spec.rb
+++ b/spec/system/prior_authority/view_applications_spec.rb
@@ -164,7 +164,7 @@ RSpec.describe 'View reviewed applications' do
             information_requested: 'Tell me more',
             information_supplied: 'More info',
             created_at: 1.day.ago,
-            supporting_documents: [build(:supporting_document, file_name: 'evidence.pdf')])
+            supporting_documents: [build(:supporting_document, file_name: 'evidence–with-weird-char.pdf')])
     end
 
     it 'shows uodate details' do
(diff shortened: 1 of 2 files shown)


spec/system/prior_authority/view_applications_spec.rb
```diff
@@ -175,8 +175,11 @@
     end
 
     it 'lets me download my uploaded file' do
-      click_on 'evidence.pdf'
+      click_on 'evidence–with-weird-char.pdf'
       expect(page).to have_current_path(%r{/test_path})
+      expect(page.driver.request.params['response-content-disposition']).to eq(
+        'attachment; filename="evidencewith-weird-char.pdf"'
```
jsugarman, line 181: it looks like its just removing the _first_ hyphen here - what is the weird char?

patrick-laa: @jsugarman annoyingly GitHub doesn't highlight it, but my IDE does - that first hyphen isn't a hyphen, it's an "en dash", which isn't included in ISO-8859-1 and therefore isn't accepted by S3.

Screenshot 2024-07-12 at 11 57 48

patrick-laa: I've added a comment to make clearer what's going on.

jsugarman (approved):

PR #996, opened 2024-07-11
put update_and_advance method in transaction to lock application whil…

ivanELEC:
## Description of change

[Provider: Fast Submit Clicks Can Wrongly Create Multiple Records](https://dsdmoj.atlassian.net/browse/CRM457-1746)

## Notes for reviewer
This is just part 1 of the bug fix - also need to lock button once form is submitted, this can be done in separate PR

### DIFF
--- a/gems/laa_multi_step_forms/app/controllers/steps/base_step_controller.rb
+++ b/gems/laa_multi_step_forms/app/controllers/steps/base_step_controller.rb
@@ -31,25 +31,27 @@ module Steps
 
     # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
     def update_and_advance(form_class, opts = {})
-      hash = permitted_params(form_class).to_h
-      record = opts.fetch(:record, current_application)
-      @form_object = form_class.new(
-        hash.merge(application: current_application, record: record)
-      )
+      current_application.transaction do
+        hash = permitted_params(form_class).to_h
+        record = opts.fetch(:record, current_application)
+        @form_object = form_class.new(
+          hash.merge(application: current_application, record: record)
+        )
 
-      if params.key?(:commit_draft)
-        # Validations will not be run when saving a draft
-        @form_object.save!
-        redirect_to opts[:after_commit_redirect_path] || nsm_after_commit_path(id: current_application.id)
-      elsif params.key?(:reload)
-        reload
-      elsif params.key?(:save_and_refresh)
-        @form_object.save!
-        redirect_to_current_object
-      elsif @form_object.save
-        redirect_to decision_tree_class.new(@form_object, as: opts.fetch(:as)).destination, flash: opts[:flash]
-      else
-        render opts.fetch(:render, :edit)
+        if params.key?(:commit_draft)
+          # Validations will not be run when saving a draft
+          @form_object.save!
+          redirect_to opts[:after_commit_redirect_path] || nsm_after_commit_path(id: current_application.id)
+        elsif params.key?(:reload)
+          reload
+        elsif params.key?(:save_and_refresh)
+          @form_object.save!
+          redirect_to_current_object
+        elsif @form_object.save
+          redirect_to decision_tree_class.new(@form_object, as: opts.fetch(:as)).destination, flash: opts[:flash]
+        else
+          render opts.fetch(:render, :edit)
+        end
       end
     end
     # rubocop:enable Metrics/MethodLength, Metrics/AbcSize


app/javascript/handle-forms.js
```diff
@@ -0,0 +1,21 @@
+import $ from 'jquery'
+
+window.$ = $
+
+$(function() {
+  $('button[type="submit"]').on("click", function(e) {
```
patrick-laa, line 6: We should trigger on a form submit, not a button click, otherwise submitting a form with the enter key won't do anything.

app/javascript/handle-forms.js
```diff
@@ -0,0 +1,21 @@
...
+    e.preventDefault()
+    let form = $(this).closest("form")
+    let formButtons = form.find('button[type="submit"]')
+    formButtons.each(function(){
+      $(this).attr("disabled", true)
+    })
+
+    let buttonAction = $(this).attr("name")
+    if(buttonAction){
+      form.append(`<input type="hidden" name="${buttonAction}" value="" >`)
```
patrick-laa, line 16: I don't think I understand what this is doing - can you add a comment to explain?

gems/laa_multi_step_forms/app/controllers/steps/base_step_controller.rb
```diff
@@ -29,30 +29,36 @@ def decision_tree_class
...
 
-    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
     def update_and_advance(form_class, opts = {})
       hash = permitted_params(form_class).to_h
       record = opts.fetch(:record, current_application)
       @form_object = form_class.new(
         hash.merge(application: current_application, record: record)
       )
 
+      current_application.transaction do
```
patrick-laa, line 39: I may be wrong here, but transactions aren't locks - a transaction simply means if one statement fails, all the rest will be rolled back. We want locking to ensure that no statements are executed while another process is also executing the statements.

robinsk1: `current_application.with_lock` ?

app/javascript/handle-forms.js
ivanELEC: Added: 
>       this code ensures the button's submission type is appended to params
>       so that the BaseStepController still has this available to know
>       whether the form should commit_draft/save_and_refresh/reload

gems/laa_multi_step_forms/app/controllers/steps/base_step_controller.rb
ivanELEC: yes you are correct - I had it in my ActiveRecord notes that transaction does both of those things and didn't double check the official docs :/ 
that said - in this case maybe makes sense to do both as this would ensure that when a form performs multiple actions say changing the application and an associated record on save, it will roll back all changes on error

app/javascript/handle-forms.js
ivanELEC: true - huge oversight on my part, not very accessible at all!

gems/laa_multi_step_forms/app/controllers/steps/base_step_controller.rb
patrick-laa: I think locking always creates a transaction, it's just that transactions don't always create locks. c.f. https://blog.saeloun.com/2022/03/23/rails-7-adds-lock_with/

app/javascript/handle-forms.js
ivanELEC: wait the strange thing is though - this works without triggering on the form submit and triggering on the button click and I don't know why

ivanELEC: [click ](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) covers Enter key and the form has [implicit submission](https://www.w3.org/TR/2011/WD-html5-20110525/association-of-controls-and-forms.html#implicit-submission) on enter when focused on any non submit button part of the form

patrick-laa: Fair enough!

elken: Does that still handle touch events though? eg on a phone with no fields focused

gems/laa_multi_step_forms/app/controllers/steps/base_step_controller.rb
ivanELEC: I see! I'll remove the transaction block then

ivanELEC: (done)

patrick-laa (approved):

PR #1002, opened 2024-07-16
Experiment: With minimal modsec config, is phase 4 still broken?

patrick-laa:
Yes. https://crm457-1721-b-expl-nscc-provider-dev.cloud-platform.service.justice.gov.uk/assets/file-upload-a2e2c5e8.js.map triggers a false positive on rule 952100, and modsec thinks it is blocking it and replacing the status with a 403, but the body is still returned to the HTTP requester. The status code is coming across as either 200 or ERR_HTTP2_PROTOCOL_ERROR on the client-side, and the kibana logs list it as a 500.


patrick-laa:
This is to enable investigation of https://github.com/ministryofjustice/cloud-platform/issues/5906

PR #1006, opened 2024-07-17
CRM457-1361: PII out of logs, provider IDs in

patrick-laa:
## Description of change
At the moment we're logging PII to kibana, which we should stop
We're also not identifying which logged in user is taking each action, which we should start.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1361)


elken (commented):
If the form submission fails for one reason or another, how would we verify what the user submitted in the case this is merged?

patrick-laa:
@elken good question. I don't think we would be able to do it once this is merged. But I think our current ability to do that is an accidental happy consequence of us doing the wrong thing (having PII in logs), and I think it is more important for us to stop doing the wrong thing than retain the nice side effect. I'd change my stance if we could point to a requirement of our system to be able to reconstruct submitted information in cases where processing fails. (I guess this would be a bigger deal if, e.g., providers spent an hour filling in all their work items in a single page and then something went wrong when they clicked submit and they lost an hour of data/time).

elken (approved):
True, I don't have any other gripes then just something to be aware of.

If we encounter an issue where this would have been useful, we can look at Sentry events or something

PR #1011, opened 2024-07-23
[CRM457-1738] CRM7 Provider: Claimed costs page - changes to cost schedule display

imtiazAhmd:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1738)


app/views/nsm/steps/view_claim/work_items.html.erb
```diff
@@ -8,67 +8,76 @@
...
+      'line_item' => false,
+      'item' => false,
+      'date' => false,
+      'fee_earner' => false,
+      'time' => [],
+      'uplift' => [],
+      'net_cost' => []
+    }
+  %>
+  <%=
```
jsugarman, line 56: do we need a check for no work items here
```erb
  <% if work_items.size.zero? %>
    <p><%= t('.no_data') %></p>
  <% else %>
```

app/views/nsm/steps/view_claim/work_items.html.erb
```diff
@@ -8,67 +8,76 @@
...
-          { text: format_period(work_item.time_spent), numeric: true },
-          { text: NumberTo.percentage(work_item.uplift.to_f, multiplier: 1), numeric: true },
-          { text: NumberTo.pounds(work_item.total_cost), numeric: true },
-          { text: link, numeric: true }
-        ]
-      end
+      [
+        { header: true, text: index + 1, numeric: false},
+        { header: true, text: item_with_link, numeric: false},
+        { text: work_item.completed_on&.strftime("%-d %b %Y"), numeric: false},
```
jsugarman, line 63:
```suggestion
        { text: work_item.completed_on&.to_fs(:stamp), numeric: false},
```

imtiazAhmd: `date.to_fs(:stamp)` is giving me '22 July 2024' whereas We wanted to show as '22 Jul 2024'. I went through rails api and found out `date.to_fs(:rfc822)` this works.
[Api Doc Link](https://api.rubyonrails.org/classes/Date.html#method-i-to_formatted_s)

app/views/nsm/steps/view_claim/work_items.html.erb
imtiazAhmd: Work items are mandatory, you can't proceed the application without adding at least one work item.

app/views/nsm/steps/view_claim/work_items.html.erb
patrick-laa: I'd query that with Mary/Chris - I'd be somewhat surprised if we were deliberately choosing inconsistent date formatting across our service

imtiazAhmd: Added custom formatting named `:short_stamp` in initializers.

app/views/nsm/steps/view_claim/allowed_disbursements.html.erb
```diff
@@ -5,45 +5,54 @@
     <%= t('.disbursements') %>
   </h3>
 
-  <% pagy, disbursements = pagy(@disbursements.where.not(adjustment_comment: nil)) %>
+  <%
+    pagy, disbursements = pagy_array(@disbursements.filter{_1.adjustment_comment.present?}, items: 100)
```
patrick-laa, line 9: It might be more performant to do this in the DB, i.e.:

```
pagy, disbursements = pagy(@disbursements.where.not(adjustment_comment:[nil, '']), items: 100)
```

app/views/nsm/steps/view_claim/allowed_work_items.html.erb
```diff
@@ -8,88 +8,101 @@
...
+          { text: format_period(allowed_time, style: :minimal_html), numeric: true },
           { text: NumberTo.pounds(allowed_net_cost), numeric: true },
         ]
       end
 
       foot = [
         { text: t(".total"), width: 'govuk-!-width-one-quarter' },
         { text: '', numeric: true },
-        { text: NumberTo.pounds(total_cost), numeric: true },
+        { text: govuk_visually_hidden(t('.accessibility.net_cost')) + ' ' + NumberTo.pounds(total_cost), numeric: true },
```
patrick-laa, line 43: String concatenation isn't hugely idiomatic in ruby, and this will lead to unnecessarily unsafe strings. Consider:

```
safe_join([govuk_visually_hidden(t('.accessibility.net_cost')) , NumberTo.pounds(total_cost)])
```

app/views/nsm/steps/view_claim/allowed_work_items.html.erb
```diff
@@ -8,88 +8,101 @@
...
-            { text: format_period(work_item.time_spent), numeric: true },
-            { text: NumberTo.percentage(work_item.uplift.to_f, multiplier: 1), numeric: true },
-            { text: NumberTo.pounds(work_item.total_cost), numeric: true },
-            { text: format_period(work_item.allowed_time_spent || work_item.time_spent), numeric: true },
-            { text: NumberTo.percentage((work_item.allowed_uplift || work_item.uplift).to_f, multiplier: 1), numeric: true },
-            { text: NumberTo.pounds(work_item.allowed_total_cost), numeric: true },
-            { text: link, numeric: true }
-          ]
-        end
+  <% if work_items.size.zero? %>
```
patrick-laa, line 68: `work_items.none?` works here I think.

patrick-laa (approved):
Some suggestions but no blockers

.rubocop.yml
```diff
@@ -102,6 +102,8 @@ Metrics/ClassLength:
   Exclude:
     - app/services/decisions/**/*
     - app/services/submit_to_app_store/prior_authority/event_builder.rb
+    - app/presenters/nsm/check_answers/cost_summary_card.rb
```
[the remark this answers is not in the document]
jsugarman: Another pattern of putting it off I have worked with is to put into the `rubocop_todo.yml` list?! I was going to be cheeky and just remove a couple of line breaks in a couple of places where i would have preferred to have them to get around this also but... 🤷

app/models/disbursement.rb
```diff
@@ -31,6 +31,10 @@ def translated_disbursement_type
     end
   end
 
+  def position
```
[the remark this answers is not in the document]
jsugarman: This is because we have no persisted position for work and disbursement items as yet (separate ticket). Will add a comment for now

app/presenters/nsm/check_answers/cost_summary_card.rb
```diff
@@ -138,17 +140,28 @@ def vat_rate
...
+
+        text = I18n.t("nsm.steps.check_answers.groups.cost_summary.#{scope}.accessibility.#{key}")
+        sanitize("#{govuk_visually_hidden(text)} #{value}")
       end
 
       def t(key, numeric: true, width: nil)
         scope = show_adjustments ? 'with_adjustments' : 'base'
         {
-          text: I18n.t("nsm.steps.check_answers.groups.cost_summary.#{scope}.#{key}"),
+          text: sanitize(I18n.t("nsm.steps.check_answers.groups.cost_summary.#{scope}.#{key}")),
```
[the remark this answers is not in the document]
jsugarman: This comes out of requirement to either use an empty <td> for the 1:1 cell (used to be "Item") or put visually hidden content in it. govuk_table does not allow <td> inside a thead [even if you specify `header: false` for the cell]. An empty <th> fails accessibility tests (although oddly other places where there is an empty <th> are not being caught by the accessibility checks it seems). The i18n value is html to put a govuk-visually-hidden span in, so it needs sanitization, despite it being called `items_html` 🤷

app/views/nsm/steps/view_claim/allowed_disbursements.html.erb
jsugarman: would have been nice, but `@disbursements` is a sorted array now rather than an activerecord relation

app/presenters/nsm/check_answers/cost_summary_card.rb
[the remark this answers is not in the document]
patrick-laa: So the issue is that apparently I18n.t doesn't automatically mark "_html" key outputs as HTML-safe, even though the `t` helper in views and controllers does. 

BUT that's all it's doing, so rather than `sanitize` which parses the string and tries to scrub out illegal chars, you can simply call `.html_safe` to mark it as safe, because you know it is safe.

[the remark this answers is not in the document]
jsugarman: yep, `html_safe`  works. changed that

app/helpers/sortable_table_helper.rb
```diff
@@ -1,26 +1,37 @@
...
   end
 
   def reorder_form(path, column, next_direction, i18n_stem, index)
-    tag.form(action: path, method: 'get') do
-      safe_join([tag.input(type: 'hidden', name: 'sort_by', value: column),
-                 tag.input(type: 'hidden', name: 'sort_direction', value: next_direction),
-                 tag.button(type: 'submit', 'data-index': index) do
-                   I18n.t(column, scope: i18n_stem)
-                 end])
+    tag.form(action: path, params: params, method: 'get') do
```
dwhenry, line 12: do we need/want the params in the URL here given this is a post request and we have a heap of form fields as well?

hard to know when this is used/what can appear here - at the very least I feel like we should explicitly pass in any params we want to use

app/presenters/nsm/check_answers/cost_summary_card.rb
```diff
@@ -138,17 +140,28 @@ def vat_rate
...
+
+        text = I18n.t("nsm.steps.check_answers.groups.cost_summary.#{scope}.accessibility.#{key}")
+        sanitize("#{govuk_visually_hidden(text)} #{value}")
       end
 
       def t(key, numeric: true, width: nil)
         scope = show_adjustments ? 'with_adjustments' : 'base'
         {
-          text: I18n.t("nsm.steps.check_answers.groups.cost_summary.#{scope}.#{key}"),
+          text: I18n.t("nsm.steps.check_answers.groups.cost_summary.#{scope}.#{key}").html_safe,
```
dwhenry, line 164: we shouldn't need to add html_safe here... if it needs to contain html you should postfix the key with `_html`

config/initializers/time_formats.rb
```diff
@@ -5,3 +5,6 @@
 #
 Date::DATE_FORMATS[:stamp] = '%-d %B %Y' # DD MONTH YYYY
 Time::DATE_FORMATS[:stamp] = '%-d %B %Y' # DD MONTH YYYY
+Date::DATE_FORMATS[:short_stamp] = '%-d %b %Y'
+Time::DATE_FORMATS[:short_stamp] = '%-d %b %Y'
```
dwhenry, line 9: Is this used?

spec/rails_helper.rb
```diff
@@ -51,6 +51,14 @@
   # Use the faster rack test by default for system specs if possible
   config.before(:each, type: :system) { driven_by :rack_test }
 
+  # silence warning about "allow[ing] expectations on `nil`" without disabling entirely
+  config.around(:each, :allow_message_expectations_on_nil) do |example|
```
dwhenry, line 55: I understand what is does... but would be nice to know why we need this?

i.e. what exactly broke that we decided this was the best way to fix?

dwhenry (approved):
Overall looked pretty good to me...

app/presenters/nsm/check_answers/cost_summary_card.rb
jsugarman: is is postfixed with _html but renders escaped html without this - i think because this is not the view but a precursor to the rendered view

config/initializers/time_formats.rb
jsugarman: yep, https://github.com/ministryofjustice/laa-submit-crime-forms/pull/1011/files#diff-9a2a39ff9660e35e1abafcde500269713f3c08003e54f119144018b827b50d0dR32

spec/rails_helper.rb
jsugarman: removed as simpler method was implemented

app/helpers/sortable_table_helper.rb
jsugarman: This is specifically to handle the reused 'allowed_` logic of the view claims controller via the prefix param, at line 14. will have a look at passing the specific param but it feels more flexible/extensible to pass all and use the ones we want to me

jsugarman: ahh, have removed the params from the form tag as it is not needed there. instead we can just access params['prefix'] given the view context for the helper

PR #1012, opened 2024-07-23
[CRM457-789] letters calls validation if uplift apply true

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-789)

## Notes for reviewer
adds validation to catch 0 or blank uplifts on letter and calls form in CRM7 

Screenshot 2024-07-23 at 09 15 19

Screenshot 2024-07-23 at 09 15 40

### Before changes:
Screenshot 2024-07-23 at 09 18 30

Provider could enter 0 or leave blank and apply an uplift percentage to that

### After changes:
Validation catches uplift applied to 0 or blank and feebacks warning message to provider on save

- create a CRM7 and on letters and calls form enter a blank or 0 for either/both letter or calls and apply and uplift
- save for will respond with expected validation errors

### DIFF
--- a/app/forms/nsm/steps/letters_calls_form.rb
+++ b/app/forms/nsm/steps/letters_calls_form.rb
@@ -10,7 +10,11 @@ module Nsm
       attribute :calls_uplift, :integer
 
       validates :letters, numericality: { only_integer: true, greater_than_or_equal_to: 0, allow_blank: true }
+      validates :letters, numericality: { only_integer: true, greater_than_or_equal_to: 1, allow_blank: false },
+        if: :apply_letters_uplift
       validates :calls, numericality: { only_integer: true, greater_than_or_equal_to: 0, allow_blank: true }
+      validates :calls, numericality: { only_integer: true, greater_than_or_equal_to: 1, allow_blank: false },
+        if: :apply_calls_uplift
       validates :letters_uplift, presence: true,
         numericality: { only_integer: true, greater_than_or_equal_to: 1, less_than_or_equal_to: 100 },
         if: :apply_letters_uplift
--- a/app/forms/nsm/steps/letters_calls_form.rb
+++ b/app/forms/nsm/steps/letters_calls_form.rb
@@ -10,14 +10,12 @@ module Nsm
       attribute :calls_uplift, :integer
 
       validates :letters, numericality: { only_integer: true, greater_than_or_equal_to: 0, allow_blank: true }
-      validates :letters, numericality: { only_integer: true, greater_than_or_equal_to: 1, allow_blank: false },
-        if: :apply_letters_uplift
       validates :calls, numericality: { only_integer: true, greater_than_or_equal_to: 0, allow_blank: true }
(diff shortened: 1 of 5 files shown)


spec/forms/nsm/steps/letters_calls_form_spec.rb
```diff
@@ -71,14 +71,14 @@
         let(:calls) { '' }
 
         it 'have an error' do
-          expect(subject).to be_valid
+          expect(subject).not_to be_valid
```
patrick-laa, line 74: These tests no longer test what they're supposed to be testing (although I will grant that they're so badly written it's hard to tell what they're supposed to be testing). Rather than just change the expectations here, I think we should change the setup so that the expectations remain valid _and_ add extra tests that demonstrate that in a different context we get a different result.

app/forms/nsm/steps/letters_calls_form.rb
```diff
@@ -10,7 +10,11 @@ class LettersCallsForm < ::Steps::BaseFormObject
       attribute :calls_uplift, :integer
 
       validates :letters, numericality: { only_integer: true, greater_than_or_equal_to: 0, allow_blank: true }
+      validates :letters, numericality: { only_integer: true, greater_than_or_equal_to: 1, allow_blank: false },
```
patrick-laa, line 13: We're going to need some input from Chris on the wording of the error messages - we'll need full sentences explaining that you can't apply an uplift if no value is entered rather than 'must be greater than or equal to 1' which is likely to be confusing.

Also, to be fair, arguably the highlighted error field should be the 'Apply an uplift' checkbox rather than the number input? It might be worth getting Mary's eyes on this too as to what the intended behaviour is. Looking at the 'Expected behaviour' field on Jira, I think we might be a little bit early for code on this ticket.

spec/forms/nsm/steps/letters_calls_form_spec.rb
robinsk1: The way that test description was wrote it expects to have an error but was actually validating true on it _not_ having an error 😄 so i guess this fix was just aligning that. Ok, no worries i'll have a look at extra tests

app/forms/nsm/steps/letters_calls_form.rb
robinsk1: yeah, true! probably needs refinement on those two counts

patrick-laa: I've asked a question on Jira about this - we'll see what Mary and Chris say

robinsk1:
I've updated this with additional tests and messaging in jira ticket from Chris - i've created a provider application at letters_calls step for QA at https://crm457-789-uplift-nscc-provider-dev.cloud-platform.service.justice.gov.uk/non-standard-magistrates/claims eg 
Screenshot 2024-08-02 at 15 09 26
Screenshot 2024-08-02 at 15 09 54

patrick-laa (approved):

PR #1025, opened 2024-07-30
Crm457 1547 security context cleanup

ivanELEC:
## Description of change

[Check securityContext for all services](https://dsdmoj.atlassian.net/browse/CRM457-1547)

## Notes for reviewer

- Put Security Context default settings under a re-usable helper template
- Add information about the settings

### DIFF
--- a/helm_deploy/templates/_helpers.tpl
+++ b/helm_deploy/templates/_helpers.tpl
@@ -51,6 +51,18 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
 
 {{/*
+Default Security Context
+*/}}
+{{- define "helm_deploy.defaultSecurityContext" -}}
+runAsNonRoot: true
+allowPrivilegeEscalation: false
+seccompProfile:
+  type: RuntimeDefault
+capabilities:
+  drop: [ "ALL" ]
+{{- end }}
+
+{{/*
 Create the name of the service account to use
 */}}
 {{- define "laa-submit-crime-forms.serviceAccountName" -}}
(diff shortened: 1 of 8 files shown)


jsugarman (approved):

helm_deploy/templates/_helpers.tpl
```diff
@@ -50,6 +50,18 @@ app.kubernetes.io/name: {{ include "laa-submit-crime-forms.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
 
+{{/*
+Default Security Context
+*/}}
+{{- define "helm_deploy.defaultSecurityContext" -}}
```
jsugarman, line 56: for consistency it should be prefixed with the what other "methods" are prefixed with, namely 

```
{{- define "laa-submit-crime-forms.defaultSecurityContext" -}}
```

jsugarman: Although we are already using a combination, as at line 81 and 89.

jsugarman (approved):

helm_deploy/templates/_helpers.tpl
ivanELEC: yeah I've updated it throughout now so it's consistent and all are using laa-submit-crime-forms (I think that's clearer because helm_deploy gives off the impression we are using in built helm stuff and not custom stuff)

helm_deploy/templates/deployment-custom-errors.yaml
```diff
@@ -27,12 +27,7 @@ spec:
         - name: custom-error-pages
           mountPath: /www
         securityContext:
-          runAsNonRoot: true
-          allowPrivilegeEscalation: false
-          seccompProfile:
-            type: RuntimeDefault
-          capabilities:
-            drop: [ "ALL" ]
+          {{- include "laa-submit-crime-forms.defaultSecurityContext" . | nindent 8 }}
```
jsugarman, line 30: needs to be nindent 10

PR #1013, opened 2024-07-23
[CRM457-1547] Add linting for the output helm template

elken:
## Description of change

Mostly setup to catch issues with duplicate keys, there are a number of other rules that can be found below.

https://yamllint.readthedocs.io/en/stable/rules.html

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1547)

Started from a [conversation](https://mojdt.slack.com/archives/C06HFRK36KY/p1721746330169129) on Slack

PR #1033, opened 2024-07-31
Add additional helm chart and yaml linting/verifying

ivanELEC:
## Description of change

[Last outcome for this ticket](https://dsdmoj.atlassian.net/browse/CRM457-1547)

Note: this runs helm lint AND helm template (debug) because from my experience helm lint won't always necessarily catch issues that can break a chart (I tested helm lint on some misconfigurations and it didn't catch some of them) where helm template will AND helm template doesn't give us warning level suggestions like helm lint does - so both in tandem covers all bases for breaking changes and suggestions

Also including yamllint to check for things like duplicate keys upon suggestion from @elken[ in this PR](https://github.com/ministryofjustice/laa-submit-crime-forms/pull/1013)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -94,6 +94,14 @@ references:
     attach_workspace:
       at: ~/repo/tmp
 
+  _install-yamllint: &install-yamllint
+    run:
+      name: Install yamllint to use to check yaml formatting issues
+      command: |
+        sudo apt update
+        sudo apt install -y python3-pip
+        pip install yamllint
+
 # ------------------
 # COMMANDS
 # ------------------
@@ -127,6 +135,26 @@ commands:
       - run:
           name: Run rubocop
           command: bundle exec rubocop
+
+  # using a more recent version of yamllink so we can use the key_duplicates rule
+  # https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.key_duplicates
+
+  lint-and-verify-helm-config:
+      description:
+      parameters:
+        environment:
+          description: Name of environment to check helm template of
+          type: string
+      steps:
+        - helm/install_helm_client:
+            version: 3.7.2
+        - *install-yamllint
+        - run:
+            name: Lint helm chart for <<parameters.environment>>
+            command: helm lint helm_deploy --values helm_deploy/values-<<parameters.environment>>.yaml
+        - run:
+            name: Check helm template for <<parameters.environment>>
+            command: helm template --debug helm_deploy --values helm_deploy/values-<<parameters.environment>>.yaml | yamllint -
 
   db-setup:
     steps:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -128,6 +136,26 @@ commands:
...
+  lint-and-verify-helm-config:
+      description:
+      parameters:
+        environment:
+          description: Name of environment to check helm template of
+          type: string
+      steps:
+        - helm/install_helm_client:
+            version: 3.7.2
+        - *install-yamllint
```
jsugarman, line 151: These steps only need doing once. Here they would be installed for each environment run, slowing things down slightly.

jsugarman (commented):
Personally I am not really sure we need this as part of CI and the cost may outweigh the benefit in this regard. Its only ever something i have done locally while working on helm config and would expected deployments to fail if something was wrong. Is there a risk of false negatives? 

The second command using yamllint gives me lots of errors locally related to indentation, line length and trailing whitespaces but none on CI. Is it supposed to fail on finding linting problem?

elken:
> Personally I am not really sure we need this as part of CI and the cost may outweigh the benefit in this regard. Its only ever something i have done locally while working on helm config and would expected deployments to fail if something was wrong. Is there a risk of false negatives?
> 
> The second command using yamllint gives me lots of errors locally related to indentation, line length and trailing whitespaces but none on CI. Is it supposed to fail on finding linting problem?

The branch includes a config file to ignore the majority of useless errors, it fails for things like duplicate keys that are actually issues but otherwise reports nothing

ivanELEC:
> > Personally I am not really sure we need this as part of CI and the cost may outweigh the benefit in this regard. Its only ever something i have done locally while working on helm config and would expected deployments to fail if something was wrong.
> 
> You would expect it, but things like the `securityContext` being duplicated and just silently applied slipped through without issues. It'll prevent the potentially breaking deployment from going out, it's a safety net.
> 
> > Is there a risk of false negatives?
> 
> No tool is perfect, but it's preferential to the alternative.
> 
> > The second command using yamllint gives me lots of errors locally related to indentation, line length and trailing whitespaces but none on CI. Is it supposed to fail on finding linting problem?
> 
> The branch includes a config file to ignore the majority of useless errors, it fails for things like duplicate keys that are actually issues but otherwise reports nothing

if there are false negatives we can still disable rules like we do with Rubocop etc

jsugarman:
> It'll prevent the potentially breaking deployment from going out,

A truly broken deployment will fail and no pods be replaced in any event i believe, but it depends on the nature of the breakage. The last thing that broke a deployment was an `nindent` with too small a number. These steps do not catch that error when i tried it locally.

I guess my main concern is whether the undoubted benefit of extra checks actually unnecessarily blocks a deployment, not too mentioned requiring more knowledge and maintenance.

Anyway, do not let this stop you putting it in as we can trial it regardless. I think it is worth separating the install step from the command if possible though

elken:
> > It'll prevent the potentially breaking deployment from going out,
> 
> A truly broken deployment will fail and no pods be replaced in any event i believe, but it depends on the nature of the breakage. The last thing that broke a deployment was an `nindent` with too small a number. These steps do not catch that error when i tried it locally.

Indentation is one of the disabled rules, if it's one we want to catch then we can add it in. There are a few options available for it: https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.indentation

jsugarman (approved):

PR #1040, opened 2024-08-05
CRM457-1769: Remove tuboframes from provider

elken:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1769)

This PR removes all of the turbo-frame views in a similar vein to https://github.com/ministryofjustice/laa-assess-crime-forms/pull/638.

The View Claim screen in CRM7 has been refactored from using a tree of prefixes and sections to being static routes with admittedly long names. I left these mostly as they were to match the translations setup, but if preferred I can shorten them. The prefix setup has been removed in favour of a `before_action` method that uses the method name to determine which section to load and defaults to `:overview`.

Similarly for CRM4, but this was much simpler to remove.

Some of the javascript code in `application.js` has been left in because I'm not sure if it relates directly to turbo-frame specifically or some other component that's part of turbo.


app/views/nsm/steps/view_claim/calls.html.erb
```diff
@@ -1,5 +1,5 @@
 <% title t('.page_title') %>
-<% decision_step_header(path: nsm_steps_view_claim_path(current_application.id, section: params.fetch(:section, 'claimed_costs'), anchor: 'letters-and-calls-tab')) %>
+<% decision_step_header(path: :back) %>
```
elken, line 2: Investigate why this is a fixed path

app/views/nsm/steps/view_claim/show.html.erb
```diff
@@ -16,16 +16,14 @@
     <nav class="moj-sub-navigation" aria-label="Sub navigation" id="claim_nav">
       <ul class="moj-sub-navigation__list">
         <li class="moj-sub-navigation__item">
-          <%= link_to t('.tabs.overview'), nsm_steps_view_claim_path(@claim, section: 'overview', anchor: 'claim_nav'), class: 'moj-sub-navigation__link', 'aria-current': (@section == :overview ? "page" : nil) %>
+          <%= link_to t('.tabs.overview'), nsm_steps_view_claim_path(@claim, anchor: 'claim_nav'), class: 'moj-sub-navigation__link', 'aria-current': (@section == :overview ? "page" : nil) %>
         </li>
         <li class="moj-sub-navigation__item">
-          <%= link_to t('.tabs.claimed_costs'), nsm_steps_view_claim_path(@claim, section: 'claimed_costs', anchor: 'claim_nav'), class: 'moj-sub-navigation__link', 'aria-current': (@section == :claimed_costs ? "page" : nil) %>
+          <%= link_to t('.tabs.claimed_costs'), claimed_costs_work_items_nsm_steps_view_claim_path(@claim, anchor: 'claim_nav'), class: 'moj-sub-navigation__link', 'aria-current': (@section == :claimed_costs ? "page" : nil) %>
```
elken, line 22: Add a tech debt ticket to reduce path function names

app/views/nsm/steps/view_claim/calls.html.erb
elken: This was introduced in [this](https://github.com/ministryofjustice/laa-submit-crime-forms/commit/3d3e43dd09139e246930411a015c550269ba4ba3#diff-7909a93b507473f87db01fffc27ee7a7f51d49668c8245a6b317ac8fcaccb602) commit without an obvious indicator as to why other than it seeming related to turbo frames.

Will ask David when he's back

app/views/nsm/steps/view_claim/show.html.erb
elken: Done via https://dsdmoj.atlassian.net/browse/CRM457-1843

app/controllers/nsm/steps/view_claim_controller.rb
```diff
@@ -18,53 +18,89 @@ def item
...
       end
 
-      def work_items
+      def claimed_work_items
+        @report = CheckAnswers::ReadOnlyReport.new(current_application)
+        @claim = current_application
         @work_items = Sorters::WorkItemsSorter.call(current_application.work_items, @sort_by, @sort_direction)
 
-        render prefixed_view_for('work_items')
+        render 'nsm/steps/view_claim/show'
```
patrick-laa, line 26: I _think_ this can just be `render :show`. The "nsm/steps/view_claim" stuff is implied by the namespace/name of the controller, and once you've got rid of the slashes the string is simple enough it can be replaced by a symbol.

app/controllers/nsm/steps/view_claim_controller.rb
```diff
@@ -18,53 +18,89 @@ def item
         render report_params[:item_type]
       end
 
-      def work_items
+      def claimed_work_items
+        @report = CheckAnswers::ReadOnlyReport.new(current_application)
```
patrick-laa, line 22: Setting @report and @claim in every action doesn't feel very DRY. We might be able to do away with `@claim` entirely, cos I think you might be able to reference `current_application` directly from the view. But failing that, maybe we could set these in a `before_action`?

app/controllers/prior_authority/applications_controller.rb
```diff
@@ -1,12 +1,11 @@
 module PriorAuthority
   class ApplicationsController < ApplicationController
     before_action :set_default_table_sort_options
-    before_action :load_drafts, only: %i[index draft]
-    before_action :load_reviewed, only: %i[index reviewed]
-    before_action :load_submitted, only: %i[index submitted]
     layout 'prior_authority'
 
     def index
+      @pagy, @model = order_and_paginate(PriorAuthorityApplication.for(current_provider).reviewed)
```
patrick-laa, line 7: This may be overthinking, but I'm wondering if we can DRY this up as follows. If we adjust order_and_paginate to look like this:

```ruby
def  order_and_paginate
  query = yield PriorAuthorityApplication.for(current_provider)
  # ... same as before from here on
end

Then in our actions we can do: 

```ruby
@pagy, @model = order_and_paginate(&:reviewed)
```

So in the action we define a block that tells `order_and_paginate` what method to use to filter applications on.

config/locales/en/prior_authority.yml
```diff
@@ -18,6 +18,12 @@ en:
       expired: Expired
       provider_updated: Resubmitted
     applications:
+      reviewed:
+        page_title: Reviewed
+      submitted:
+        page_title: Submitted
+      drafts:
+        page_title: Submitted
```
patrick-laa, line 26: This doesn't look right

spec/system/nsm/view_claim_spec.rb
```diff
@@ -97,11 +97,19 @@
   end
 
   it 'show the work items and summary' do
-    visit work_items_nsm_steps_view_claim_path(claim.id)
+    visit claimed_costs_work_items_nsm_steps_view_claim_path(claim.id)
 
     # items
     expect(all('table caption, table td, table th').map(&:text)).to eq(
       [
+        'Cost summary', 'Item', 'Net cost', 'VAT', 'Total',
```
patrick-laa, line 105: I'm not sure that this is adding anything to the test, which is about claimed work items. Rather than include all of this new extra stuff in every expectation, given it seems kinda irrelevant to the test, perhaps we could change how we do the test from an `eq` to an `include`? Or change the selector string from `'table caption, table td, table th'` to something that just captures the bits we're interested in?

spec/system/prior_authority/application_lists_spec.rb
```diff
@@ -18,17 +18,19 @@
     create(:prior_authority_application, laa_reference: 'LAA-DDDDD', status: 'draft', updated_at: 4.days.ago)
     create(:prior_authority_application, laa_reference: 'LAA-EEEEE', status: 'draft',
            office_code: 'OTHER', provider: create(:provider, :other))
-
-    visit prior_authority_applications_path
   end
 
-  it 'only shows right applications in the right tabs' do
-    within '#submitted' do
+  context 'only shows right applications in the right tabs' do
```
patrick-laa, line 23: Nit: Contexts should be about when a certain condition applies. Really I think you want to use a `describe` block, maybe something like `describe 'tab contents'`. Then your `it` lines should really read like a grammatical sentence, like `it 'has correct content in submitted tab'`.

spec/system/prior_authority/view_applications_spec.rb
```diff
@@ -7,8 +7,6 @@
     travel_to arbitrary_fixed_date
     visit provider_saml_omniauth_callback_path
     application
-    visit prior_authority_applications_path
```
patrick-laa, line 10: How come we're un-DRYing this? If there's one test that needs _not_ to be on this URL, could we use a context/describe block for all the others that contains this stuff in a `before` so we don't need to copy-paste so much?

patrick-laa (commented):
There's some code in `application.js`, right at the bottom, that I think can be removed, as it's to do with turboframes.

Also, I think maybe we could now remove "@hotwired/turbo-rails" as a dependency entirely, which would be awesome as Turbo sucks and we have a million "just in case" instances of `data: { turbo: false }` that I'd love to be able to ditch.

app/controllers/nsm/steps/view_claim_controller.rb
elken: I'm sure I did and it picked up another but I'll try again

app/controllers/nsm/steps/view_claim_controller.rb
elken: Was just matching the existing convention but yeah I agree

config/locales/en/prior_authority.yml
elken: Oh you're right, good spot

spec/system/nsm/view_claim_spec.rb
elken: `eq` to `include` makes sense, will do

elken:
> There's some code in `application.js`, right at the bottom, that I think can be removed, as it's to do with turboframes.
> 
> Also, I think maybe we could now remove "@hotwired/turbo-rails" as a dependency entirely, which would be awesome as Turbo sucks and we have a million "just in case" instances of `data: { turbo: false }` that I'd love to be able to ditch.

Yeah I left a note about it in the body (next time I'll put it in a big "Patrick, read this" block 😉) but if it's just related to turboframes then I'll remove it too

elken:
@patrick-laa looks like the multi-form gem relies on turbo too. I think it'd be out of scope to remove it for this ticket (correct me otherwise) so I'll partially revert the last commit to remove the dep

jsugarman (commented):
one problem found with jumping when hitting the sortable headers for claimed and adjusted costs on the view_claim page

https://www.loom.com/share/81c5d6fccfb04fe394cd99170cda0b26?sid=94333349-f778-42b4-9d75-c6efd086eb8b

elken:
> one problem found with jumping when hitting the sortable headers for claimed and adjusted costs on the view_claim page
> 
> https://www.loom.com/share/81c5d6fccfb04fe394cd99170cda0b26?sid=94333349-f778-42b4-9d75-c6efd086eb8b

You mean it doesn't jump to the anchor? What's the issue here exactly? 😄

elken:
> There's some code in `application.js`, right at the bottom, that I think can be removed, as it's to do with turboframes.

RE: this, it seems to break a lot of tests and seems more related to multi-form gem than anything else so I've left it for now. Also prompted me to add in a comment about that.

app/javascript/application.js
```diff
@@ -7,6 +7,8 @@ import $ from 'jquery'
 
 initAll()
 
+// Most of this Turbo setup code is needed for the multi-form gem
```
patrick-laa, line 10: Sorry, I wasn't clear (I blame covid), and Turbo isn't very clear. So: Turbo comprises a couple of different things, including Turbolinks and Turboframes. The config here is basically trying to turn Turbolinks off because it's a plague and an abomination, but you're absolutely right, it does seem like removing the whole library is a bit more of a task than I naively thought (because, e.g. the cookie banner seems to rely on Turboframes still).

BUT the bit in application.js that I was referring to is the bit at the very bottom, line 55 onwards, which is specifically about the turboframes code that you've just removed from the views. So _that_ stuff can safely go, because otherwise it's orphaned JS that no longer does anything.

patrick-laa (approved):

app/javascript/application.js
elken: I tried removing just that and the tests still failed though

patrick-laa: That seems quite suspect - the code only does anything if there's HTML with CSS selector `.reload-visible-turboframes`, and you just removed the only instance of that class.

elken: Either that or I removed more than I thought. It passed locally so let's see what CI says

app/views/nsm/steps/view_claim/calls.html.erb
elken: David confirmed this was fine to change

PR #1049, opened 2024-08-07
CRM457-1831: Prog and prom questions

patrick-laa:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1831)


elken (commented):
I don't think we need to store all 3 of these booleans just the resulting PROG or PROM. Or are they there to track the step progress?

elken (approved):

PR #1052, opened 2024-08-08
CRM457-1025: Allow deleting draft CRM7 claims

elken:
## Description of change

Allows deletion of CRM7 claims.

Could potentially improve this by using the same view for both and passing in the URLs as a local.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1025)

### DIFF
--- a/app/controllers/nsm/claims_controller.rb
+++ b/app/controllers/nsm/claims_controller.rb
@@ -32,6 +32,16 @@ module Nsm
       render 'index'
     end
 
+    def confirm_delete
+      @model = Claim.for(current_provider).find(params[:id])
+    end
+
+    def destroy
+      @model = Claim.for(current_provider).find(params[:id])
+      @model.destroy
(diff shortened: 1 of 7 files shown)


patrick-laa (approved):

app/controllers/nsm/claims_controller.rb
```diff
@@ -32,6 +32,16 @@ def draft
       render 'index'
     end
 
+    def confirm_delete
+      @model = Claim.for(current_provider).find(params[:id])
+    end
+
+    def destroy
+      @model = Claim.for(current_provider).find(params[:id])
```
patrick-laa, line 40: I think a little system spec walking through the user journey for deleting a claim would be ideal.

elken: Yeah I realised I missed tests, just writing some up now 😄

patrick-laa (approved):

PR #1062, opened 2024-08-13
CRM457-1858: Move perform_laters out of transactions

patrick-laa:
## Description of change
Calling `perform_later` inside a transaction (including a lock) risks a race condition where the job might run before the transaction has been committed so uses stale DB data. This could cause incorrect data to be sent to the app store.  We've previously just lived with this, which is a high-risk strategy, as it relies on the app store always spotting and rejecting incorrect data. This PR moves the perform_laters until after the transactions have been committed.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1858)

### DIFF
--- a/app/forms/nsm/steps/solicitor_declaration_form.rb
+++ b/app/forms/nsm/steps/solicitor_declaration_form.rb
@@ -11,9 +11,9 @@ module Nsm
           application.status = :submitted
           application.update_disbursement_positions!
           application.update!(attributes)
-          SubmitToAppStore.perform_later(submission: application)
-          true
         end
+        SubmitToAppStore.perform_later(submission: application)
+        true
       end
     end
   end
--- a/app/forms/prior_authority/steps/check_answers_form.rb
+++ b/app/forms/prior_authority/steps/check_answers_form.rb
@@ -19,7 +19,7 @@ module PriorAuthority
         end
         SubmitToAppStore.perform_later(submission: application) if updated
 
-        true
+        updated
       end
 
       def update_application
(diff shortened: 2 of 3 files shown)


app/forms/prior_authority/steps/check_answers_form.rb
```diff
@@ -13,15 +13,15 @@ class CheckAnswersForm < ::Steps::BaseFormObject
...
 
       def persist!
+        updated = false
         application.with_lock do
-          update_application
+          updated = update_application
         end
+        SubmitToAppStore.perform_later(submission: application) if updated
+
+        true
```
elken, line 22: Is this return value used? Wouldn't returning `updated` be more appropriate?

patrick-laa: You're absolutely correct. Now fixed.

app/forms/nsm/steps/solicitor_declaration_form.rb
```diff
@@ -11,9 +11,9 @@ def persist!
           application.status = :submitted
           application.update_disbursement_positions!
           application.update!(attributes)
-          SubmitToAppStore.perform_later(submission: application)
-          true
         end
+        SubmitToAppStore.perform_later(submission: application)
+        true
```
elken, line 16: Do we not also want to capture the result of `update!` here then?

patrick-laa: Ah, not in this case, because `update!` will never return false - the exclamation means succeed or raise an error.

elken (approved):

PR #1067, opened 2024-08-14
CRM457-1828: Validation message when decimal entered into items field

patrick-laa:
## Description of change
Add specific messaging for not-whole-numbers
Remove unused options (item_type_dependant is always pluralized, cost_item_type_dependant is never pluralized)
Remove redundant values in specs

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1828)

### DIFF
--- a/app/validators/item_type_dependant_validator.rb
+++ b/app/validators/item_type_dependant_validator.rb
@@ -1,10 +1,12 @@
 class ItemTypeDependantValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    item_type = item_type_for(record)
-    item_type = item_type.pluralize if options[:pluralize]
+    item_type = item_type_for(record).pluralize
 
     record.errors.add(attribute, :blank, item_type:) if value.blank?
-    record.errors.add(attribute, :not_a_number, item_type:) if value.is_a?(String)
+    if value.is_a?(String)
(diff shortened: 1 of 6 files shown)


app/validators/item_type_dependant_validator.rb
```diff
@@ -1,10 +1,12 @@
 class ItemTypeDependantValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    item_type = item_type_for(record)
-    item_type = item_type.pluralize if options[:pluralize]
+    item_type = item_type_for(record).pluralize
 
     record.errors.add(attribute, :blank, item_type:) if value.blank?
-    record.errors.add(attribute, :not_a_number, item_type:) if value.is_a?(String)
+    if value.is_a?(String)
+      error = value.to_f.to_s == value ? :not_a_whole_number : :not_a_number
```
dwhenry, line 7: so I think I understand how the works, but it is very confussing (IMO) might be worth adding a comment above line 6... something like:

```
The value will only be a string when the automatica type conversion has failed, this means the value is NOT an integer and is either:
* has 1 or more decimal places
* is not a number at all which is why we can use the `to.f` check
```

or similar with smaller amount of words

elken (approved):

ivanELEC (approved):

dwhenry (approved):

dwhenry (approved):

PR #1076, opened 2024-08-16
CRM457-1808: Constrain long lines without whitespace in tables

elken:
## Description of change

`max-width` value was chosen arbitrarily, so can change if it's too small/large.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1808)

### Before changes:
image

### After changes:
image

### DIFF
--- a/app/assets/stylesheets/custom/govuk-table.scss
+++ b/app/assets/stylesheets/custom/govuk-table.scss
@@ -20,3 +20,8 @@
 .govuk-table .govuk-table__foot .govuk-table__row .govuk-table__footer:last-child {
   padding-right: 0;
 }
+
+.govuk-table .govuk-table__cell {
+  max-width: 4rem;
+  word-wrap: break-word;
+}


app/assets/stylesheets/custom/govuk-table.scss
```diff
@@ -19,4 +19,10 @@
...
   padding-right: 0;
-}
\ No newline at end of file
+}
+
+.govuk-table .govuk-table__cell {
+  max-width: 300px;
+  word-wrap: break-word;
+  white-space: break-spaces;
+}
```
dwhenry, line 28: So this broke other pages...

Screenshot 2024-08-16 at 11 34 17

dwhenry (changes requested):
It broke other tables..

```white-space: break-spaces;```
does something bad in some instances

app/assets/stylesheets/custom/govuk-table.scss
elken: Was afraid of that, the one table I didn't check 🤣

elken:
> It broke other tables..
> 
> `white-space: break-spaces;` does something bad in some instances

Yes it does, removed now

ivanELEC (approved):
The width works on desktop but not well for tablet or mobile. We don't have evidence people use mobile but we do have evidence of people using [tablets](https://mojdt.slack.com/archives/C034YPREUSF/p1722442334551609?thread_ts=1722440965.442539&cid=C034YPREUSF)

That said, I can't think of a quick way to rectify this, given the overwhelming majority are most likely desktop, approving this but I think we should make a separate ticket for making this work in (at least) tablet as well

elken:
> The width works on desktop but not well for tablet or mobile. We don't have evidence people use mobile but we do have evidence of people using [tablets](https://mojdt.slack.com/archives/C034YPREUSF/p1722442334551609?thread_ts=1722440965.442539&cid=C034YPREUSF)
> 
> That said, I can't think of a quick way to rectify this, given the overwhelming majority are most likely desktop, approving this but I think we should make a separate ticket for making this work in (at least) tablet as well

I agree with the points, I think I can fix it here too but I agree that we do need a general ticket to investigate tablet support.

elken:
Oh I see what you mean more, it's more of a table component issue so yeah definitely something to fix.

dwhenry (approved):

PR #1082, opened 2024-08-19
CMR457-1812: Update work item positions on submission of NSM claim

jsugarman:
## Description of change
Update work item positions on submission of NSM claim

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1812)

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -97,13 +97,20 @@ class Claim < ApplicationRecord
       ).slice!('letters', 'letters_uplift', 'calls', 'calls_uplift', 'app_store_updated_at')
   end
 
-  def disbursement_position(disbursement)
-    sorted_disbursement_ids.index(disbursement.id) + 1
+  def work_item_position(work_item)
+    sorted_work_item_ids.index(work_item.id) + 1
   end
 
-  def work_item_position(work_item)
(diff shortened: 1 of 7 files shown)


elken (approved):

ivanELEC (approved):

robinsk1 (approved):

app/models/claim.rb
```diff
@@ -116,6 +123,22 @@ def update_disbursement_positions!
 
   private
 
+  def sorted_work_item_ids
+    @sorted_work_item_ids ||= work_items.sort_by do |workitem|
+      [
+        workitem.completed_on || 100.years.ago,
```
dwhenry, line 129: so this won't do quite what you think... as `100.years.ago` returns a time value and it will be different for each row, hense records without a date will be arbitrarily sorted instead of following the worktype, created_at pattern... 

thsi also means the order may change on page refresh..
use this instead:
```
workitem.completed_on || Time.new(2000, 1, 1),
```
might also be worth checking if workj_type needs a default 
```
workitem.work_type&.downcase || '',
```

dwhenry (commented):
:) happy to discuss

app/models/claim.rb
jsugarman: ooh, good spot, albeit I think i checked this locally and it puts the "incompletes" at the top, by default, which is what we kind of want.

jsugarman: done.

PR #1095, opened 2024-08-21
CRM457-1902: Update the email regex to handle missing top-level domains

elken:
## Description of change

The regex added is the same as URI::MailTo::EMAIL_REGEXP with the domain part made to 1-or-more rather than 0-or-more

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1902)


app/lib/regular_expressions.rb
```diff
@@ -0,0 +1,4 @@
+class RegularExpressions
+  EMAIL_REGEXP =
+    %r{\A[a-zA-Z0-9.!\#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+\z}
```
patrick-laa, line 3: So what I don't like about this is that this regular expression is now entirely our problem. We're taking it on trust there's no typo in there. We have no tests to prove that it works, and no way of confirming that it's the same in most regards as a proven, battle-tested third party regex. It's also complex enough to be hard to just read and grok. So doing this means we just have to kind of trust and hope.

I can think of a couple of possible approaches that might make this less scary:

* Construct our regex by building it from `URI::MailTo::EMAIL_REGEXP` _plus_ some extra rules, and just adding tests for those rules
* Adding a boatload of tests of this regex, to demonstrate that all sane email formats are allowed and all insane formats aren't.
* Leaving the existing validation in place, and adding a separate validation that also checks for the presence of a `.` in the second half of the address, and just adding tests for this second validation

elken: I'm adding tests for it now yeah, we can't easily just construct it from the existing REGXP...

I like #3 the best, how would you pass multiple parameters into `with:` then?

patrick-laa: So you can't pass multiple regexes to `with`. You could possibly add a _second_ `validates :email` line, but from memory that leads to some slightly weird UI quirks. So I think you'd be best off adding a `validate :secondary_email_regex` line (note lack of `s` on validate), then defining something like:

```
def secondary_email_regex
  errors.add(:email, :invalid) unless /.*@.*\..*_or_whatever_/ =~ contact_email
end
```

Note you'd also want to do this in the NSM solicitor form

elken: Adjusted to only validate the TLD now

app/forms/prior_authority/steps/case_contact/solicitor_form.rb
```diff
@@ -9,6 +9,7 @@ class SolicitorForm < ::Steps::BaseFormObject
         validates :contact_first_name, presence: true
         validates :contact_last_name, presence: true
         validates :contact_email, presence: true, format: { with: URI::MailTo::EMAIL_REGEXP }
+        validates :contact_email, presence: true, format: { with: RegularExpressions::EMAIL_DOMAIN_REGEXP }
```
patrick-laa, line 12: Cool, looks like my fear about duplicate `validates :contact_email` was unfounded. It would be good to add this to the NSM solicitor form too, otherwise we'll have the same problem when they put invalid emails in that too.

elken: Yeah I thought it would override but it just checks all of them which is what we want here.

elken: Added to the solicitor form too (hopefully the one you meant?)

patrick-laa (approved):

PR #1109, opened 2024-08-23
CRM457-1554: More FI details

patrick-laa:
## Description of change
Show more RFI info on application details page

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1554)

### DIFF
--- a/spec/system/prior_authority/view_applications_spec.rb
+++ b/spec/system/prior_authority/view_applications_spec.rb
@@ -1,4 +1,4 @@
-require 'rails_helper'
+require 'system_helper'
 
 RSpec.describe 'View applications' do
   let(:arbitrary_fixed_date) { DateTime.new(2024, 3, 22, 15, 23, 11) }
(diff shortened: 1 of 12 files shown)


spec/system/prior_authority/view_applications_spec.rb
```diff
@@ -1,4 +1,4 @@
-require 'rails_helper'
+require 'system_helper'
```
elken, line 1: Why is this needed?

patrick-laa: So there's a bunch of config in `spec/system/support` that gets loaded by `system_helper`. If you run the entire test suite, system_helper gets loaded by the first system test so that config is applied for all system tests. But if you run any individual system spec files that don't have system_helper required, the tests run without that config applied, meaning they behave wonkily. The most obvious thing is that, I think because `Capybara.apybara.default_normalize_ws = true` is not set, `page.text` adds a bunch of newlines instead of spaces where text is separated by HTML elements, so a lot of `expect(page).to have_content` assertions fail.

TL;DR: All system specs files should require 'system_helper' so that when they're run in isolation they behave consistently.

elken (approved):

PR #1118, opened 2024-08-27
Crm457 1910 notification banner

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1910)

## Notes for reviewer

- Config to handle a single notification at a time 
- Uses GOV UK Notification Banner
- Has a time range based on UTC DateTime 
- Can set whether or not the notification is enabled per environment


lib/notification_banner.rb
```diff
@@ -0,0 +1,22 @@
+class NotificationBanner
+  include Singleton
+  attr_reader :message
+
+  def load_config
+    config = YAML.load(
```
patrick-laa, line 6: I know this code isn't ready for review yet, but just to flag up that this is a lot of custom code to do things in a custom way, when you could save a lot of lines of code and keep things more consistent if you follow the pattern for, e.g. inactive offices. So you just need one line in application.rb that says:

```
config.x.banners= config_for(:banners, env: ENV.fetch('ENV', 'localhost'))
```

Then in here you just need:

```
def active_banner
  Rails.configuration.x.banners.find do |banner|
    ( DateTime.parse(banner['date_from']..DateTime.parse(banner['date_to']).include?(DateTime.now)
  end
```

And then you're done. You don't need HostEnv, or YAML.load, or Singleton patterns or anything, and your banners.yml file can have a standardised format to handle different environments.

elken (dismissed):
I don't know if you need an initializer for it, `FeatureFlags` just loads the config as a singleton and freezes it.

Could this be simplified similarly?

lib/notification_banner.rb
elken: Gotta love github and its fake single-page app liveness...

But yeah I agree, there's room to simplify here I think

ivanELEC: indeed this does simplify things a great deal - got stuck in a rabbit hole trying to understand what the best pattern was for Singleton classes and lost sight of other ways to do it

config/notification_banner.yml
```diff
@@ -0,0 +1,12 @@
...
+#   notification:
+#     message: Insery message here
+#     date_from: "2024-08-27T18:00:00"
+#     date_to: "2024-08-28T19:00:00"
+
+production:
+  notification:
+    message: This service will be down on Wednesday 27th August between 6 and 8pm for security updates
+    date_from: "2024-08-27T18:00:00"
+    date_to: "2024-08-28T19:00:00"
```
jsugarman, line 12: should be 2024-08-28T19:00:00 - we agreed 2 hours i think. in case we can do app store

lib/notification_banner.rb
```diff
@@ -0,0 +1,18 @@
+class NotificationBanner
+  def self.active_banner
+    current_date = DateTime.now
```
jsugarman, line 3:
```
return nil unless date_from && date_to
```

spec/lib/notification_banner_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    context 'when no config provided' do
+      let(:config) { nil }
+
+      it 'correctly returns no notification message' do
+        expect(described_class.active_banner).to be_nil
+      end
+    end
+
+    # context 'when date_from present but not date_to' do
```
jsugarman, line 49: need to remove but can clear up after

jsugarman (approved):

spec/lib/notification_banner_spec.rb
jsugarman: have another pr to cleanup

PR #1125, opened 2024-08-28
CRM457-1863: Update tabs to sub-navigation

elken:
## Description of change

Update tabs to sub-navigation & change the delete link to meet new a11y requirements.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1863)

### DIFF
--- a/app/views/prior_authority/applications/_drafts.html.erb
+++ b/app/views/prior_authority/applications/_drafts.html.erb
@@ -40,8 +40,13 @@
             <%= application.laa_reference %>
           </td>
           <td class="govuk-table__cell">
-            <%= link_to t("prior_authority.applications.tabs.delete"),
-                        confirm_delete_prior_authority_application_path(application) %>
+            <%= link_to confirm_delete_prior_authority_application_path(application), class: "govuk-link--no-visited-state" do %>
+              <%= t("prior_authority.applications.tabs.delete") %>
+              <span class="govuk-visually-hidden">
+                <%= t(".application") %>
+                <%= application.laa_reference %>
+              </span>
+            <% end %>
           </td>
         </tr>
       <% end %>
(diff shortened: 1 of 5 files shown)


app/views/prior_authority/applications/_drafts.html.erb
```diff
@@ -40,8 +40,16 @@
             <%= application.laa_reference %>
           </td>
           <td class="govuk-table__cell">
-            <%= link_to t("prior_authority.applications.tabs.delete"),
-                        confirm_delete_prior_authority_application_path(application) %>
+            <a
```
patrick-laa, line 43: It doesn't really make a difference, but FYI you could have continued to use `link_to` rather than manually constructing the <a> html by doing:

```erb
<%= link_to confirm_delete_prior_authority_application_path(application) , class: "govuk-link--no-visited-state" do %>
  <%= t("prior_authority.applications.tabs.delete") %>
  <span class="govuk-visually-hidden">
    applications
    <%= application.laa_reference %>
  </span>
<% end %>
```

```

File: app/views/prior_authority/applications/_drafts.html.erb [hunk 2]
```diff
@@ -40,8 +40,16 @@
...
+              href="<%=confirm_delete_prior_authority_application_path(application)%>"
+              class="govuk-link--no-visited-state"
+            >
+              <%= t("prior_authority.applications.tabs.delete") %>
+              <span class="govuk-visually-hidden">
+                applications
```
patrick-laa, line 49: I think this should be `application` rather than `applications` as there's only one of them, and strictly this should be I18nised as `<%= t('.application') %>`

patrick-laa (approved):

app/views/prior_authority/applications/_drafts.html.erb
elken: Ah I tried something like that but was getting errors, thanks

app/views/prior_authority/applications/_drafts.html.erb
elken: Oh applications is what was on the ticket so I just copied that text, but you're right

PR #1126, opened 2024-08-28
CRM457-1845: Uplift validation

patrick-laa:
## Description of change
Change how work item uplift is validated and update wording

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1845)


app/forms/nsm/steps/work_item_form.rb
```diff
@@ -8,17 +8,19 @@ class WorkItemForm < ::Steps::BaseFormObject
...
       validates :time_spent, presence: true, time_period: true
       validates :completed_on, presence: true,
               multiparam_date: { allow_past: true, allow_future: false }
       validates :fee_earner, presence: true
-      validates :uplift, presence: true,
-              numericality: { only_integer: true, greater_than_or_equal_to: 1, less_than_or_equal_to: 100 },
-              if: :apply_uplift
+      validates :uplift,
+                presence: true,
+                numericality: { allow_blank: true, only_integer: true, greater_than_or_equal_to: 1, less_than_or_equal_to: 100 },
```
elken, line 21: Doesn't `allow_blank` contradict `presence: true`?

elken (approved):

app/forms/nsm/steps/work_item_form.rb
patrick-laa: So allow_blank in the `numericality` section mean that the numericality errors won't trigger if the item is blank, meaning we can guarantee that if the item is blank it will be the `presence` validation that triggers, meaning that the error will be `blank` rather than `less_than_or_equal_to` etc. It doesn't really matter here, as the error message is currently the same for either type of error, but it can be fiddly when it's not clear which of several different validation rules would apply if the field was blank.

elken: Right okay, maybe add a comment for that then?

patrick-laa: Cool will do

PR #1127, opened 2024-08-28
CRM457-1910: Add integration test for notification banner

jsugarman:
## Description of change
Add integration test for notification banner

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1910)


elken (approved):
Worth adding a case for "when there is an upcoming banner that shouldn't be displayed"?

PR #1137, opened 2024-08-30
Add notification for Production DB Upgrade

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1907)


jsugarman (approved):

config/notification_banner.yml
```diff
@@ -7,18 +7,18 @@ localhost:
 
 development:
   notification:
-    message: This service will be unavailable on Wednesday 28 August 2024 between 6pm and 8pm.
-    date_from: "2024-08-27T18:00:00"
-    date_to: "2024-08-28T19:00:00"
+    message: This service will be unavailable on Wednesday 4 September 2024 between 6pm and 8pm.
+    date_from: "2024-08-30T16:00:00"
```
robinsk1, line 11: is that date right? should that not be 2024-09-04T18:00:00 - same for other envs

config/notification_banner.yml
```diff
@@ -7,18 +7,18 @@ localhost:
...
+    date_to: "2024-09-04T19:00:00"
```
robinsk1, line 12: should that be "2024-09-04T20:00:00" if its 8pm?

jsugarman: its in UTC, but agreed it could be clearer, other than the comment at the top pf the file

PR #1143, opened 2024-09-02
CRM457-1950: show granted adjustments

jsugarman:
## Description of change
Show adjustments and allowed total for granted applications with adjustments

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1950)

Claims can be adjusted upward and still granted. In such instances the
provider needs to be made aware of the adjustments nonetheless.

## Notes for reviewer
It has been agreed with PM and designer that if granted applications are adjusted downwards (which is currently possible) we would show the adjustments regardless for transparency. Going forward CRM457-1679 is intended to prevent caseworkers granting claims with downward adjustments.

### DIFF
--- a/app/presenters/nsm/check_answers/cost_summary_card.rb
+++ b/app/presenters/nsm/check_answers/cost_summary_card.rb
@@ -67,7 +67,7 @@ module Nsm
         sum_allowed(data_for_footer, :allowed_gross_cost)
       end
 
-      def total_gross_adjusted?
+      def show_adjusted?
         return false unless claim.granted? || claim.part_grant?
 
         total_gross_allowed != total_gross
--- a/app/presenters/nsm/check_answers/cost_summary_card.rb
+++ b/app/presenters/nsm/check_answers/cost_summary_card.rb
@@ -67,6 +67,12 @@ module Nsm
         sum_allowed(data_for_footer, :allowed_gross_cost)
       end
 
+      def total_gross_adjusted?
+        return false unless claim.granted? || claim.part_grant?
+
+        total_gross_allowed != total_gross
+      end
+
       private
 
       def data_for_footer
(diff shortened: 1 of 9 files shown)


app/presenters/nsm/check_answers/cost_summary_card.rb
```diff
@@ -67,6 +67,12 @@ def total_gross_allowed
         sum_allowed(data_for_footer, :allowed_gross_cost)
       end
 
+      def total_gross_adjusted?
+        return false unless claim.granted? || claim.part_grant?
```
robinsk1, line 71: would it be more SRP if we move that guard out and its that something we check separately before hitting `total_gross_adjusted?` ie  `return "#{NumberTo.pounds(total_gross_allowed)} allowed" if not_granted_or_part_grant? && total_gross_adjusted?`?

robinsk1 (approved):

app/presenters/nsm/check_answers/cost_summary_card.rb
jsugarman: This could be because it is not really named appropriately. It could/should be named `show_adjustments?`. It is being used to display (or not) the adjusted total and the adjusted tab with a breakdown of the adjusted values. 

Would that make sense?

robinsk1: `show_adjustments?` feels clearer 😄 - cheers!

jsugarman: unfortunately `show_adjustments` is a var heavily used in the cost summary card and elsewhere. need another name that makes sense really. 🤔

patrick-laa: `display_adjustments?`

PR #1167, opened 2024-09-11
[CRM457-1990] CRM7 Provider: Addition of print instructions on the Claim complete page

imtiazAhmd:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1990)

### Before changes:
Screenshot 2024-09-11 at 09 38 17

### After changes:
Screenshot 2024-09-11 at 09 39 15

### DIFF
--- a/config/locales/en/nsm/steps.yml
+++ b/config/locales/en/nsm/steps.yml
@@ -322,7 +322,9 @@ en:
             process_1: "We'll assess your claim."
             process_2_html: "<p class='govuk-body'>We’ll send you an email to let you know the outcome of your claim or to ask for more information.</p>
               <p class='govuk-body'>It normally takes up to 10 working days to assess a claim. After this you can contact us to check the status of your claim:</p>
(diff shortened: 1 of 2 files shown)


ivanELEC (approved):

config/locales/en/nsm/steps.yml
```diff
@@ -322,7 +322,9 @@ en:
             process_1: "We'll assess your claim."
             process_2_html: "<p class='govuk-body'>We’ll send you an email to let you know the outcome of your claim or to ask for more information.</p>
```
patrick-laa, line 323: It's making me uncomfortable just how much markup is in our translation files now. Strictly speaking, all the html tags should be in the ERB files, and just the text should be in here. Maybe it's doesn't need to be fixed right away, but if not I think we're approaching the point where we might need a ticket to rework this chunk as a piece of tech debt.

patrick-laa (approved):

config/locales/en/nsm/steps.yml
imtiazAhmd: May be we should talk about this, I have just followed the current practice, i.e. `config/locales/en/nsm/steps.yml:293`. Personally I completely agree with you.

PR #1181, opened 2024-09-17
Use ConfigMap for Maintenance Mode Flag

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1800)

- configuration.yaml adds a configmap for general settings we can change on the fly including MAINTENANCE_MODE
- MAINTENANCE_MODE configmap value use for environment variable 
- Maintenance mode feature uses environment variable OR feature flag if environment variable isn't present (note: want opinions on whether to remove feature flag completely or not)
- enable_maintenance_mode/disable_maintenance_mode scripts can be ran locally if you have kubernetes auth, 1st arg is the namespace you want to run this in, 2nd arg is an optional BRANCH_RELEASE_NAME value which defaults to laa-submit-crime-forms-app

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,3 +120,12 @@ We have a default [k8s security context ](https://kubernetes.io/docs/reference/g
 - allowPrivilegeEscalation - AllowPrivilegeEscalation controls whether a process can gain more privileges than its parent process. Currently defaults to false, this limits the level of access for bad actors/destructive processes
 - seccompProfile.type - The Secure Computing Mode (Linux kernel feature that limits syscalls that processes can run) options to use by this container. Currenly defaults to RuntimeDefault which is the [widely accepted default profile](https://docs.docker.com/engine/security/seccomp/#significant-syscalls-blocked-by-the-default-profile)
 - capabilities - The POSIX capabilities to add/drop when running containers. Currently defaults to drop["ALL"] which means all of these capabilities will be dropped - since this doesn't cause any issues, it's best to keep as is for security reasons until there's a need for change
+
+## Maintenance Mode
+This service has a maintenance mode which can be used to prevent users taking actions on the system.
+
+You can use the enable_maintenance_mode/disable_maintenance_mode scripts via `sh bin/enable_maintenance_mode <namespace>` or `sh bin/disable_maintenance_mode <namespace>`.
+
+If you want to change maintenance mode on a branch deployment, add the branch release name as a 2nd arg like this `sh bin/enable_maintenance_mode <namespace> <branch release name>`
+
+The helm_deploy/templates/configuration.yaml file initialises the configmap that is used to store the setting. It has a helm.sh/resource-policy of keep so that new deployments won't reset the setting. If the configmap is deleted for whatever reason, simply restart the deployment and the configmap will get regenerated with default values.
--- a/helm_deploy/templates/configuration.yaml
+++ b/helm_deploy/templates/configuration.yaml
@@ -1,0 +1,10 @@
+{{- if not (lookup "v1" "ConfigMap" .Release.Namespace (include "laa-submit-crime-forms.fullname" .)) -}}
+apiVersion: v1
+kind: ConfigMap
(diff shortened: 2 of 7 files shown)


helm_deploy/templates/configuration.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if not (lookup "v1" "ConfigMap" .Release.Namespace (include "laa-submit-crime-forms.fullname" .)) -}}
```
patrick-laa, line 1: Can you explain what this line means? Is this a "if this isn't a branch" check?

ivanELEC: having this template in our helm means that it will get recreated on every deployment - this includes re-establishing the default value of maintenance mode = false - we don't want that to happen as we want whatever we've already set on the configmap to take presedence - we only want the configmap to get recreated if it's been deleted for whatever reason

This checks if the configmap exists by doing a lookup of the version, kind, namespace and name and will only generate the configmap if it doesn't already exist

patrick-laa: Ah. So the readme says that what's preventing it from being reset on every deployment is that it "has a helm.sh/resource-policy of keep". Does the "if" clause here make the resource-policy redundant, and if so is it worth updating the readme?

ivanELEC: oh sorry the readme/this doesn't fully explain the situation: 
1. We don't want the config map to get reset to defaults --> only generate if doesn't exist because generating it again resets the values
2. We don't want the config map to get deleted --> because the if statement removes the template from the deployment, when it's false, the configmap doesn't get left alone, it gets deleted - therefore we have the resource policy to ensure it doesn't get deleted

I should probably be more clear about this in the readme as another caveat is that if we want to make changes for example to the metadata of the resource, we would also have to delete the existing resource to allow the new changes to come into effect

ivanELEC: I would have preferred a policy that doesn't reset the data to simplify things but I couldn't find anything
only other option I could think of is to keep the data of the configmap empty, default to false in the code and only use the configmap when data is present - however that would mean awkwardly adding a fallback to the helm deployment when bringing in the environment variable and settings being less explicit as to where they come from

README.md
```diff
@@ -120,3 +120,16 @@ We have a default [k8s security context ](https://kubernetes.io/docs/reference/g
...
 - capabilities - The POSIX capabilities to add/drop when running containers. Currently defaults to drop["ALL"] which means all of these capabilities will be dropped - since this doesn't cause any issues, it's best to keep as is for security reasons until there's a need for change
+
+## Maintenance Mode
+This service has a maintenance mode which can be used to prevent users taking actions on the system.
+
+You can use the enable_maintenance_mode/disable_maintenance_mode scripts via `sh bin/enable_maintenance_mode <namespace>` or `sh bin/disable_maintenance_mode <namespace>`.
+
+If you want to change maintenance mode on a branch deployment, add the branch release name as a 2nd arg like this `sh bin/enable_maintenance_mode <namespace> <branch release name>`
+
+The helm_deploy/templates/configuration.yaml file initialises the ConfigMap that is used to store the setting. It only re-generates if the ConfigMap exists (using a lookup to check) so that new deployments won't reset the setting and has helm.sh/resource-policy of keep so that the deployment doesn't delete the ConfigMap when we don't want to re-generate it.
```
jogilder, line 131: @ivanELEC I don't think this says what you have discussed with Patrick. Shouldn't 

`It only re-generates if the ConfigMap exists (using a lookup to check)`  

be 

`It only re-generates if the ConfigMap does not exist (using a lookup to check)`

jogilder (changes requested):
see comment

README.md
ivanELEC: done

jogilder (approved):

robinsk1 (approved):

PR #1189, opened 2024-09-19
log auth_hash info when checking provider is enrolled

ivanELEC:
## Description of change

This will be reverted after we do the check 
Only will apply to UAT so we don't put PII in logs


jogilder (approved):
LGTM

robinsk1 (approved):

PR #1195, opened 2024-09-23
CRM457-1755: SolidCache

patrick-laa:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1755)

## Notes for reviewer
* I've played with stuffing 1,000,000 records into the cache table to see if it noticeably impacts performance (it doesn't)
* I've also messed around with very low limits of records in the cache table to verify what happens when the limit is reached. As you'd expect, the oldest keys get evicted, meaning that the earliest logged-in user gets logged out. Since we log out after an hour of idle time this isn't a problem - it would be practically impossible for someone to remain active on the service for enough time to pass for a million other cache entries to be written after they first logged in, and even if that happened the only consequence would be that they were abruptly logged out.


robinsk1 (approved):
looks good - might need a think about the max_age, max_entries values though

PR #1196, opened 2024-09-24
CRM457-2006: NSM expiry

patrick-laa:
## Description of change
UI to display expired claim details.

As we've agreed to reuse the same further information payload as prior authority, and the FI UI cards look the same in both services, I've made the FurtherInformation model polymorphic so it can be reused across both submission types, and extracted the presenter logic for the FI card into a concern to make that reusable too. I've added a TODO marking how this information will need to be populated when a claim is sent back. 

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2006)


app/presenters/further_information_presentable.rb
```diff
@@ -0,0 +1,40 @@
+module FurtherInformationPresentable
```
robinsk1, line 1: maybe we could keep this in a `concerns` folder under presenters just it as a potential bit of re-usable code

db/migrate/20240924090158_add_resubmission_deadline_to_further_informations.rb
```diff
@@ -0,0 +1,5 @@
+class AddResubmissionDeadlineToFurtherInformations < ActiveRecord::Migration[7.2]
+  def change
```
robinsk1, line 2: is it worth popping that through first seperately, not sure if there are race conditions with pod creations and things

robinsk1 (approved):

db/migrate/20240924090158_add_resubmission_deadline_to_further_informations.rb
patrick-laa: We should be ok in that migrations happen before new code is deployed, so so long as we're not making a DB change that makes the DB incompatible with the old code we're safe.

PR #1203, opened 2024-09-25
SPIKE: CRM457-1920: PII enforcement

patrick-laa:
## Description of change
2 proof-of-concept specs to enforce keeping PII definitions up-to-date. 

One assumes a PII yaml file has been created which explicitly marks certain DB columns as containing PII. To ensure this yaml file is maintained when new columns are added, this spec will fail unless every DB column has been marked as either PII or not.

The other spec assumes the repo contains a JSON Schema file which will be used to validate every payload that is generated or stored. It parses the file and checks to make sure that for every property (including nested object properties), a custom PII boolean annotation has been set.

These will ensure that adding new stuff to the database or payload will break the test suite unless those changes also include updating the PII definitions to match.

### DIFF
--- a/config/pii.yml
+++ b/config/pii.yml
@@ -1,0 +1,13 @@
+# For each database table, and each column in that table, flag whether
+# there is PII in the column:
+firm_offices:
+  name: true
+  address_line_1: true
+  address_line_2: true
+  town: false
+  postcode: true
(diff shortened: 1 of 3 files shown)


patrick-laa:
Closing as superseded by full implementation

PR #1210, opened 2024-09-27
CRM457-2083: Add cost summary to claim payload

patrick-laa:
## Description of change
This will allow us to query against the blob in caseworker and app store to filter against `WHERE ({jsonb_column}->'cost_summary'->'profit_costs'->>'gross_cost')::numeric >= {threshold}"`

I tweaked the cost summary card to make sure that zeroes are always big decimals so they are rendered consistently in JSON (bigdecimals get turned into strings to retain precision while integers don't), and to filter out a bunch of useless nil hash values.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2083)


app/presenters/nsm/check_answers/cost_summary_card.rb
```diff
@@ -206,6 +207,14 @@ def name_row(name, work_type_changed)
...
       end
+
+      def allowed_total_cost(allowed_rows)
+        allowed_rows.sum(zero) { _1.allowed_total_cost || zero }
+      end
+
+      def zero
+        BigDecimal(0)
+      end
     end
```
ivanELEC, line 218: please could you add a quick comment re: why we are using the BigDecimal zero. I understand from your comments in the PRs channel but I feel like this might be one of those things that's hard to understand the intention of if you have no context

patrick-laa: Sure! Done.

ivanELEC (approved):

PR #1213, opened 2024-10-02
CRM457-2096: Validate data before submission

patrick-laa:
## Description of change
And update test data to ensure that it's valid.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2096)

### DIFF
--- a/app/services/submit_to_app_store/prior_authority_payload_builder.rb
+++ b/app/services/submit_to_app_store/prior_authority_payload_builder.rb
@@ -9,10 +9,18 @@ class SubmitToAppStore
       { application_id: application.id,
         json_schema_version: 1,
         application_state: application.state,
-        application: data,
+        application: validated_data,
         application_type: 'crm4',
         application_risk: 'N/A',
         events: @include_events ? PriorAuthority::EventBuilder.call(application, data) : [] }
+    end
+
+    def validated_data
+      built_data = data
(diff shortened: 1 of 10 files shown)


app/services/submit_to_app_store/prior_authority_payload_builder.rb
```diff
@@ -9,12 +9,20 @@ def payload
...
         application_state: application.state,
-        application: data,
+        application: validated_data,
         application_type: 'crm4',
         application_risk: 'N/A',
         events: @include_events ? PriorAuthority::EventBuilder.call(application, data) : [] }
     end
 
+    def validated_data
+      built_data = data
```
elken, line 19: Should this logic not match https://github.com/ministryofjustice/laa-assess-crime-forms/pull/791/files#diff-d3a41f06a5d16530196db1064f66524b3318790b22adce1fd254a8c9f060c653 ?

elken (approved):

app/services/submit_to_app_store/prior_authority_payload_builder.rb
patrick-laa: It won't exactly match it because this class already knows it's building a PA payload, and it actually needs to construct it from scratch so it needs to be memoised. On the other side the payload already exists, but on the flipside it could be PA or NSM.

PR #1216, opened 2024-10-04
CRM457-2129: Use NSM schema validation

patrick-laa:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2129)

### DIFF
--- a/app/services/submit_to_app_store/nsm_payload_builder.rb
+++ b/app/services/submit_to_app_store/nsm_payload_builder.rb
@@ -1,4 +1,5 @@
 class SubmitToAppStore
+  # rubocop:disable Metrics/ClassLength
   class NsmPayloadBuilder
     DEFAULT_IGNORE = %w[claim_id created_at updated_at].freeze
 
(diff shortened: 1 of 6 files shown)


app/services/submit_to_app_store/nsm_payload_builder.rb
```diff
@@ -14,14 +14,22 @@ def payload
...
+        application: validated_data,
         application_risk: scorer.calculate(claim),
         application_type: 'crm7'
       }
     end
 
     private
 
+    def validated_data
+      built_data = data
```
elken, line 26: Why is this being copied? `validate` shouldn't mutate it, right?

patrick-laa: It's because `data` isn't an attribute, it's a method that does work, so we don't want to call it twice. Instead I'm assigning its output to a variable.

elken: Ah in that case then maybe it's cleaner to use `tap`? (untested napkin Ruby)

```ruby
construct_payload.tap do |payload|
  issues = LaaCrimeFormsCommon::Validator.validate(:nsm, payload)
  raise "Validation issues detected for #{claim.id}: #{issues.to_sentence}" if issues.any?
end
 ```

patrick-laa: Very good shout! Yes, have updated to use exactly that.

elken (approved):

PR #1217, opened 2024-10-07
Crm457 2004: Fix CRM7 RFI Syncing Bug

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2004)

## Notes for reviewer
[Mentioned in Slack thread](https://mojdt.slack.com/archives/C034YPREUSF/p1728293648205059)


app/services/nsm/assessment_syncer.rb
```diff
@@ -82,13 +82,14 @@ def sync_disbursements
 
     def sync_further_info_requests
       data['further_information'].each do |further_info|
-        claim.further_informations.find_or_create_by(
+        record = claim.further_informations.find_or_create_by(
```
patrick-laa, line 85: This bit looks redundant to me. The output of `find_or_create_by` will always be persisted to the database. `save` won't do anything because Rails knows nothing has changed.

ivanELEC: was thinking save would be needed for when we are setting the resubmission_deadline value but thinking about it, I guess because it's being set via a block from the find_or_create_by method I don't need to?

patrick-laa: Exactly. First it queries the database for an object matching the arguments. If it can't find one it writes to the database combining the arguments with whatever's set in the block. By the time the method has returned a value, that value is definitely in the DB.

patrick-laa (approved):

PR #1219, opened 2024-10-07
CRM457-2004: Ensure application is saved when provider is updating

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2004)

## Notes for reviewer
[Slack Thread](https://mojdt.slack.com/archives/C034YPREUSF/p1728293648205059)


patrick-laa (approved):
Arguably you want `.save!` instead of `.save` so that if there's a problem things blow up and we get notified, but I don't think it's critical.

PR #1231, opened 2024-10-10
CRM457-2101: Move clam to separate deployment/pods

patrick-laa:
## Description of change
- Instead of having clamav in a container on every app pod, have it running as its own pod in its own deployment accessed via its own service
- This allows clamav to have its own liveness probe, and not have to be restarted every time new app pods are rolled out
- Switch to using the latest image
- Only have one clam service on dev, not one per branch, reusing the same mechanism employed for the custom error service
[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2101)

## Note to reviewer
When this is merged, dev main will be responsible for the dev clam service, and branches will simply use the service. However, for the purposes of this getting clamby working for this PR we need the branch to run the service. Hence in the second commit I've hacked the config to have the branch deploy the service, and added 3 'TODO' statements to revert it on merge to giving responsibility to main.

### DIFF
--- a/helm_deploy/templates/deployment-clamav.yaml
+++ b/helm_deploy/templates/deployment-clamav.yaml
@@ -1,0 +1,67 @@
+{{ if not .Values.branch }}
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: clamav
+spec:
+  replicas: {{ .Values.replicaCount.clamav }}
+  revisionHistoryLimit: 2
+  minReadySeconds: 15
+  serviceName: clamav
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 50%
+  selector:
+    matchLabels:
+      app: submit-crime-forms-clamav
+  template:
+    metadata:
+      labels:
+        app: submit-crime-forms-clamav
+        tier: backend
+    spec:
+      securityContext:
+        fsGroup: 1000
+        runAsUser: 1000
+        runAsGroup: 1000
+      containers:
+      - name: clamav
+        image: ghcr.io/ministryofjustice/hmpps-clamav-freshclammed:latest
+        imagePullPolicy: Always
+        command: ['/bin/sh', '-c', 'freshclam && clamd && touch startup.txt && clamdscan startup.txt']
+        securityContext:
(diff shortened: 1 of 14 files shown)


jogilder:
@patrick-laa do we only want one replica for all environments? I know that it was combined with the web service but there were 4 before in prod. 
Do we need a different security context for the clam-av deployment? I don't really understand how that works.

patrick-laa:
@jogilder good point re replicas. 4 seems overkill given how few requests involve file uploads, but having just one of anything in production is a little bit uneasy-making. I have adjusted to have 2 in UAT and prod.

And re the security context settings, I'm out of my depth but I don't see anything in the default settings that feels like it doesn't belong in the clam container.

helm_deploy/templates/deployment-clamav.yaml
```diff
@@ -0,0 +1,71 @@
...
+      maxUnavailable: 50%
+  selector:
+    matchLabels:
+      app: submit-crime-forms-clamav
+  template:
+    metadata:
+      labels:
+        app: submit-crime-forms-clamav
+        tier: backend
+    spec:
```
elken, line 27: We'll almost certainly want `allowPrivilegeEscalation: false` and `capabilities: drop: ["ALL"]`, otherwise if this deploys and runs fine it should be good

patrick-laa: @elken I originally tried having those here, and got `ValidationError(StatefulSet.spec.template.spec.securityContext): unknown field "allowPrivilegeEscalation"` and `ValidationError(StatefulSet.spec.template.spec.securityContext): unknown field "capabilities"`, so I moved those things to line 38 (the security context of the container as opposed to the template) and they seemed to work fine.

elken: Oh yes I see, okay I'm happy then. 

Do we want Joel or someone from cloud platform to look this over?

jogilder: @patrick-laa and @elken we could but I think if it deploys then we are unlikely to get any more from CP.

jogilder (approved):
LGTM assuming the TODOs come out

PR #1233, opened 2024-10-11
CRM457-2142: Tweak additional cost deletion flow

patrick-laa:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2142)


spec/system/prior_authority/additional_costs_spec.rb
```diff
@@ -61,12 +61,24 @@
...
 
+        context 'when I have added multiple costs' do
+          before do
+            create :additional_cost, prior_authority_application: PriorAuthorityApplication.first
+          end
+
+          it 'removes the cost then takes me back to the cost list' do
+            click_on 'Delete'
+            click_on 'Yes, delete it'
+            expect(page).to have_content 'You have deleted the additional cost'
```
elken, line 77: It's not relevant to this PR but I have noticed that we dont' really assert the path of the page much.

This might help times when we refactor a path (like the recent turboframe stuff) and the tests pass but things still break. Is there a good reason for this that I'm not aware of? 😄

elken (approved):

PR #1236, opened 2024-10-11
CRM457-2064: Get office codes from provider data API

patrick-laa:
## Description of change
- Stop trusting portal to tell us correct office codes on each login
- Instead use the provider data API (which we also query about which office codes are active, so that it becomes the single point of truth about office codes)

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2064)

### DIFF
--- a/app/services/provider_data_api_client.rb
+++ b/app/services/provider_data_api_client.rb
@@ -1,0 +1,53 @@
+class ProviderDataApiClient
+  class << self
+    def contract_active?(office_code)
+      response = HTTParty.get("#{base_url}/provider-office/#{office_code}/office-contract-details",
+                              headers: { 'X-Authorization': api_key })
+
+      case response.code
+      when 200
+        true
+      when 204
+        false
+      else
+        raise "Unexpected status code #{response.code} when checking office code #{office_code}"
+      end
+    end
+
+    def user_details(email)
+      response = HTTParty.get("#{base_url}/provider-users/exists?userEmail=#{email}",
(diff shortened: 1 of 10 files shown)


app/services/provider_data_api_client.rb
```diff
@@ -0,0 +1,47 @@
...
+      query(
+        "provider-users/#{user_login}/provider-offices",
+        200 => ->(data) { data['officeCodes'] },
+        204 => ->(_) { [] },
+      )
+    end
+
+    private
+
+    def query(endpoint, return_values)
```
elken, line 28: Is there a reason `return_values` can't be a map? if the value is a function call and return it, otherwise return it

elken (approved):

app/services/provider_data_api_client.rb
patrick-laa: Good shout, it's barely more code inside `query`, and makes calling it much cleaner.

patrick-laa:
There are two known issues with this. The first is that the list of office codes returned by the PDA is not the same as the list returned from Portal for a given user (see [thread](https://mojdt.slack.com/archives/C06F768A3F1/p1728898304689929)). The second is that Portal gives us an email address, and we swap that for a username to then retrieve the office codes. But it turns out that one email address can be associated with multiple usernames, and we don't know which one the user just logged in as (see [thread](https://mojdt.slack.com/archives/C06F768A3F1/p1728896992506409))

PR #1238, opened 2024-10-14
Experiment: With minimal config is modsec phase 4 still broken?

patrick-laa:
Yes: https://crm457-1721-b-expl-nscc-provider-dev.cloud-platform.service.justice.gov.uk/assets/file-upload-b450e66e.js.map is triggering a modsec false positive in phase 4 against rule 952100. Modsec is trying to block it but the file is still being sent to the requester, just with a weird header.


patrick-laa:
Closing as this has been addressed (by switching off response body scanning with `SecResponseBodyAccess Off` in the default CP modsec config)

PR #1250, opened 2024-10-16
Adds moj scrollable pane

imtiazAhmd:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2150)

### Before changes:
dev submit-crime-forms service justice gov uk_prior-authority_applications(iPhone SE)

### After changes:
ezgif-1-0d0fddaa89

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -15,6 +15,7 @@
 @import "@ministryofjustice/frontend/moj/utilities/hidden";
 @import "@ministryofjustice/frontend/moj/components/multi-file-upload/multi-file-upload";
 @import "@ministryofjustice/frontend/moj/components/banner/banner";
+@import "@ministryofjustice/frontend/moj/objects/_scrollable-pane";
 
 .govuk-button-group.confirmation-button-group > form {
     display: inline-flex;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -15,7 +15,12 @@
 @import "@ministryofjustice/frontend/moj/utilities/hidden";
 @import "@ministryofjustice/frontend/moj/components/multi-file-upload/multi-file-upload";
 @import "@ministryofjustice/frontend/moj/components/banner/banner";
+@import "@ministryofjustice/frontend/moj/objects/_scrollable-pane";
 
 .govuk-button-group.confirmation-button-group > form {
     display: inline-flex;
 }
+
+.fixed-to-right {
```
elken, line 24: So this is a weird one that seems to only affect Chrome and I wonder why this isn't upstreamed. `govuk-visually-hidden` should probably also have `inset: 0` applied to it, I might query that but in the meantime yeah this works.

Worth adding a comment maybe?

elken (approved):

app/assets/stylesheets/application.scss
imtiazAhmd: I was wondering that too, but in the `govuk-visually-hidden` we have `position: absolute` and no `inset` set. It only acts weird when placed in a `scrollable-pane` which is MOJ component and not GOVUK component. May be thats why it's never been flagged.

elken: `inset` is just shorthand for `top`, `left`, `bottom` and `right`, so just a more complete version of what you had.

And yes perhaps, I'll email them shortly.

PR #1252, opened 2024-10-16
Scrollable-Pane- Fixes error when class doesn't exist

imtiazAhmd:
## Description of change

Fixes to JS script for scrollable-pane, So it only calls if scrollabe-pane element exists.

## Notes to Reviewer

Further optimisation to the script and updates jest test

### DIFF
--- a/app/javascript/scrollable_pane.js
+++ b/app/javascript/scrollable_pane.js
@@ -1,40 +1,33 @@
 export function handleResizeOnScrollablePane(scrollablePane) {
+  if (!scrollablePane) return;
   const isLargeScreen = window.innerWidth >= 1024;
 
-  if (isLargeScreen) {
-    applyLargeScreenStyles(scrollablePane);
-  } else {
+  isLargeScreen ?
+    applyLargeScreenStyles(scrollablePane) :
     applySmallScreenStyles(scrollablePane);
-  }
 }
 
 function applyLargeScreenStyles(element) {
-  if (!element.classList.contains('hide-horizontal-scroll')) {
-    element.classList.add('hide-horizontal-scroll');
-  }
-
+  element.classList.add('hide-horizontal-scroll');
   // Force repaint on Safari by toggling overflow properties
   element.style.overflow = 'hidden';
   element.style.overflowY = 'hidden';
 
-  requestAnimationFrame(() => {
-    element.style.overflow = 'auto';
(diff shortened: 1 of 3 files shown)


app/javascript/scrollable_pane.js
```diff
@@ -1,18 +1,40 @@
...
+  if (element.classList.contains('hide-horizontal-scroll')) {
+    element.classList.remove('hide-horizontal-scroll');
+  }
+
+  element.style.overflow = 'scroll';
+  element.style.overflowY = 'auto';
+}
+
+export function checkAndHandleResizeOnScrollablePane() {
+  const scrollablePane = document.querySelector('.moj-scrollable-pane');
```
patrick-laa, line 36: It might make this more robust if instead of trying to find one pane and then check `if` it exists, we do  something like:
```js
MOJFrontend.nodeListForEach(document.querySelector('.moj-scrollable-pane'), (scrollablePane) => {
  handleResizeOnScrollablePane(scrollablePane);
}
```

That way we're future-proofed against more than one pane being on the same page.

patrick-laa (approved):

app/javascript/scrollable_pane.js
```diff
@@ -1,18 +1,32 @@
-export function handleResizeOnScrollablePane() {
-  const scrollablePane = document.querySelector('.moj-scrollable-pane');
+export function handleResizeOnScrollablePane(scrollablePane) {
+  if (!scrollablePane) return;
+  const isLargeScreen = window.innerWidth >= 1024;
 
-  if (window.innerWidth >= 1024) {
-      scrollablePane.classList.add('hide-horizontal-scroll');
-      // Force repaint on Safari by changing overflow values slightly
```
patrick-laa, line 6: I think it's really worth keeping this comment. Otherwise it's not at all clear why `hide-horizontal-scroll` can't just be controlled by a media query in CSS instead of javascript, why the overflow styles can't just be added to the hide-horizontal-scroll CSS class, and why we need a setTimeout.

To be honest, I still don't really understand why all this JS is needed, but the comment about force repainting on Safari makes me trust that it's not just because we don't understand how CSS works!

imtiazAhmd: Added the comment back, now in Line 12

patrick-laa (approved):

PR #1253, opened 2024-10-17
CRM457-2169: Update email validation to match Notify

patrick-laa:
## Description of change
Make sure that if the user enters an email that would be rejected by Notify, we reject it too

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2169)


jogilder (approved):
LGTM

PR #1255, opened 2024-10-17
CRM457-2106: NSM search

patrick-laa:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2106)

### DIFF
--- a/app/assets/stylesheets/custom/search.scss
+++ b/app/assets/stylesheets/custom/search.scss
@@ -1,3 +1,16 @@
+/* override due to nesting govuk-width-container result in no visible margin */
+@media (min-width: 1020px) {
+  .moj-primary-navigation__container {
+    margin: 0 10px !important;
+  }
+}
+
+@media (min-width: 40.0625em) {
+  .moj-primary-navigation__container {
+    margin: 0 10px !important;
+  }
+}
+
 .search-panel {
   padding: 15px 20px 1px 20px;
   background-color: govuk-colour("light-grey");
--- a/app/assets/stylesheets/custom/search.scss
+++ b/app/assets/stylesheets/custom/search.scss
@@ -1,11 +1,6 @@
 /* override due to nesting govuk-width-container result in no visible margin */
-@media (min-width: 1020px) {
-  .moj-primary-navigation__container {
-    margin: 0 10px !important;
-  }
-}
-
-@media (min-width: 40.0625em) {
+@media (min-width: 1020px),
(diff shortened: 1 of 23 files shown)


app/assets/stylesheets/custom/search.scss
```diff
@@ -1,3 +1,16 @@
+/* override due to nesting govuk-width-container result in no visible margin */
+@media (min-width: 1020px) {
+  .moj-primary-navigation__container {
+    margin: 0 10px !important;
+  }
+}
+
+@media (min-width: 40.0625em) {
```
elken, line 8: You can use multiple media queries for a block

```css
@media (min-width: 1020px),
       (min-width: 16rem) // same as the em value you have
```

app/forms/search_form.rb
```diff
@@ -1,23 +1,49 @@
...
-    @params.key?(:search_string)
+  def office_codes
+    [show_all] + current_provider.office_codes.map { Option.new(_1, _1) }
   end
 
-  def valid?
-    @params[:search_string].present?
+  def states
+    [show_all] + %i[
+      draft
```
elken, line 34: This is just all statuses no? Shouldn't this be a global array?

app/javascript/date-picker.js
```diff
@@ -0,0 +1,13 @@
+
+import MOJFrontend from '@ministryofjustice/frontend'
+import $ from 'jquery'
+window.$ = $
+
+function init() {
+  const $datepickers = document.querySelectorAll('[data-module="moj-date-picker"]')
```
elken, line 7:
```js
[...document.querySelectorAll('[data-module="moj-date-picker"]')].forEach(picker => {
    new MOJFrontend.DatePicker(picker, {}).init()
});
```

elken (approved):
Couple of smaller comments, but otherwise this looks good

app/assets/stylesheets/custom/search.scss
patrick-laa: Oh cool, I didn't know that.

elken: Yeah since it's the same rule, works the same as multiple selectors (since they are just selectors really)

app/forms/search_form.rb
elken: Ah I didn't consider order, okay then. Just seems like we're bound to forget one some time.

PR #1258, opened 2024-10-21
CRM457-2106: Backfill timestamps

patrick-laa:
## Description of change
Now that we're tracking when a submission was originally submitted (to filter on by search) we need to backfill all of these timestamps for previous submissions. I got the from the timestamp when the production app store submission record was originally created, and dumped them into CSVs. We iterate through and update. There's only a few thousand so it only takes a few seconds, so I'm not super concerned about efficiency here.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2106)


db/migrate/20241021092835_backfill_originally_submitted_at.rb
```diff
@@ -0,0 +1,13 @@
+require 'csv'
+
+class BackfillOriginallySubmittedAt < ActiveRecord::Migration[7.2]
+  def change
+    CSV.parse(File.open(Rails.root.join('db/migrate/20241021_nsm_originally_submitted.csv'))).each do |id_and_timestamp|
```
elken, line 5: Minor nit but you should be able to do `|timestamp, id|`

elken (approved):

robinsk1 (approved):

ivanELEC (approved):

PR #1261, opened 2024-10-22
CRM457-2099: Improve disbursement flows

elken:
## Description of change

Adjust various flows around disbursements in NSM, similar PR to #1239 which a lot of this is borrowed from.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2099)


app/controllers/nsm/steps/disbursement_cost_controller.rb
```diff
@@ -29,7 +32,24 @@ def ensure_disbursement
...
+      end
+
+      def new_record?
+        disbursement.disbursement_date.nil? || disbursement.miles.nil?
+      end
+
+      def build_flash
+        if new_record?
+          if params.dig(:nsm_steps_disbursement_cost_form, :add_another) == YesNoAnswer::YES.to_s
+            count = current_application.disbursements.count + 1
```
patrick-laa, line 45: This won't work. When you're adding your first disbursement, by this point it already exists in the database so the `current_application.disbursements.count` is 1. Adding 1 to it gets you 2, so on the next screen you'll see "You've added 2 disbursements" when in fact you've only added 1.

app/controllers/nsm/steps/disbursement_cost_controller.rb
```diff
@@ -29,7 +32,24 @@ def ensure_disbursement
       end
 
       def additional_permitted_params
-        [:apply_vat]
+        [:apply_vat, :add_another]
+      end
+
+      def new_record?
+        disbursement.disbursement_date.nil? || disbursement.miles.nil?
```
patrick-laa, line 39: This is buggy. For "other" disbursement types, disbursement.miles will always be nil, because for those you just enter the total cost, not the mileage. so this will always return true.

app/controllers/nsm/steps/disbursement_cost_controller.rb
```diff
@@ -29,7 +32,24 @@ def ensure_disbursement
...
+      end
+
+      def build_flash
+        if new_record?
+          if params.dig(:nsm_steps_disbursement_cost_form, :add_another) == YesNoAnswer::YES.to_s
+            count = current_application.disbursements.count + 1
+            { success: t('.added', count: count, disbursements: t('.disbursement').pluralize(count)) }
```
ivanELEC, line 46: I don't think this needs the "+ 1" - at the point of a user saying they want to add another disbursement, they have not yet completed the additional disbursement. 

An example of this is if you click add another for the first time, it reports back "You have added 2 disbursements" when you've only added one and are currently trying to add the 2nd

spec/system/nsm/disbursements_spec.rb
```diff
@@ -145,14 +193,15 @@
     fill_in 'Number of miles', with: 100
     fill_in 'Enter details of this disbursement', with: 'details'
 
+    expect(page).to have_content 'Do you need to add another disbursement?'
+    choose 'No'
+
     click_on 'Save and continue'
 
-    expect(page).to have_content 'Do you want to add another disbursement?'
+    expect(page).to have_no_content 'You cannot save and continue if any disbursements are incomplete'
```
patrick-laa, line 201: I think you want this expectation to happen after the final 'Save and continue'. At this current point you've just submitted the disbursement cost form, so _of course_ this message won't be here, because the disbursement cost form has no mechanism for adding that message. Whereas after line 204 below you've just submitted the DisbursementsForm which could generate this message but hopefully won't because there are now no incomplete disbursements. Do it's there that we want to verify that this message doesn't appear.

spec/system/nsm/disbursements_spec.rb
```diff
@@ -280,6 +378,12 @@
 
       expect(all('table td, table th').map(&:text)).to eq(
         [
+          'Item', 'Total claimed',
```
patrick-laa, line 381: To be the endless repetition of these same snippets across a bunch of tests that aren't about the snippets is just tech debt. It makes the tests harder to read and brittler. Is there any way we can switch out `all('table td, table th').map(&:text)` for something a bit more targeted so these tests can go back to just testing the things they care about?

patrick-laa (approved):

app/controllers/nsm/steps/disbursement_cost_controller.rb
elken: That makes sense, but I wasn't sure what else to use to determine if it's a new record or not.

patrick-laa: You need an `&&`, not an `||`. A new record will have neither a number of miles nor a `total_cost_without_vat`, so if _both_ are nil then you know this is new.

elken: Ahh you're right

patrick-laa (approved):

PR #1262, opened 2024-10-22
Change input type of period field to number

elken:
## Description of change

Change the input type of the period field to be number instead of text since it's intended that numbers be entered.

### DIFF
--- a/gems/laa_multi_step_forms/app/lib/govuk_design_system_formbuilder/elements/period.rb
+++ b/gems/laa_multi_step_forms/app/lib/govuk_design_system_formbuilder/elements/period.rb
@@ -105,7 +105,7 @@ module GOVUKDesignSystemFormBuilder
           id: id(segment, link_errors),
           class: classes(width),
           name: name(segment),
-          type: 'text',
+          type: 'number',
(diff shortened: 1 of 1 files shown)


elken:
> this makes total sense ..just curious but were there any downstream hacks/fixes in view etc for this that this change would now make obsolete?

Not AFAICT, the big issue is I don't know if this impacts anything major. Was going to ping patrick once this was ready to review

patrick-laa (changes requested):
Weirdly I think we actually don't want to do this: https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/ and https://design-system.service.gov.uk/components/text-input/#avoid-using-inputs-with-a-type-of-number

elken:
> Weirdly I think we actually don't want to do this: https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/ and https://design-system.service.gov.uk/components/text-input/#avoid-using-inputs-with-a-type-of-number

That's what I was worried about, annoying but understandable. Thanks for that.

PR #1271, opened 2024-10-24
CRM457-2168: Clamby not updating

patrick-laa:
## Description of change
We have been trying to update ClamAV's virus definitions nightly by running `Clamby.update` in our Ruby code. However this silently fails because  under the hood it's calling`freshclam` locally, and `freshclam` isn't installed locally. There appears to be no way to trigger freshclam remotely from within the Ruby runtime.

So instead I'm adding.a series of cron-triggered Github actions that each trigger `freshclam` via kubectl exec. It appears as though each clamav pod in a namespace has its own set of virus definitions, so I think we need to call it on each pod. So I'm spacing out these 5 calls (1 dev, 2 uat, 2 prod) over 20 minutes to minimise risk of throttling.

In the longer term we may consider pulling a new clamav image for the pods each day but TBH I don't know how to do that so 🤷  

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2168)


ivanELEC:
FYI  - Supposedly there will be movement within weeks to start using Secure Document Storage utilities shared across LAA services. Part of Secure Document Storage will have it's own ClamAV service that handles definition updates etc that will be shared between services and that supposedly will be ready earlier - so it's actually quite likely we will be able to remove our Clam AV pods and use what Secure Document Storage provides in the actual near future.

ivanELEC (approved):

PR #1273, opened 2024-10-25
CRM457-2038 Accessibility issues on Disbursement cost page

imtiazAhmd:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2038)


app/views/nsm/steps/claim_details/edit.html.erb
```diff
@@ -6,9 +6,9 @@
     <%= govuk_error_summary(@form_object) %>
     <h1 class="govuk-heading-xl"><%= t('.heading') %></h1>
     <%= step_form @form_object do |f| %>
-      <%= f.govuk_number_field :prosecution_evidence,label:{ size: 's'}, min: 1, step: 1, width: 3 %>
-      <%= f.govuk_number_field :defence_statement,label:{ size: 's'}, min: 1, step: 1, width: 3 %>
-      <%= f.govuk_number_field :number_of_witnesses,label:{ size: 's'}, min: 1, step: 1, width: 3 %>
+      <%= f.govuk_text_field :prosecution_evidence,label:{ size: 's'}, width: 3 %>
```
patrick-laa, line 9: Couple of things with these. First off, if we're using text fields for numbers we should include `inputmode=numeric` as this prompts mobile devices to show the right sort of keyboard.

Second, now that these fields do accept non-numeric inputs, the request from Mary in PA was that if a user enters a non-numeric value and submits the form, the thing they originally entered should be pre-populated in the field along with a validation error message. As it stands this won't happen, because all these attributes are marked as `:integer` in the form classes, meaning they'll get parsed to integers before validation even happens, so will be converted to `0`, which may not even trigger a validation error, and may not be what the provider intended to enter.

The way around this one is to change the attribute type from `:integer` to `:fully_validatable_integer`. But note that when you do that, you need to validate `is_a_number: true` and add a `not_a_number` error string.

Finally, it'd be worth checking the existing validations on all these fields. E.g. the message for number_of_hearing if you put a 0 in is `must be greater than 0`, i.e. there's no explicit error message defined so Rails is making up its own, bad,  one.

app/forms/nsm/steps/disbursement_cost_form.rb
```diff
@@ -4,13 +4,13 @@ class DisbursementCostForm < ::Steps::BaseFormObject
       include DisbursementCosts
       attr_writer :apply_vat
 
-      attribute :miles, :decimal, precision: 10, scale: 3
-      attribute :total_cost_without_vat, :decimal, precision: 10, scale: 2
+      attribute :miles, :fully_validatable_decimal, precision: 10, scale: 3
+      attribute :total_cost_without_vat, :fully_validatable_decimal, precision: 10, scale: 2
```
patrick-laa, line 8: You could use `:gbp` here, which has the advantage of being forgiving of if providers enter the "£" symbol in the field itself

app/forms/nsm/steps/disbursement_cost_form.rb
```diff
@@ -4,13 +4,13 @@ class DisbursementCostForm < ::Steps::BaseFormObject
...
       attribute :details, :string
       attribute :prior_authority, :value_object, source: YesNoAnswer
 
-      validates :miles, presence: true, numericality: { greater_than: 0 }, unless: :other_disbursement_type?
-      validates :total_cost_without_vat, presence: true, numericality: { greater_than: 0 },
+      validates :miles, presence: true, is_a_decimal: { greater_than: 0 }, unless: :other_disbursement_type?
```
patrick-laa, line 12: How come we can't do what we do for fully_validatable_integers and gbps, i.e. `validates :miles, presence: true, is_a_number: true, numericality: { greater_than: 0 }`? I'm not sure I understand why we need to add an IsADecimal validator when `is_a_number` and `numericality` together give us what we need.

app/forms/nsm/steps/disbursement_cost_form.rb
```diff
@@ -19,19 +19,27 @@ def apply_vat
...
-            html_attributes: { id: 'total-without-vat' }
-          },
-           {
-             text: NumberTo.pounds(total_cost),
-             html_attributes: { id: 'total-with-vat' },
-           }],
-        ]
+        # this logic is needed to prevent invalid values erroring in the NumberTo methods
+        if total_cost_pre_vat.is_a?(String)
+          record.errors.add(:total_cost_without_vat, :not_a_decimal)
```
patrick-laa, line 26: I don't think we should be doing any `errors.add` stuff anywhere other than in validation methods. It's a weird side effect to add, and a bit unpredictable, because if we call `calculation_rows` after we've rendered the relevant bits of the form it will have no effect, but if we call it beforehand it won't. Plus this error should be added by the existing validation, right?

app/forms/nsm/steps/disbursement_cost_form.rb
```diff
@@ -19,19 +19,27 @@ def apply_vat
...
+          []
```
patrick-laa, line 27: I'm also not sure we want this method to return an empty array, as that will result in the UI doing something new, i.e. showing a table with no body. We probably want this to do what it would do if the `total_cost_without_vat` string was left blank, i.e. return the same structure just with empty `text` attributes.

app/forms/nsm/steps/disbursement_cost_form.rb
ivanELEC: thanks for interrogating this, the problem was due to code I fixed when I was adding tests (weird casting issues in the decimal type) and I can actually get ride of the IsADecimal validator and the other code you mentioned below on the calculation_rows method completely

app/forms/nsm/steps/disbursement_cost_form.rb
```diff
@@ -23,11 +23,11 @@ def calculation_rows
         [
           [translate(:before_vat), translate(:after_vat)],
           [{
-            text: NumberTo.pounds(total_cost_pre_vat),
+            text: NumberTo.pounds(total_cost_pre_vat.to_f),
```
patrick-laa, line 26: This is still changing the behaviour slightly. When you first load the form page for an "other" disbursement type, total_cost_without_vat is nil, so previously `total_cost_pre_vat` would be nil and so we'd pass in nil to `NumberTo` and that would result in the UI showing "£". Now, because you're calling `.to_f`, when `total_cost_pre_vat` is nil, we end up passing 0 (`nil.to_f`) to NumberTo, so what shows up in the UI is "£0.00". 

If you want to keep it the same you'd want something like `NumberTo.pounds(total_cost_pre_vat unless total_cost_pre_vat.is_a?(String)`

ivanELEC: added a method that does this

app/views/nsm/steps/disbursement_cost/edit.html.erb
```diff
@@ -7,19 +7,20 @@
     <h1 class="govuk-heading-l"><%= t('.heading') %></h1>
     <%= step_form @form_object do |f| %>
       <% if @form_object.other_disbursement_type? %>
-        <%= f.govuk_number_field :total_cost_without_vat, min: 1, step: 0.01, width: 5, label: { tag: 'strong' }, data: { selector: 'valueField', multipler: @form_object.multiplier } %>
+        <%= f.govuk_text_field :total_cost_without_vat, width: 5, label: { tag: 'strong' }, data: { selector: 'valueField', multipler: @form_object.multiplier }, inputmode: 'numeric' %>
```
patrick-laa, line 10: Sorry, another nit: As these are decimal fields the inputmode should be `decimal`, not `numeric`, so that on mobile the soft keypad includes a decimal point as an option

app/views/nsm/steps/disbursement_cost/edit.html.erb
```diff
@@ -7,19 +7,20 @@
     <h1 class="govuk-heading-l"><%= t('.heading') %></h1>
     <%= step_form @form_object do |f| %>
       <% if @form_object.other_disbursement_type? %>
-        <%= f.govuk_number_field :total_cost_without_vat, min: 1, step: 0.01, width: 5, label: { tag: 'strong' }, data: { selector: 'valueField', multipler: @form_object.multiplier } %>
```
patrick-laa, line 10: Also, you want to add `value: gbp_field_value(@form_object.total_cost_without_vat)` to this. That way if a user has entered e.g. "9.50", it won't prepopulate the field as "9.5"

ivanELEC: true and on both accounts and updated!

elken (approved):
1 minor q, otherwise think Patrick is happy now

app/models/concerns/letters_and_calls_costs.rb
```diff
@@ -74,10 +74,10 @@ def letters_and_calls_total_cost_inc_vat
   private
 
   def letters_before_uplift
-    letters.to_d * pricing.letters if letters && !letters.zero?
+    letters.to_d * pricing.letters if letters && !letters.to_i.zero?
   end
 
   def calls_before_uplift
-    calls.to_d * pricing.letters if calls && !calls.zero?
+    calls.to_d * pricing.letters if calls && !calls.to_i.zero?
```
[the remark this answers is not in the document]
ivanELEC: I added the to_i on the 2nd part of the clauses because with the changes to the form (use can put in non numeric characters into the field), these methods will cause a 500 error if you try to run a calculation via clicking the calculate changes button (as this is pre-validation). So adding the to_i makes it so that invalid entries just change to 0 instead of error-ing out 

My assumption re: the .to_d (which was already there), is so that the type of the calls value matches the type of the pricing value (which is a decimal). A decimal multiplied by an integer _should_ ouput a decimal - but I guess this just makes it more secure if Ruby decides to be weird and change how that works?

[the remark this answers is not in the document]
ivanELEC: Re: the first paragraph - there could potentially be a better experience re: the calculate changes but thought that should be a separate ticket

robinsk1 (approved):

PR #1276, opened 2024-10-28
Completely remove sidekiq-cron gem

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2206)

Removes sidekiq-cron gem as it's currently not needed and is causing deleted scheduled jobs to run when the schedule file has been deleted

## Notes for reviewer
This PR is to be reverted if it causes any problems when cron jobs would have run tomorrow


elken (approved):
Approved with the note, if this breaks anything it has to be reverted ASAP

PR #1277, opened 2024-10-28
[CRM457-2207] add user_chosen_cost_type to app-store blob and calculate correct quote allowed totals

robinsk1:
## Description of change

If a provider selects a custom service type in their claim we are not including `user_chosen_cost_type` to the app-store submission. This means that when syncing back to the provider the allowed totals for part_grants with custom service types are not showing the correct allowed total if there are adjustments. 

When a case-worker updates the submission the provider app is unable to determine the cost type for the item with the net result of returning zero allowed costs in the provider claim summary UI

This PR adds  `user_chosen_cost_type` during the sync process using the already present `cost_type` attribute retrieved form the app-store payload. With `user_chosen_cost_type` present the `ServiceCostForm` form object correctly calculates the allowed base totals in the provider claim summary UI

This PR also includes a rake task that identifies and fixes all past primary_quotes that would have been affected by this bug. This fix re-runs syncing from the app-store using the now included `user_chosen_cost_type` within the synced data. The fix is name-spaced using the the jira ticket number for traceability/context

This fix can be run in production via `rake CRM457_2207:Fix`

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2207)

### DIFF
--- a/app/services/prior_authority/assessment_syncer.rb
+++ b/app/services/prior_authority/assessment_syncer.rb
@@ -45,6 +45,8 @@ module PriorAuthority
     end
 
     def sync_primary_quote(quote, quote_data)
+      quote_data['user_chosen_cost_type'] = quote_data['cost_type']
+
       base_cost_form = build_form(application, quote, Steps::ServiceCostForm, quote_data)
       travel_cost_form = build_form(application, quote, Steps::TravelDetailForm, quote_data)
 
--- a/app/services/submit_to_app_store/prior_authority/quote_payload_builder.rb
+++ b/app/services/submit_to_app_store/prior_authority/quote_payload_builder.rb
@@ -28,7 +28,6 @@ class SubmitToAppStore
         {
           cost_type: service_cost_form.cost_type,
           item_type: service_cost_form.item_type,
-          user_chosen_cost_type: service_cost_form.user_chosen_cost_type,
           cost_item_type: service_cost_form.cost_item_type,
           cost_multiplier: service_cost_form.cost_multiplier
         }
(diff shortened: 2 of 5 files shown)


app/services/prior_authority/assessment_syncer.rb
```diff
@@ -45,6 +45,8 @@ def sync_allowances
     end
 
     def sync_primary_quote(quote, quote_data)
+      quote_data['user_chosen_cost_type'] = quote_data['cost_type']
```
patrick-laa, line 48: I would like to see this covered by a test. As it stands, if there was a typo in one of these strings I don't think anything would fail.

spec/forms/prior_authority/steps/service_cost_form_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      items: 2,
+      cost_per_item: 3,
+      period: nil,
+      cost_per_hour: nil,
+      application: application,
+      record: application.primary_quote }
+  end
+
+  describe '#cost_type' do
+    it 'returns cost_type' do
```
patrick-laa, line 19: Nit: This is not a good test. First of all, the text says "#cost_type returns cost_type", which is meaningless. Second, the code appears to demonstrate that #cost_type returns the input `user_chosen_cost_type`. But this only happens in a very specific set of circumstances, namely if the service_type is associated with a variable cost type, which _happens_ to be the case with the test fixture you're using but isn't made at all explicit in the test. So as it stands this is pretty misleading.

app/services/prior_authority/assessment_syncer.rb
robinsk1: test added, let me know if thats enough to cover it

spec/forms/prior_authority/steps/service_cost_form_spec.rb
robinsk1: yeah, binned

ivanELEC (approved):

PR #1284, opened 2024-10-30
CRM457-2198: Start using pricing gem

patrick-laa:
## Description of change
- Use the gem as the source of truth for pricing
- Use the gem as the source of truth for NSM calculations
- Because we don't want to have to trust calculated fields in the payload (the schema now marks calculated fields and pricing fields as deprecated), as they should always be recalculated dynamically, add a new `high_value` boolean attribute to the payload, as a half-way-house towards getting this stuff out entirely but also enabling the app store and caseworker app to filter in/out high value claims

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2198)


app/forms/nsm/steps/letters_calls_form.rb
```diff
@@ -39,13 +39,13 @@ def letters_row
         [
           translate(:letters),
           {
-            text: NumberTo.pounds(letters_before_uplift),
+            text: NumberTo.pounds((letters_before_uplift if letters.presence)),
```
ivanELEC, line 42: this might be me being dim and not seeing something obvious, but where is this letters method coming from?

patrick-laa: It's an attribute defined on line 7. `letters_before_uplift` used to be nil of `letters` hadn't been set, but now it will be `0`, so to preserve existing behaviour I'm checking for `letters` being set here.

app/services/risk_assessment/high_risk_assessment.rb
```diff
@@ -6,17 +6,14 @@ class HighRiskAssessment
...
     end
 
     def assess
       high_cost? || counsel_assigned? || enhanced_rates_claimed? || extradition? || rep_order_withdrawn?
     end
 
     def high_cost?
-      profit_cost_summary = @summary.table_fields(formatted: false).find { _1[:name] == 'profit_costs' }
-      profit_cost_summary[:gross_cost] >= HIGH_RISK_THRESHOLD
+      @claim.totals[:cost_summary][:profit_costs][:claimed_total_inc_vat] >= HIGH_RISK_THRESHOLD
```
ivanELEC, line 16: Since this threshold is shared between the caseworker and provider apps - would it make sense to encorporate it into the common gem in the future?

patrick-laa: Thoroughly agree! I've pitched doing exactly that in my "Move the biz logic around" ADR: https://dsdmoj.atlassian.net/wiki/spaces/CRM457/pages/5246943990/Draft+ADR+16+Business+logic+placement#%E2%9D%8C-D.-The-provider-app-is-currently-responsible-for-flagging-claims-as-high-risk

ivanELEC (commented):
Just want to confirm my thinking re: the "steps" needed for applying the new library rather than look at the code in an isolated way: 

1. The LaaCrimeFormsCommon::Pricing module completely replaces the pricing library and therefore the library and config for the library is removed - any instance where we would have used the pricing library is replaced by it's respective method/attribute from the new common module

2. The claim model is updated so we get total values and rates as individual methods that can be used to replace the pricing module. The method's data_for_calculation and full_data_for_calculation are used to format data from the claim so that they can be input into the rates and totals method respectively.
3. A similar process for step 2 is applied to the disbursment_costs, letters_and_calls_costs and work_item_costs concerns. The rates method that is already in the claim is delegated so that it can be reused in the concerns. 
4. With these additional methods ready for the claim and individual costing items, we replace pricing calculations in the forms, presenters and views (also using some delegation to reuse methods from the model, rather than replicate logic) - also some additional cleanup on the views

ivanELEC (approved):

ivanELEC:
Taking back the approval until the common gem is approved and merged and we've done the todo in the gemfile (to prevent us accidentally merging without doing that)

elken:
You have to actually dismiss the review to undo it
image

patrick-laa:
I've now removed the TODO

PR #1287, opened 2024-11-05
[CRM457-2239] fix draft pa search link

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2239)

### DIFF
--- a/app/views/prior_authority/applications/search.html.erb
+++ b/app/views/prior_authority/applications/search.html.erb
@@ -39,10 +39,10 @@
         </thead>
         <tbody class="govuk-table__body app-task-list__items">
           <% @model.each do |application| %>
-            <% @link = application.state == 'draft' ? prior_authority_steps_start_page_path(application) : prior_authority_application_path(application) %>
+            <% link = application.draft? ? prior_authority_steps_start_page_path(application) : prior_authority_application_path(application) %>
             <tr class="govuk-table__row app-task-list__item">
               <td class="govuk-table__cell">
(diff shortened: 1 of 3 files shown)


ivanELEC (approved):
Would be good to have some additional tests to ensure the bug that happened gets caught, but I won't let that opinion hold up merging

app/views/prior_authority/applications/search.html.erb
```diff
@@ -39,9 +39,10 @@
         </thead>
         <tbody class="govuk-table__body app-task-list__items">
           <% @model.each do |application| %>
+            <% @link = application.state == 'draft' ? prior_authority_steps_start_page_path(application) : prior_authority_application_path(application) %>
```
patrick-laa, line 42: Nits:

- as this is in an `each` loop you don't really want to be setting an ivar here. `link` would be better than `@link` as otherwise once you exit the loop you'll have a random value sitting around that could mess us up if elsewhere in the view or controller we want an actual `@link` ivar in future.

- It's slightly less error-prone to use `application.draft?` instead of `application.state == 'draft'` - shorter and will blow up if there are typos

- I think the best fix would be to handle the situation where somehow a user tries to visit `prior_authority_application_path` for a draft application. Doing what NSM does and gracefully redirecting means that if we ever have another situation where we accidentally link to the wrong place it won't blow up again.

- Agree with Ivan that tests would be valuable.

PR #1291, opened 2024-11-07
CRM457-2248: Escape the query to prevent syntax errors

elken:
## Description of change
Include the same change from https://github.com/ministryofjustice/laa-crime-application-store/pull/312 to ensure that we protect against incorrect syntax error exceptions.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2248)

### DIFF
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -10,15 +10,20 @@ class SearchService
 
     private
 
+    def escape_string(string)
+      %('''#{string.gsub("'", "''")}''')
+    end
+
     def filter_on_search_string(base_query, params)
       return base_query if params[:search_string].blank?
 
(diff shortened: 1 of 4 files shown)


app/services/search_service.rb
```diff
@@ -15,10 +15,11 @@ def filter_on_search_string(base_query, params)
 
       params[:search_string].split.reduce(base_query) do |built_query, token|
         word = token.strip
```
patrick-laa, line 17: The other thing you did on the other PR was to swap `'` for `''`. I think we could stand to do that here, otherwise it breaks if you search for "O'Reilly"

elken: Oh yes so I did

PR #1292, opened 2024-11-07
Nicely format Snyk output

elken:
## Description of change

Use Ruby to generate a nicely formatted output.

image

[Sample step](https://app.circleci.com/pipelines/github/ministryofjustice/laa-submit-crime-forms/6318/workflows/fb3968a1-a7a4-4565-8239-541edc44c41e/jobs/44093)

It'll likely break randomly over time, which is why I've also made the script save the full output as a JSON artifact. Even if the script fails, we can still check and see what broke.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2252)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -356,7 +356,10 @@ jobs:
           command: snyk monitor --all-projects
       - run:
           name: Scan container image for vulnerabilities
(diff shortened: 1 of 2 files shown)


ivanELEC (approved):

ivanELEC:
Appreciate the fallback added if it breaks

PR #1296, opened 2024-11-11
Crm457 1968 file upload

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1968)

## Notes for reviewer

- Generalise single-file-upload-validator to a general enhancements for single file uploads
- Update "uploaded files" section to be dynamic and update when someone selects a few files


app/javascript/single-file-upload.js
```diff
@@ -0,0 +1,117 @@
...
+          }
+        })
+      }
+    }
+  }
+});
+
+function handleUploadList(fileUploader){
+  //change upload progress to indicator
+  let uploadProgressCell = document.querySelector('.moj-multi-file-upload__progress');
```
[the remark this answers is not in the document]
ivanELEC: what error were you getting on the alternative quote screen? I haven't been able to reproduce it. 

I did notice that the file upload validator runs whether a file is selected or not so fixing that now (this happens on both primary quote and alternative quote)

app/javascript/single-file-upload.js
```diff
@@ -0,0 +1,117 @@
...
+</svg>`
+
+const progress =
+`<td class="govuk-table__cell moj-multi-file-upload__progress" data-label="Upload Progress">
+<progress>working</progress>
+</td>`
+
+
+window.addEventListener("DOMContentLoaded", (event) => {
+  const fileUploader = document.querySelector('.govuk-file-upload');
```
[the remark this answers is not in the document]
ivanELEC: updated the naming

app/javascript/single-file-upload.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+});
+
+function handleUploadList(fileUploader){
+  //change upload progress to indicator
+  let uploadProgressCell = document.querySelector('.moj-multi-file-upload__progress');
+  uploadProgressCell.innerHTML = progress;
+  //replace uploaded files text and change indicator to cross if unsuccessful
+  let filenameCell = document.querySelector('.moj-multi-file-upload__filename');
+  let uploadedFile = fileUploader.files[0];
```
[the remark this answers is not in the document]
ivanELEC: removed the progress indicator and the fail state indicator

app/javascript/single-file-upload.js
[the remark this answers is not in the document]
patrick-laa: In the console on your dev branch I'm getting:

```
Uncaught TypeError: Cannot set properties of null (setting 'textContent')
    at handleUploadList (single-file-upload.js:43:18)
    at HTMLInputElement.<anonymous> (single-file-upload.js:16:9)
```

It's as soon as I select a file on the alternative quote screen. Try it yourself at https://crm457-1968-file-u-nscc-provider-dev.cloud-platform.service.justice.gov.uk/prior-authority/applications/402357c9-bfd2-45d4-8802-031791c1966e/steps/alternative_quote_details/new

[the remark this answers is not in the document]
ivanELEC: oh I see, got the error now. Not sure why I didn't get it before as it should happen on any instance of changing the file. Will fix this, but want to talk to Mary first about whether this indicates a need for a more consistent ui between primary quote and alternative quote

app/javascript/multi-file-upload.js
```diff
@@ -11,9 +11,9 @@ MOJFrontend.MultiFileUpload.prototype.uploadFiles = async function(files) {
 
   const uploads = []
 
-  for(var i = 0; i < files.length; i++) {
+  for(const file of files) {
     uploads.push(
-      this.uploadFile(files[i]).then(
+      this.uploadFile(file).then(
         function(response) { console.log(response); }
```
imtiazAhmd, line 17: Should these console logs be removed?

elken: Removed yeah and we need to actually handle whatever comes back error-wise

app/javascript/single-file-upload.js
```diff
@@ -0,0 +1,101 @@
+window.addEventListener("DOMContentLoaded", (event) => {
```
elken, line 1: I'd rather export this inner function and call it than having this run when the file is required. If there's a couple of these on a page it'll add a listener for each one.

app/javascript/single-file-upload.js
```diff
@@ -0,0 +1,101 @@
...
+  const fileInput = document.querySelector('.govuk-file-upload');
+  const maxFileSize = fileInput.dataset.maxSize;
+  const errorHeading = fileInput.dataset.errorHeading;
+  const errorMessage = fileInput.dataset.sizeErrorMessage;
+  const feedback = $(".single-file-upload__message");
+  const saveButtons = document.querySelectorAll('button[type="submit"]');
+
+  if(fileInput){
```
elken, line 9: `document.querySelectorAll('.govuk-file-upload')?.forEach` instead makes the branch redundant. Rather just have an early `if (!fileInput) return;` since we can't do anything without it anyway.

app/javascript/single-file-upload.js
```diff
@@ -0,0 +1,101 @@
...
+        if(fileInput.files.length > 0){
+          handleUploadList(fileInput);
+        }
+    });
+
+    /*
+      handle checking file size, if exceeds limit,
+      refresh with error shown and move page to file upload section.
+    */
+    if (saveButtons) {
```
elken, line 24: Same as above here

app/javascript/single-file-upload.js
```diff
@@ -0,0 +1,101 @@
+window.addEventListener("DOMContentLoaded", (event) => {
+  const fileInput = document.querySelector('.govuk-file-upload');
+  const maxFileSize = fileInput.dataset.maxSize;
+  const errorHeading = fileInput.dataset.errorHeading;
+  const errorMessage = fileInput.dataset.sizeErrorMessage;
+  const feedback = $(".single-file-upload__message");
```
elken, line 6: Don't think this needs to be a jquery call

app/javascript/single-file-upload.js
```diff
@@ -0,0 +1,101 @@
...
+      }
+    }
+  }
+});
+
+function handleUploadList(fileInput){
+  //replace selected file text
+  let filenameCell = document.querySelector('.moj-multi-file-upload__filename');
+  let uploadedFile = fileInput.files[0];
+  if(uploadedFile){
```
elken, line 44: No null check for `filenameCell`

app/javascript/single-file-upload.js
```diff
@@ -0,0 +1,101 @@
...
+
+    return validationPassed;
+  }
+}
+
+function generateFileSizeError(fileInput, errorMessage, errorHeading, feedback) {
+  feedback.html(govukErrorSummary(fileInput.id, errorMessage, errorHeading));
+  addInlineError(fileInput, errorMessage)
+  let errorSummary = document.querySelector('.govuk-error-summary');
+  errorSummary.scrollIntoView();
```
elken, line 64: `errorSummary.scrollIntoView({ behavior: 'smooth' });` will be less visually jarring. Also no null check here.

elken (commented):
Couple of comments, I also wonder if it might just be cleaner as a single function rather than several.

app/javascript/multi-file-upload.js
ivanELEC: I'll remove the console logs for security purposes - but the actual handling of the error seems like a separate tech debt ticket (this is code we already had in the system, the only reason the PR says I added it is because the file name was changed)

app/javascript/single-file-upload.js
ivanELEC: sorry this comment is unclear for me - couple of what on a page?

app/javascript/single-file-upload.js
ivanELEC: [intentionally didn't use forEach here as it doesn't work on NodeLists like it does an Arrays and apparently the only way you can use it is unsafe](https://dev.to/jess/how-do-i-use-foreach-on-dom-elements-3m9h)

app/javascript/single-file-upload.js
ivanELEC: agreed, will update that

app/javascript/single-file-upload.js
ivanELEC: updated

app/javascript/single-file-upload.js
ivanELEC: updated

app/javascript/single-file-upload.js
ivanELEC: see comment above

app/javascript/single-file-upload.js
elken: Components

app/javascript/single-file-upload.js
elken: It does work the same, only doesn't work on older browsers we don't support

app/javascript/single-file-upload.js
ivanELEC: Understand the concerns re: if we had multiple file uploads per page. Currently this javascript is designed for only expecting one file input per page (particularly because there's not really a GDS precedence for forms with multiple single file uploads, so we're not sure if a design would ever look like that) - which is why we use querySelector to get the file input instead of querySelectorAll - this is why I also don't do forEach on the file upload component. forEach does make sense for the saveButtons because we actually expect multiple buttons (Save and continue, Save and come back alter)

ivanELEC: I will add this explicit information as a comment so that it's clearer

app/javascript/single-file-upload.js
ivanELEC: updated this for loop suggestion but see [this comment ](https://github.com/ministryofjustice/laa-submit-crime-forms/pull/1296#discussion_r1839912529)re: the other for loop suggestion

imtiazAhmd (approved):
LGTM

PR #1305, opened 2024-11-18
CRM457-2276: Do things synchronously

patrick-laa:
## Description of change
- Submit to app store synchronously
- Check the app store response to see if there's been an immediate state change (i.e. autogrant)
- Now the only thing that uses Sidekiq is the submission confirmation email

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2276)

### DIFF
--- a/app/jobs/submit_to_app_store.rb
+++ b/app/jobs/submit_to_app_store.rb
@@ -14,8 +14,9 @@ class SubmitToAppStore < ApplicationJob
     # means waiting until the previous lock is released, at which point any
     # modifications made as part of that transaction will have been committed
     submission.with_lock do
-      submit(submission)
+      new_data = submit(submission)
       notify(submission)
+      update_records(new_data, submission)
       submission.update!(submit_to_app_store_completed: true)
     end
   end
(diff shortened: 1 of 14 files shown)


elken (approved):

spec/services/app_store_client_spec.rb
```diff
@@ -113,20 +120,9 @@
...
-                headers: { authorization: 'Bearer test-bearer-token' })
-
-        subject.put(message)
-      end
-    end
-
     context 'when response code is 201 - created' do
-      it 'returns a created status' do
-        expect(subject.put(message)).to eq(:success)
+      it 'returns the response bodyu' do
```
robinsk1, line 124: typo

app/jobs/submit_to_app_store.rb
```diff
@@ -14,8 +14,9 @@ def perform(submission:)
     # means waiting until the previous lock is released, at which point any
     # modifications made as part of that transaction will have been committed
     submission.with_lock do
-      submit(submission)
+      new_data = submit(submission)
       notify(submission)
+      update_records(new_data, submission)
       submission.update!(submit_to_app_store_completed: true)
```
robinsk1, line 20: if  new_data['application_state'] == submission.state is true do we still want `submission.update!(submit_to_app_store_completed: true)` to flag as `true` ?

patrick-laa: I think so, cos either way we have successfully submitted to the app store.

PR #1308, opened 2024-11-20
CRM457-2287 Provider: Youth Court Fee Form

imtiazAhmd:
## Description of change
Adds Youth court additional fee page
[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2287)

## Notes to reviewer
In the controller update method is there with commented out code, that should be uncommented after DecisionTree implementation.
`nocov:` has been added as cant test the controller without decision tree.

## Screenshot
Screenshot 2024-11-20 at 15 02 27


app/views/nsm/steps/youth_court_claim_additional_fee/edit.html.erb
```diff
@@ -0,0 +1,19 @@
+<% title t('.page_title') %>
+<% decision_step_header %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary(@form_object) %>
+    <h1 class="govuk-heading-l"><%= t('.heading') %></h1>
```
ivanELEC, line 7: nit: this is technically not the correct element to use, discovered this when doing my ticket after a discussion with Mary - the [GDS for radio buttons is formatted ](https://design-system.service.gov.uk/components/radios/) so that there is a h1 tag with class govuk-fieldset__heading within a legend tag. 

I don't think this should block deployment though as I think we have this problem in other places as well

app/views/nsm/steps/youth_court_claim_additional_fee/edit.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    <h2 class="govuk-hint govuk-!-margin-bottom-6"><%= t '.subheading' %></h2>
+
+    <%= step_form @form_object do |f| %>
+      <%= f.govuk_radio_buttons_fieldset :youth_court_fee_claimed, legend: nil do %>
+        <%= f.govuk_radio_button :youth_court_fee_claimed, YesNoAnswer::YES.to_s, label: { text: t('helpers.label.nsm_steps_youth_court_claim_additional_fee_form.options.yes' ) } %>
```
patrick-laa, line 12: If you have multiple radio buttons, you need to add `link_errors: true` to the first one so that if there's a validation error, the summary message will link to the form element.

You might be better off using `govuk_collection_radio_buttons` which will handle that sort of thing for you.

app/forms/nsm/steps/youth_court_claim_additional_fee_form.rb
```diff
@@ -0,0 +1,21 @@
+# this is a form to determine where to move to the equality questions
+# or skip them, as such it does not persist anything to DB
+module Nsm
+  module Steps
+    class YouthCourtClaimAdditionalFeeForm < ::Steps::BaseFormObject
+      attribute :youth_court_fee_claimed, :value_object, source: YesNoAnswer
+
+      validates :youth_court_fee_claimed, presence: true, inclusion: { in: YesNoAnswer.values }
```
patrick-laa, line 8: Don't forget to add error strings to the locales files

db/migrate/20241120130043_add_column_youth_court_fee_claimed_to_claims.rb
```diff
@@ -0,0 +1,5 @@
+class AddColumnYouthCourtFeeClaimedToClaims < ActiveRecord::Migration[8.0]
+  def change
+    add_column :claims, :youth_court_fee_claimed, :string
```
patrick-laa, line 3: So at the moment NSM has three ways of indicating boolean fields. Some are strings with "yes"/"no" values, some are strings with "true"/"false" values, and some are actual booleans.

They should really all be booleans.

I would recommend using an actual boolean here as it's the most appropriate data type, but equally there are so many "yes"/"no" strings that if you want to go with that for maximal consistency I think it's fair enough.

ivanELEC:
Needs error message for when no option selected filled in

app/controllers/nsm/steps/youth_court_claim_additional_fee_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+      end
+
+      private
+
+      def decision_tree_class
+        Decisions::DecisionTree
+      end
+
+      def additional_permitted_params
+        [:youth_court_fee_claimed]
```
patrick-laa, line 22: I'm surprised you need this - it should be inferred from being part of the `YouthCouldClaimAdditionalFeeForm.attributes`

db/migrate/20241120130043_add_column_youth_court_fee_claimed_to_claims.rb
ivanELEC: actually changed my mind - things aren't consistent enough that going one way or the other = more consistency so going with the more appropriate type

robinsk1 (approved):

PR #1309, opened 2024-11-20
CRM457-2283 - Provider: Case Category Form

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2283)

## TODO

- [x] Update common form gem for option translations
- [x] Use common form gem for case category translations
- [x] Feature flagging
- [x] Unit tests
- [x] See if snytax for option legend can be simplified

## Notes for reviewer
Not including controller for coverage for now as the behaviour of the controller is widely affected by decision tree logics which will be done in later ticket


app/views/nsm/steps/case_category/edit.html.erb
```diff
@@ -0,0 +1,20 @@
+<% title t('.page_title') %>
+<% decision_step_header %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary(@form_object) %>
+
+    <%= step_form @form_object do |f| %>
+      <%= f.govuk_radio_buttons_fieldset :plea_category, legend: -> { radio_legend(t('.heading')) } do %>
```
patrick-laa, line 9: I'm curious about this. In loads of places we're doing `legend: { text: t('.legend'), size: 'xl', tag: 'h1' }`. What's the problem with doing that, and can that problem be explained as a comment in the `radio_legend` method?

config/locales/en.yml
```diff
@@ -3,6 +3,12 @@ en:
   service:
     name: Submit a crime form
   shared:
+    plea_category:
```
patrick-laa, line 6: Since we now have these in the gem, can we avoid duplication and use the gem translations?

config/locales/en/nsm/steps.yml
```diff
@@ -40,6 +40,10 @@ en:
         edit:
           page_title: Case details
           heading: Case details
+      case_category:
+        edit:
+          page_title: Case category
```
patrick-laa, line 45: Nit: Is this page title in the requirements? In general I think we want our page titles to be the same as our h1s unless specified otherwise (can't find that spelled out in GDS, but in HMRC they make that link: https://design.tax.service.gov.uk/hmrc-design-patterns/page-title/)

app/services/decisions/decision_tree.rb
```diff
@@ -64,6 +64,8 @@ class DecisionTree < DslDecisionTree
     from(:case_details).goto(edit: 'nsm/steps/hearing_details')
     from(:hearing_details).goto(edit: 'nsm/steps/case_disposal')
     from(:case_disposal).goto(edit: 'nsm/steps/reason_for_claim')
+    # To be updated when we do ticket: CRM457-2288
```
patrick-laa, line 67: Nit: The syntax we have usually used for these is `# TODO CRM457-2288: [explanation of what needs to change]`

config/locales/en.yml
ivanELEC: yeah this is just placeholder - was waiting for my update to the common gem to be approved

app/views/nsm/steps/case_category/edit.html.erb
ivanELEC: tbh, I don't thnk there's a problem with this - I just didn't know the correct syntax and went down kind of a roundabout route that was based on an example in the documentation, will update this and remove the helper

it does seem like some of the other pages with radio button options might need updating though as they have this legend as the page heading instead (which doesn't follow GDS) - but that's a different story

config/locales/en.yml
elken: You can just update the `Gemfile` to point to a local path instead

```diff
@@ -13,7 +13,9 @@ gem 'govuk_notify_rails', '~> 3.0.0'
 gem 'grover'
 gem 'httparty'
 gem 'jsbundling-rails'
-gem 'laa_crime_forms_common', '~> 0.3.0', github: 'ministryofjustice/laa-crime-forms-common'
+# gem 'laa_crime_forms_common', '~> 0.3.0', github: 'ministryofjustice/laa-crime-forms-common'
+# TODO: Remove this before merge
+gem 'laa_crime_forms_common', path: '../laa-crime-forms-common'
 gem 'laa_multi_step_forms', path: './gems/laa_multi_step_forms'

```

config/routes.rb
```diff
@@ -87,6 +87,7 @@
         edit_step :contact_details
         edit_step :office_code
         edit_step :case_details
+        edit_step :case_category
```
elken, line 90: Nothing stopping someone from accessing this early, should be a constrained route.

spec/system/nsm/case_category_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  let(:claim) { create(:claim, claim_type:, youth_court:) }
+  let(:claim_type) { 'non_standard_magistrate' }
+  let(:youth_court) { 'yes' }
+
+  before do
+    visit provider_saml_omniauth_callback_path
+  end
+
+  it 'can do green path' do
+    visit edit_nsm_steps_case_category_path(claim.id)
```
elken, line 13: All tests go here so this can go into `before`

config/routes.rb
ivanELEC: agreed - but seen what the feature flag is now so will add the code in based on the fact that I've already seen what Robin is adding for this

elken (approved):
Minor change but otherwise looks good

PR #1314, opened 2024-11-25
Crm457 2290: Rename youth court field

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2290)

Needed so that we don't break the crime form common gem, [see conversation](https://mojdt.slack.com/archives/C05212WRK5J/p1732537649569579)

### DIFF
--- a/app/forms/nsm/steps/youth_court_claim_additional_fee_form.rb
+++ b/app/forms/nsm/steps/youth_court_claim_additional_fee_form.rb
@@ -3,9 +3,9 @@
 module Nsm
   module Steps
     class YouthCourtClaimAdditionalFeeForm < ::Steps::BaseFormObject
-      attribute :claimed_include_youth_court_fee, :boolean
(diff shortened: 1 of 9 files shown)


elken (approved):
Annoying timing as I'll have a bunch of merge conflicts from this 🤣

patrick-laa:
I'm not _entirely_ sure this is necessary - in the context of the gem you want to prefix with `claimed`, but that's because it's juxtaposed against the `assessed` version. But you don't really need it in the provider DB because the `claimed` bit is implied. In the same way disbursements have `miles`, not `claimed` miles. Remember, you have to manually feed the data into the calculation, you don't just shove the payload in, so that may be the point to make the naming transfer.

ivanELEC:
> Annoying timing as I'll have a bunch of merge conflicts from this 🤣

apologies haha

PR #1315, opened 2024-11-26
CRM457-2221: Load more from the app store

patrick-laa:
## Description of change
- Clear our the old batch-sync logic (note that this is the stuff that predates pub/sub. We're not getting rid of pub/sub yet, just getting 1 step closer to it)
- Load all submitted application lists by calling the app store search endpoint instead of checking the local DB
- Load submission details views from app store, not local DB
  - (To make this work with UI elements designed for db-backed models, create a bunch of presenters that mirror the basic API of activerecord objects)
- (Move a bunch of presenter specs so that they're not split equally between 2 folders named "presenter" and "presenters")
- Search now looks at the local DB for drafts only, the app store for everything else (and if there are results in both places it repaginates on the fly)
- Instead of relying on pub/sub to keep the local DB in sync, we do just-in-time syncing (as we only need syncing for RFI loops which will always start with a provider looking at the application details)

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2221)


app/services/app_store_list_service.rb
```diff
@@ -0,0 +1,29 @@
+module AppStoreListService
```
elken, line 1: To semi-future proof the whole "one single LAA data store" requirement, should this be named/structured as such?

elken (dismissed):
Makes sense to me

app/controllers/prior_authority/applications_controller.rb
```diff
@@ -97,7 +97,11 @@ def set_default_table_sort_options
     end
 
     def application_locals(skip_links: false)
-      application = PriorAuthorityApplication.for(current_provider).find(params[:id])
+      data = AppStoreClient.new.get(params[:id])
+      combined = data.merge(data.delete('application'))
```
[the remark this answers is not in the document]
patrick-laa: Ah, what you're missing is that `hash.delete(key)` doesn't return the hash without the key, it returns the value of the key). What we're doing is "flattening" the hash. So we start with:

`{ application_state: 'submitted', application: { foo: :bar, ... }, ... }`

We delete the `application` key and merge its _contents_ with the top level stuff, so we end up with:

`{ application_state: 'submitted', foo: :bar, .... }`

app/services/app_store_list_service.rb
patrick-laa: Yeah, I'm not sure about the best way to do that. My thinking is that as far as the provider app is concerned, there is a store of applications, that it can call "the app store", and it doesn't need to know or care whether "the" app store is a dedicated thing that's only used for CRM4 and 7, or whether it's an LAA-wide thing that also contains a bunch of other stuff. It's just the store, and it can be used to ask for specific lists of applications by this service that knows about the API of the relevant endpoint.

elken: Yeah that's true, I suppose there wouldn't be much to do and that's probably so far down the road it doesn't warrant worrying

elken:
@patrick-laa I approved this but is this PR just for that one case or are you going to implement everything under this PR?

patrick-laa:
> are you going to implement everything under this PR?

@elken I think I'll do it all as one gulp, as I'm nearly there now anyway. So do feel free to clear and I'll flag up when the whole thing is ready to go.

app/models/concerns/disbursement_details.rb
```diff
@@ -0,0 +1,25 @@
+module DisbursementDetails
+  extend ActiveSupport::Concern
+
+  included do
+    include DisbursementCosts
+  end
+
+  def translated_disbursement_type
+    if disbursement_type == DisbursementTypes::OTHER.to_s
+      # rubocop:disable Performance/InefficientHashSearch (False positive)
```
elken, line 10: Why is this a false positive?

app/views/nsm/steps/view_claim/_adjustments.html.erb
```diff
@@ -1,13 +1,13 @@
 <div class="govuk-tabs" data-module="govuk-tabs">
   <ul class="govuk-tabs__list">
     <li class="govuk-tabs__list-item <%= 'govuk-tabs__list-item--selected' if @scope == :work_items %>">
-      <%= link_to t(".#{prefix}work_items"), polymorphic_path([@section, :work_items, :nsm, :steps_view, claim], anchor: 'claim_nav') %>
+      <%= link_to t(".#{prefix}work_items"), send("#{@section}_work_items_nsm_steps_view_claim_path", claim, anchor: 'claim_nav') %>
```
elken, line 4: What is `send` here?

elken (approved):

app/models/concerns/disbursement_details.rb
patrick-laa: Rubocop sees `.values` and wrongly concluses that OtherDisbursementTypes is a hash. If it was then `values.include?` would be less efficient than an alternative method of interrogating the values of a hash.

I've made the comment more explicit.

elken: Ah and because it's a static analysis tool, it can't know if it's a hash. Makes sense.

app/views/nsm/steps/view_claim/_adjustments.html.erb
patrick-laa: `send` is how you call a method on a ruby object if all you have is the name's method as a string. `a.send("foo")` is broadly the same as `a.foo`. In this case we can no longer use `polymorphic_path` because that uses some magic that hinges on taking the `claim` and, because it's a `Claim` object, inferring that the required path helper method has `_claim_` in its name. But Now `claim` is a `AppStore::V1::Nsm::Claim`, that same magic tries to find a path helper method with `_app_store_v1_nsm_claim_` in its name, and there obviously is no such path helper. So we have to do a different sort of metaprogramming, in this case constructing the helper method name ourselves as a string and invoking it with `send`

robinsk1 (approved):

PR #1319, opened 2024-11-27
CRM457-2289: Youth Court Fee Payment Pages/CYA

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2289)

### After changes:
 
image


ivanELEC:
FYI: will have merge conflicts with @elken ticket for the tasklist, probably better to merge that in before merging this in

app/models/claim.rb
```diff
@@ -120,6 +124,10 @@ def can_claim_youth_court?
       plea_category.match(/category_[12]a/)
   end
 
+  def additional_fees_applicable?
+    [can_claim_youth_court?].excluding(false).present?
```
elken, line 128: `[can_claim_youth_court?].any?`

app/presenters/nsm/cost_summary/base.rb
```diff
@@ -30,6 +30,11 @@ def footer_row
           { text: total_cost_cell, classes: 'govuk-table__cell--numeric govuk-summary-list__value-bold' }
         ].compact
       end
+
+      # override if route key for change url does not match key of summary
+      def change_key
```
elken, line 35: Rather than doing this you can just call the function like `.try(:change_key)` without a dead base member

app/presenters/nsm/cost_summary/report.rb
```diff
@@ -22,7 +23,7 @@ def sections
           {
             card: {
               title: data.title,
-              actions: actions(name)
+              actions: actions(data.change_key || name)
```
elken, line 26: See above comment for how to update this

app/models/claim.rb
ivanELEC: yeah the intention was exactly for what @elken suggested - to future proof for future examples of "additional fee", the only reason I wrote it this way was so it would be easy to read if we had loads of different fees, but maybe it's not easier to read after all aha. Maybe for now I'll do away with the method and use can_claim_youth_court? for now and we can update once we get to that point to prevent redundant code

app/presenters/nsm/cost_summary/base.rb
ivanELEC: true - I wonder if having the dead base member makes it easier to understand the intention of the method though? or maybe that indicates the method name needs changing

app/models/claim.rb
ivanELEC: actually thinking about it, I think it's still worth having the additional_fees_applicable? method so that we don't have to overhaul loadsa tests when we get to the point where we'd need to add another additional fee

elken: Just change `match` to `match?` if you want a bool, but `nil` is still falsey so I don't know if there's a good argument to doing so

app/presenters/nsm/cost_summary/base.rb
elken: I don't know if it does make it easier to understand, seems like it'll rarely need to be overriden (since it hasn't needed to be before now) but it's not a major deal so up to you

app/models/claim.rb
```diff
@@ -121,11 +121,11 @@ def can_claim_youth_court?
     claim_type == ClaimType::NON_STANDARD_MAGISTRATE.to_s &&
       !before_youth_court_cutoff? &&
       youth_court == 'yes' &&
-      plea_category.match(/category_[12]a/)
+      plea_category.match(/category_[12]a/).present?
```
elken, line 124: `.match?`

app/presenters/nsm/cost_summary/base.rb
ivanELEC: yeah I think you're right - the intention is already in the code and there's an example of it with a test associated to it

app/models/claim.rb
ivanELEC: yeah you're right it's unnecessary - updated!

app/presenters/nsm/cost_summary/report.rb
ivanELEC: updated

app/models/claim.rb
ivanELEC: I just removed it, you're right - it's not really necessary to do

ivanELEC: (as in removed .present?)

elken (approved):

PR #1322, opened 2024-11-29
CRM457-2221 part 2: dual-source search

patrick-laa:
## Description of change
If we can run provider search against the app store where possible it means we don't have to worry about 'syncing' our local DB as soon as changes happen in the app store. So this introduces a mechanism to enable search against the local DB for drafts and the app store for submissions, and do on-the-fly sorting/pagination if we need to combine results from both sources

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2221)


app/services/search/service.rb
```diff
@@ -0,0 +1,168 @@
...
+      local_unpaginated_results.order(local_order)
+                               .offset(offset)
+                               .limit(pagination_params[:per_page])
+                               .map { Search::Result.new(_1) }
+    end
+
+    def order(combined_results)
+      sorted = combined_results.sort_by do |result|
+        raw = result.public_send(unified_sort_by)
+        raw.respond_to?(:downcase) ? raw.downcase : raw
```
elken, line 107: `raw.try(:downcase) || raw` or even `raw&.downcase || raw` maybe

elken (approved):
How does this handle the so-rare-it-probably-cant-happen case of there being an application in both stores?

I suppose because we've eliminated all sources of async, this "shouldn't" be able to happen; but bugs do occur...

app/services/search/service.rb
patrick-laa: Good point, I had totally forgotten `try` existed. FWIW `&.` won't save me here, because it will blow up if `raw` exists but does not have a `downcase` method. But `try` is exactly the right tool for the job.

patrick-laa:
> How does this handle the so-rare-it-probably-cant-happen case of there being an application in both stores?

Good question. The local search explicitly searches for only things in a draft state. The app store rules don't allow it to hold a submission in a draft state. But nevertheless, _if_ a submission was marked as draft in the local DB but somehow also existed in the app store in a different state, then it would appear as 2 results, with different statuses.

I _really_ hope we never break things badly enough that we get into that state...

elken:
> > How does this handle the so-rare-it-probably-cant-happen case of there being an application in both stores?
> 
> Good question. The local search explicitly searches for only things in a draft state. The app store rules don't allow it to hold a submission in a draft state. But nevertheless, _if_ a submission was marked as draft in the local DB but somehow also existed in the app store in a different state, then it would appear as 2 results, with different statuses.
> 
> I _really_ hope we never break things badly enough that we get into that state...

I actually did this morning, due to a couple of unrelated issues; which is what made me think of it 🤣 I had my local draft undeleted and a submitted copy in the app store.

It's good that it shows up twice then, because it means providers can flag to us and we can resolve it "in theory". Probably not worth spending MUCH time pre-empting _such_ an edge case.

ivanELEC (approved):

PR #1326, opened 2024-12-02
CRM457-2203: XML import

patrick-laa:
## Description of change
Add a new form to upload an XML file and that, if the file contains disbursements and work items in the expected format, will create a new claim and import details from the file.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2203)

### DIFF
--- a/app/services/nsm/import_service.rb
+++ b/app/services/nsm/import_service.rb
@@ -1,0 +1,101 @@
+module Nsm
+  # Currently this service only imports disbursements and work items, as it is the entering of these that is the most time
+  # consuming. However, there is more information available in the XML file that could be imported to save providers time.
+  class ImportService
+    def self.call(claim, import_form)
+      new(claim, import_form).call
+    end
+
+    attr_reader :claim, :import_form
+
+    def initialize(claim, import_form)
+      @claim = claim
+      @import_form = import_form
+    end
+
+    def call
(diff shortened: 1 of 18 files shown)


app/services/nsm/import_service.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    attr_reader :claim, :import_form
+
+    def initialize(claim, import_form)
+      @claim = claim
+      @import_form = import_form
+    end
+
+    def call
+      xml = Nokogiri::XML(File.read(@import_form.file_upload.tempfile))
```
elken, line 17: Will this validate it against the included schema?

elken (commented):
Once the q is answered, this is fine to approve

app/services/nsm/import_service.rb
patrick-laa: No, because while the sample XML file we got given contains a schema, the real-life XML files we've been given don't, and I wouldn't trust them to adhere to it. So instead we're trying to be as permissible as possible - if it contains some information that we can read, we read that information (and then prompt the user to check it)

elken: Always hesitant to handle XML we can't fully validate, but if there's no alternative I guess we have little choice.

elken (approved):

ivanELEC:
Pending a review of the visual side of things, LGTM!

PR #1331, opened 2024-12-03
CRM457-2196: Replace navigation_stack with viewed_steps

patrick-laa:
## Description of change
Stop using the old `navigation_stack` array which:
* used URL paths (which would break everything if we ever to change our URL structures)
* assumed (often wrongly) that its contents were in the same order as the steps needed to be completed in

Instead, use the new `viewed_steps` array, which just contains the names of the steps. Also, work out the order of steps from the DecisionTree (via `Decisions::OrderedSteps`) so that we don't need to make assumptions about the order of steps in `viewed_steps`

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2196)


elken (commented):
Do we want to handle completion/progress similarly too or are we happy with how we track that?

Will this say fix the long-standing issue of changing an older step and having everything go back to "Not started"?

patrick-laa:
> Do we want to handle completion/progress similarly too or are we happy with how we track that?
So for the most part I feel like task completion logic is quite sane - the relevant task says "is the associated form valid?" unless a specific task overwrites it with different logic, and likewise the in progress logic now just checks to see whether the step name is included in `viewed_steps`. So I'm cautiously hopeful that we don't need to touch those.

> Will this say fix the long-standing issue of changing an older step and having everything go back to "Not started"?
So there's 2 things. If you change an older step, everything after the immediate next step will go back to "Cannot start yet", and that's the system working as designed. Separately, there was a problem where instead of "Cannot start yet" it said "Not started", but we fixed that a while back.

elken:
> that's the system working as designed.

Oh, that seems really awkward but if that's what they want fair enough.

And I agree generally with the completion bits, so happy to approve this. It looks sane and decently covered test-wise

elken (approved):

elken:
Is there an ADR for this? If not, is it worth writing one?

ivanELEC:
> > that's the system working as designed.
> 
> Oh, that seems really awkward but if that's what they want fair enough.
> 
> And I agree generally with the completion bits, so happy to approve this. It looks sane and decently covered test-wise

yeah awkward for the user but more managable for us (both from a tech standpoint and a BA standpoint)

ivanELEC (approved):

PR #1346, opened 2024-12-10
CRM457-2204: Debounce the evidence delete button

elken:
## Description of change
The XHR call should only trigger when the button hasn't been clicked
for a second. Any clicks in that window clear the timeout function and
starts the timer again.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2204)


app/javascript/multi-file-upload.js
```diff
@@ -97,36 +98,38 @@ MOJFrontend.MultiFileUpload.prototype.getFileRowHtml = function (file, fileListL
...
-        data: { evidence_id: button.attr('value') },
-        success: function (response) {
-            feedback.html(this.getSuccessHtml(`${fileName} has been deleted`));
-            button.parents('.moj-multi-file-upload__row').remove();
-            if (this.feedbackContainer.find('.moj-multi-file-upload__row').length === 0) {
-                this.feedbackContainer.addClass('moj-hidden');
-            }
-            this.params.fileDeleteHook(this, response);
-        }.bind(this),
+MOJFrontend.MultiFileUpload.prototype.onFileDeleteClick = debounce(function(e) {
```
patrick-laa, line 101: I don't think we want to debounce here. If for some reason it takes 5, 10, 100 seconds for a 'delete' click to process we still don't want the click to be possible. And debouncing will ignore a click, which isn't a UX we want. The UX we want is: When you click the delete button, the delete button goes away immediately, so the user has no opportunity to click again. (*Possibly* we want to flash up the loading indicator until the success/error callback triggers, but I suspect that will look awful 99% of the time because it will flicker so briefly.)

elken: I disagree there, it's an established pattern to debounce click events that mutate backend state. 1s is probably too long, this can probably be half a second.

By your same logic, if it takes 5, 10, 100 seconds to process the click event then we still have the same issue with trying to hide the button.

patrick-laa: Re your first point, it's against GDS standards to have a button that does nothing when you click it, which is what your debounce does.

Re your second point, I'm saying instead of hiding the button after the time-consuming HTTP request/response, we hide it immediately - at the point where it's no longer allowable for the user to click the button. This makes the problem go away immediately.

patrick-laa: Re switching to half a second, that makes your solution worse - because it still doesn't prevent us sending am error-throwing double delete request to the server if the round trip from the original click takes more than half a second. We need a comprehensive solution that totally prevents triggering deletes if an existing delete request is still being processed

elken: Ah that's annoying, okay I'll adjust.

Yeah I get what you're saying, the debounce starts straight away. But fine, I'll have to change it anyway.

ivanELEC (approved):

PR #1352, opened 2024-12-11
CRM457-2336: Add in missing AC from QA

elken:
## Description of change

- If an additional fee was adjusted, show the comment in the table
- If there were adjustments but not to additional fees, show the tab anyway

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2336)

### DIFF
--- a/app/views/nsm/steps/view_claim/_adjustments.html.erb
+++ b/app/views/nsm/steps/view_claim/_adjustments.html.erb
@@ -6,7 +6,7 @@
     <li class="govuk-tabs__list-item <%= 'govuk-tabs__list-item--selected' if @scope == :letters_and_calls %>">
       <%= link_to t(".#{prefix}letters_and_calls"), send("#{@section}_letters_and_calls_nsm_steps_view_claim_path", claim, anchor: 'claim_nav') %>
     </li>
-    <% if @section == :adjustments && claim.youth_court_fee_adjustment_comment? || @section != :adjustments && claim.can_claim_youth_court? %>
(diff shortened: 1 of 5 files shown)


app/views/nsm/steps/view_claim/_adjustments.html.erb
```diff
@@ -6,11 +6,9 @@
     <li class="govuk-tabs__list-item <%= 'govuk-tabs__list-item--selected' if @scope == :letters_and_calls %>">
       <%= link_to t(".#{prefix}letters_and_calls"), send("#{@section}_letters_and_calls_nsm_steps_view_claim_path", claim, anchor: 'claim_nav') %>
     </li>
-    <% if @section == :adjustments && claim.youth_court_fee_adjustment_comment? || @section != :adjustments && claim.can_claim_youth_court? %>
```
patrick-laa, line 9: My guess (but check with Mary if you're not convinced) is that we don't want to remove this if block entirely. As I understand it, if this was a claim that had nothing to do with youth court fees (e.g. BOI) then we wouldn't want to show this tab at all.

elken: That's a good point yes, I think you might be right but I'll ask

elken: Confirmed to be wrong, adjusted

ivanELEC (approved):

ivanELEC (approved):

robinsk1 (approved):

PR #1354, opened 2024-12-12
CRM457-2340: Stop looking in local DB unnecessarily

patrick-laa:
## Description of change
Only refer to local DB post-submission when we really need to. This means:
- For claims, we now don't load the local value at all if we're just viewing a record
- During an RFI loop we only use the local DB to store the new information provided by a provider until we're ready to push it back to the app store
- We can even handle a claim RFI loop for a claim that doesn't exist at all in the local DB (although only by creating a stub record so that the local FI record foreign key doesn't break, which isn't ideal but probably OK given it won't ever happen in production)
- PriorAuthorityApplications are a bit more complicated as to do an RFI loop, even if it's just for further info, we still have to go via the check answers screen and that gets a bit hairy. And obviously for "corrections needed" RFIs we have to keep using the local DB. So I've left PA RFIs looking in the local DB for now, but just viewing a submitted record is already entirely avoiding the local DB.
- As a bonus, I've removed about 30 identical 1-line `decision_tree_class` methods from controllers that all inherit from the same base class

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2340)


app/controllers/nsm/steps/claim_confirmation_controller.rb
```diff
@@ -1,13 +1,19 @@
...
+
       def show
         @laa_reference = current_application.laa_reference
       end
 
       def step_valid?
         current_application.submitted? || current_application.provider_updated?
       end
+
+      def current_application
```
elken, line 14: Should this not be part of the base controller?

elken (approved):

app/controllers/nsm/steps/claim_confirmation_controller.rb
patrick-laa: This is actually overriding the base controller behaviour, which by default loads `current_application` from the local DB. Most of the many child controllers use the default behaviour. I didn't think the exceptions were quite frequent enough to make it worth complicating the inheritance tree or adding a concern just to handle an overridden one-liner.

PR #1358, opened 2024-12-16
CRM457-2188: Remove previous mentions to eforms

elken:
## Description of change

Remove all textual and error mentions to eforms

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2188)

### DIFF
--- a/app/views/home/index.html.erb
+++ b/app/views/home/index.html.erb
@@ -8,9 +8,6 @@
 
     <%= render partial: 'shared/notifications' %>
 
-    <p>
-      <%= t('.introduction_html', eforms_link: link_to(t('.eforms'), "https://portal.legalservices.gov.uk/")) %>
-    </p>
     <h2 class="govuk-heading-s">
       <%= link_to t('.pa_header'), prior_authority_applications_path, class: "govuk-link--no-visited-state" %>
     </h2>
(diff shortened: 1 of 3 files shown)


config/locales/en/errors.yml
```diff
@@ -47,14 +47,6 @@ en:
         heading: Access restricted
         lead_text: This page requires you to be signed in to LAA Portal. You will be redirected back to this service after you sign in.
         sign_in_button: Sign in with LAA Portal
-      not_enrolled:
```
jogilder, line 50: I don't think you should take all of this out - there is still the issue of someone giving them the link and they don't have the correct role from portal.

gems/laa_multi_step_forms/app/controllers/laa_multi_step_forms/errors_controller.rb
```diff
@@ -19,10 +19,6 @@ def unauthorized
       respond_with_status(:unauthorized)
     end
 
-    def not_enrolled
```
jogilder, line 22: this should stay

jogilder (commented):
not enrolled is still required I believe

config/locales/en/errors.yml
elken: Oh I see, I thought that was for something else. Will amend

robinsk1 (approved):

PR #1360, opened 2024-12-17
Remove the gem

patrick-laa:
## Description of change
Stop having half our logic in a gem-that's-basically-just-a-local-folder.
I've simplified a couple of the most egregiously complex inheritance trees, merging Tasks::Generic with Tasks::BaseTask and Decisions::DslDecisionTree with Decisions::BaseDecisionTree.


README.md
```diff
@@ -88,17 +88,23 @@ kubectl get secret sidekiq-auth -o jsonpath='{.data}' --namespace=$NAMESPACE | j
...
-* [Suggestion](gems/laa_multi_step_forms/docs/Suggestion.md)
-* [TimePeriodAndFormBuilderUpgrade](gems/laa_multi_step_forms/docs/TimePeriodAndFormBuilderUpgrade.md)
+This codebase was originally built by taking a bunch of code from the Crime Apply codebase and extracting it into a gem
+called LaaMultiStepForms, which sat in a sub-folder of this repo. The hope was apparently that Crime Apply would start using
+this gem as well, so both codebases would be kept aligned. But they never did (having no incentive to do so, and no assurances
+about the gem being maintained to suit their needs), and the gem became more and more specific to this repo's needs, to the point
+where having some of our logic in a pseudo-gem just added unnecessary complexity. So we've now moved the code back into the main
+body of the repo, but a lot of logic is still broken up into the semi-generic base logic that used to be in the gem, and then
+some more specific subclasses with NSCC-specific details. The original documentation about the gem base classes is available here:
+
```
elken, line 98: Maybe worth mentioning about the now-historical multi-step gem or no?

config/routes.rb
```diff
@@ -221,6 +218,15 @@
 
   post :app_store_webhook, to: 'sync#sync_individual'
 
+  resource :errors, only: [] do
```
elken, line 221: There already an errors namespace, should these not go there?

elken (approved):
Big fan of removing loads of no-longer needed code (or is this case mostly moving) but probably worth taking a bit longer in QA to make sure everything that it touches has been handled correctly

config/routes.rb
patrick-laa: It's a bit tricky to merge a namespace with a resource - basically we have an errors module (the namespace) and an errors class (the resource). It'd be a worthwhile refactor to unify these at some point.

PR #1364, opened 2024-12-17
CRM457-2177: Switch form field type to email

patrick-laa:
## Description of change
To help browsers autocomplete these fields

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2177)


elken (approved):
Assume these are the only missed ones, generally we're pretty good with matching the field types

PR #1371, opened 2024-12-18
CRM457-2349: Improve form un-disabling

patrick-laa:
## Description of change
- Chrome sometimes uses a 'navigate' performance entry navigation type and sometimes a 'reload' one in the situation we want, so changed the code to handle both
- Rather than brute-force re-enabling a form via a page refresh, undo the disabling by removing the dummy element and removing the 'disabled' attribute

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2349)


elken (commented):
This looks fine, but I want to test it on Firefox first before I approve. Will do once I have.

elken (approved):

PR #1372, opened 2024-12-18
CRM457-2338: Add functionality to clone claims & applications

elken:
## Description of change

Allows draft claims and draft applications to be cloned when not in
production (since it's mostly for testing) and also resolve some
incorrect `dependent` calls for `belongs_to` models that shouldn't
have it.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2338)

### DIFF
--- a/app/controllers/concerns/claim_creatable.rb
+++ b/app/controllers/concerns/claim_creatable.rb
@@ -1,5 +1,6 @@
 module ClaimCreatable
   extend ActiveSupport::Concern
+  include GeneratesLaaReference
 
   def initialize_application(&block)
     attributes = {
@@ -8,12 +9,5 @@ module ClaimCreatable
       laa_reference: generate_laa_reference
     }
     Claim.create!(attributes).tap(&block)
-  end
-
-  def generate_laa_reference
-    loop do
-      random_reference = "LAA-#{SecureRandom.alphanumeric(6)}"
-      break random_reference unless Claim.exists?(laa_reference: random_reference)
-    end
   end
 end
--- a/app/controllers/nsm/claims_controller.rb
+++ b/app/controllers/nsm/claims_controller.rb
@@ -2,6 +2,7 @@ module Nsm
   class ClaimsController < ApplicationController
     include Searchable
     include ClaimCreatable
+    include Cloneable
     layout 'nsm'
 
     before_action :set_scope, only: %i[submitted draft]
@@ -41,6 +42,17 @@ module Nsm
       @model = Claim.for(current_provider).find(params[:id])
       @model.destroy
       redirect_to draft_nsm_applications_path, flash: { success: t('.deleted') }
+    end
+
+    def clone
+      clone = clone_application
+      clone.save
+
+      redirect_to nsm_steps_start_page_path(clone.id), flash: { success: t('.cloned') }
+    end
+
+    def self.model_class
+      Claim
     end
 
     private
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -74,6 +74,15 @@ class Claim < ApplicationRecord
     defendants.where(main: false)
   end
 
+  def dup
+    super.tap do |new_record|
+      new_record.defendants = defendants.map(&:dup)
+      new_record.work_items = work_items.map(&:dup)
+      new_record.disbursements = disbursements.map(&:dup)
+      new_record.supporting_evidence = supporting_evidence.map(&:dup)
+    end
+  end
+
   private
 
   def sorted_work_item_ids
--- a/app/models/firm_office.rb
+++ b/app/models/firm_office.rb
@@ -1,2 +1,3 @@
 class FirmOffice < ApplicationRecord
+  has_many :prior_authority_applications, dependent: :destroy
 end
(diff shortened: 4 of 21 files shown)


app/controllers/concerns/claim_creatable.rb
```diff
@@ -1,5 +1,6 @@
 module ClaimCreatable
   extend ActiveSupport::Concern
+  include UsesLaaReference
```
elken, line 3: Terrible name, open to suggestions

app/controllers/nsm/claims_controller.rb
```diff
@@ -43,6 +44,12 @@ def destroy
       redirect_to draft_nsm_applications_path, flash: { success: t('.deleted') }
     end
 
+    def clone
+      clone_application.save
+
+      redirect_to draft_nsm_applications_path, flash: { success: t('.cloned') }
```
patrick-laa, line 50: If you have hundreds of drafts it might be a kindness to identify which is the new clone - perhaps taking us to the start page for that claim? Or putting its LAA reference in the flash?

spec/factories/claims.rb
```diff
@@ -39,6 +39,12 @@
       has_disbursements { 'no' }
     end
 
+    Claim.states.each_key do |state|
```
patrick-laa, line 42: This is a really neat little addition to the factory.

app/controllers/concerns/uses_laa_reference.rb
```diff
@@ -0,0 +1,8 @@
+module UsesLaaReference
+  def generate_laa_reference
+    loop do
+      random_reference = "LAA-#{SecureRandom.alphanumeric(6)}"
+      break random_reference unless Claim.exists?(laa_reference: random_reference)
```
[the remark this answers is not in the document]
elken: Ah you're right, didn't spot that. Should just rely on `current_application`

app/models/claim.rb
```diff
@@ -2,8 +2,8 @@ class Claim < ApplicationRecord
   include ClaimDetails
 
   belongs_to :submitter, class_name: 'Provider', optional: true
-  belongs_to :firm_office, optional: true, dependent: :destroy
```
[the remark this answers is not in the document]
elken: I don't know why the whole line went, I only intended to remove `dependent: :destroy`

app/views/nsm/claims/index.html.erb
```diff
@@ -139,6 +139,16 @@
               <td class="govuk-table__cell"><%= claim.laa_reference %></td>
               <% if @scope == :draft %>
                 <td class="govuk-table__cell">
+                  <% if !Rails.env.production? %>
```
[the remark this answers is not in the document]
elken: Yeah there's a whole nest of checks for this & I wasn't sure. Basically we want it on everywhere except actual live system

spec/factories/claims.rb
elken: Also reminds me to do the same for PA

app/controllers/nsm/claims_controller.rb
elken: I was leaning towards the former at first but then that felt like guessing what you wanted to do. I guess most of the time you'll want to clone then submit, so going to the start page probably makes the most sense.

app/models/claim.rb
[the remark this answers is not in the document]
patrick-laa: The line didn't go - what I'm saying is I don't think we want to remove `dependent: :destroy` - it's there for a reason.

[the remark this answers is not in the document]
patrick-laa: The docs say:

> You should not specify this option on a belongs_to association that is connected with a has_many association on the other class.

I'm saying the other class shouldn't have a has_many association, it should have a has_one, so the dependent: :destroy remains valid and useful.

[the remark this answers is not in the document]
elken: Ah okay then yes you are right

app/controllers/concerns/claim_creatable.rb
ivanELEC: got nothing better honestly

app/models/firm_office.rb
```diff
@@ -1,2 +1,4 @@
 class FirmOffice < ApplicationRecord
+  has_one :prior_authority_applications, dependent: :destroy
```
robinsk1, line 2: wouldn't that be a `has_many`? - using the plural seems odd, something else going on here?

elken: See above comments from Patrick https://github.com/ministryofjustice/laa-submit-crime-forms/pull/1372#discussion_r1890405532

robinsk1: i'll go with the Patrick ESP but surely a `firm_office` does have multiple claims though? and if we wanted to redact/anonymise we could be be scoping claims within that has_many

elken: I think the model might be slightly incorrectly named but he is right. It's the output of what you put into the "Firm Details" step and while yes in theory that should be wrong that is how the system works.

robinsk1 (approved):

PR #1374, opened 2024-12-18
CRM457-2307: Add net, vat, total to disbursement cost summary

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2307)

Screenshot 2024-12-18 at 15 09 50

### DIFF
--- a/app/presenters/nsm/cost_summary/disbursements.rb
+++ b/app/presenters/nsm/cost_summary/disbursements.rb
@@ -53,24 +53,27 @@ module Nsm
       def footer_row
         [
           { text: translate('.footer.total'), classes: 'govuk-table__header' },
-          { text: total_cost_cell, classes: 'govuk-table__cell--numeric govuk-summary-list__value-bold' },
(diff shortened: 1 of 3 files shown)


app/presenters/nsm/cost_summary/disbursements.rb
```diff
@@ -15,6 +14,8 @@ def rows
           [
             { text: translated_text(disbursement), classes: 'govuk-table__header' },
             { text: NumberTo.pounds(disbursement.total_cost_pre_vat || 0), classes: 'govuk-table__cell--numeric' },
+            { text: NumberTo.pounds(disbursement.vat || 0), classes: 'govuk-table__cell--numeric' },
```
elken, line 17: Should we only show the VAT column if VAT is applicable to the claim?

patrick-laa: Ah, weird quirk here: disbursements can have VAT even if the firm is not VAT registered (unlike work items, letters and calls and fees etc). I think it's because what matters here is whether the provider of the service charges VAT).  So even if the claim is non-vat-applicable in general, VAT can still be a thing here.

PR #1377, opened 2024-12-19
Crm457-2353: Clean up Rubocop setup

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2353)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -104,9 +104,6 @@ Metrics/ClassLength:
 # To be tweaked until we find the right balance
 Metrics/MethodLength:
   Max: 15
-  Exclude:
-    - app/presenters/nsm/check_answers/**/*
-    - spec/system/support/prior_authority/step_helpers.rb
 
 # To be tweaked until we find the right balance
 Metrics/AbcSize:
@@ -123,9 +120,6 @@ Metrics/ParameterLists:
 
 Layout/LineLength:
   Max: 130
-  Exclude:
-    - spec/decisions/**/*
-    - spec/presenters/nsm/check_answers/application_status_card_spec.rb
 
 # https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/HashSyntax
 Style/HashSyntax:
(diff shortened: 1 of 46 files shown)


.rubocop.yml
```diff
@@ -12,7 +10,11 @@ AllCops:
   Exclude:
     - 'db/**/*'
     - 'bin/**/*'
-    - 'config/**/*'
+    - 'config/clamd/*'
```
elken, line 13: Probably simpler to just do

```yaml
Include:
  - 'config/**/*.rb'
```

app/forms/nsm/steps/case_disposal_form.rb
```diff
@@ -5,6 +5,7 @@ class CaseDisposalForm < ::Steps::BaseFormObject
       attribute :plea_category
       validates :plea, presence: true, inclusion: { in: PleaOptions.values }
 
+      # rubocop:disable Style/HashEachMethods
```
elken, line 8: Not going to put this for every single one, but if we can instead fix the issue that would be preferable

spec/decisions/back_decision_tree_spec.rb
```diff
@@ -9,72 +9,103 @@
   let(:work_items) { [] }
   let(:disbursements) { [] }
 
-  it_behaves_like 'a generic decision', from: 'nsm/steps/firm_details', goto: { action: :show, controller: 'nsm/steps/start_page' }
-  it_behaves_like 'a generic decision', from: 'nsm/steps/contact_details', goto: { action: :edit, controller: 'nsm/steps/firm_details' }
+  it_behaves_like 'a generic decision', from: 'nsm/steps/firm_details',
+goto: { action: :show, controller: 'nsm/steps/start_page' }
```
robinsk1, line 13: indentation looks strange, but might just be new rubocop rules strange

robinsk1 (approved):

spec/decisions/back_decision_tree_spec.rb
elken: Thought the same but it's a macro so Rubocop can't easily guess the indentation. Could add more line breaks to help it

.rubocop.yml
ivanELEC: That would still include some non .rb files. Decided to use a more catch all approach instead as I've seen you can do regex in this, so excluding anything that doesn't end in .rb

elken: Nope it'll just glob all the ruby files

image

app/forms/nsm/steps/case_disposal_form.rb
ivanELEC: hmm, not sure this is something that should be addressed and if it should I feel like it would need to be done as a separate ticket to think about how we deal with value objects. The ValueObject is not a hash and therefore we don't have access to the each_value method - so not sure why it's calling out that we are not using hash methods.

elken: As I say it was a general remark that if we can resolve the issues we should, but feel free to write another ticket for it.

.rubocop.yml
ivanELEC: there's non .rb files at config/* also wouldn't those get included?

elken: No, the current glob does but the above glob only finds `.rb` files. See above screenshot 😄

spec/decisions/back_decision_tree_spec.rb
ivanELEC: yeah that indeed looks odd, I edited it to make the indentation more readable and rubocop didn't say anything about it though

.rubocop.yml
ivanELEC: ah fair enough - I must have been reading the syntax wrong. Still thinking the regex might be better though since this means whatever we put in, wherever it is, if it's a ruby file - we rubocop it

app/forms/nsm/steps/case_disposal_form.rb
ivanELEC: Yeah I see your point - I generally went down the route of: if there is value in conforming to the linting rule, fix it, if not - leave the ignore in. In this case it was that it seems to be applying a hash related style cop to a non hash object (though maybe there's something I'm not seeing there). Other examples is Rspec/ExampleLength for tests on things like tables as those naturally will just be long (couldn't think of a way to shorten them that didn't require unnecessary complexity or unnecessary splitting up of tests and therefore adding to test execution time)

Will have another pass, but not looking to add another ticket if we can't find tangible value from it

PR #1379, opened 2024-12-23
CRM457-2262: Use v1 endpoint for provider user office codes

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2262)

## Notes for reviewer
I want to do a test run in UAT before enabling this in production, so production feature flag is off for now 

[Old endpoint
](https://laa-provider-details-api-prod.apps.live.cloud-platform.service.justice.gov.uk/swagger-ui/index.html#/Provider%20Users%20Endpoints/getProviderUsersOffices)
[New endpoint](https://laa-provider-details-api-prod.apps.live.cloud-platform.service.justice.gov.uk/swagger-ui/index.html?urls.primaryName=V1#/Provider%20Users%20Endpoints/getProviderUsersOfficesV1)

### DIFF
--- a/app/services/office_code_service.rb
+++ b/app/services/office_code_service.rb
@@ -1,8 +1,8 @@
 class OfficeCodeService
   class << self
-    def call(username, api_version = nil)
-      office_details = ProviderDataApiClient.user_office_details(username, api_version)
-      if api_version == 1
+    def call(username)
+      office_details = ProviderDataApiClient.user_office_details(username)
+      if FeatureFlags.provider_api_v1.enabled?
         office_details.map { _1['officeCodes'].pluck('firmOfficeCode') }.flatten
       else
         office_details.pluck('firmOfficeCode')
--- a/app/services/office_code_service.rb
+++ b/app/services/office_code_service.rb
@@ -1,8 +1,12 @@
 class OfficeCodeService
   class << self
-    def call(username)
-      office_details = ProviderDataApiClient.user_office_details(username)
-      office_details.pluck('firmOfficeCode')
+    def call(username, api_version = nil)
+      office_details = ProviderDataApiClient.user_office_details(username, api_version)
+      if api_version == 1
+        office_details.map { _1['officeCodes'].pluck('firmOfficeCode') }.flatten
+      else
+        office_details.pluck('firmOfficeCode')
+      end
     end
   end
 end
--- a/app/services/provider_data_api_client.rb
+++ b/app/services/provider_data_api_client.rb
@@ -8,8 +8,8 @@ class ProviderDataApiClient
       )
     end
 
-    def user_office_details(user_login, api_version = nil)
-      if api_version == 1
+    def user_office_details(user_login)
+      if FeatureFlags.provider_api_v1.enabled?
         query(
           "api/v1/provider-users/#{ERB::Util.url_encode(user_login)}/provider-offices",
           200 => ->(data) { data['offices'] },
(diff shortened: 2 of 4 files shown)


app/services/provider_data_api_client.rb
```diff
@@ -8,12 +8,20 @@ def contract_active?(office_code)
       )
     end
 
-    def user_office_details(user_login)
-      query(
-        "provider-users/#{ERB::Util.url_encode(user_login)}/provider-offices",
-        200 => ->(data) { data['officeCodes'] },
-        204 => [],
-      )
+    def user_office_details(user_login, api_version = nil)
```
elken, line 11: We're only ever going to want the new endpoint now

ivanELEC: true, but I think we should have versioning at least until we have confirmed the new version works properly and is stable. That said, I agree there's maybe not much value in retaining the logic for every api version ever though (e.g. when it moves to v2, v3, v4 etc it's very very unlikely we would ever have a scenario where we revert down by 2 versions), so will remove this argument in favour of just having the feature flag

app/services/office_code_service.rb
```diff
@@ -1,8 +1,12 @@
 class OfficeCodeService
```
robinsk1, line 1: would a version namespacing approach avoid the if/case stuff here and in `provider_data_api_client` ?

ivanELEC: I think you're right generally tbh, if this were a case where we are completely overhauling to a new endpoint that has new logic for everything and we need to build out v1 while running v0 for users in parallel, name-spacing would be appropriate. This case feel more like us getting a single broken api endpoint and updating it due to a fix

elken: If the changes don't work, we can revert the commit. I'd rather that than introducing extra branches that will rot, and yeah generally I think versioning the namespaces is a good idea, but it'll only really save hassle if we want to say use v1 of one endpoint and v2 of another.

I'd say just use the most appropriate endpoint for the method for now, and if we do hit a problem then we can add in versioned namespaces.

robinsk1: the structure of the client endpoints _shouldnt_ be too volatile in terms of their structure i'd guess - id expect response payloads to evolve over time more though

elken: That's a point, another reason to not yet though is I don't think there are any other v1 endpoints; but I don't know the full structure of the API so will defer the decision to Ivan.

elken (commented):
Fine with the feature flag setup for now, but will need a ticket to remove the code in future, unless you want to refactor this to use a namespaced API client?

If not then I'll approve this

app/services/office_code_service.rb
ivanELEC: So: 
I have not yet reviewed the other v0 to v1 changes so I don't know how big an overhaul it is. There is basically a new version of each endpoint + some new ones, but I don't know whether they've actually changed the backend/structure for all of them or not. It's probably worth us going over the new endpoints to see if there's anything there that can be used to enhance our user experience. If we do find things that lead to us deciding "ok, lets port everything to this new endpoint" I think namespacing is a good idea, removing tech debt via deleting folder that has v1 stuff in it, to me, is safer/easier than reverting a commit (what if we ended up adding more stuff to the v0 version over time?)

elken: Ha, okay yeah ignore my latest comment then and go for the simple approach and we can look at the wider API in the new year

ivanELEC:
> Fine with the feature flag setup for now, but will need a ticket to remove the code in future

will write this up now - I think it can be part of a spike looking at new api version

elken (approved):

PR #1394, opened 2025-01-16
[CRM457-2376] - NSM Provider Update Overview Page to Show Supporting Documents Have Been Deleted

imtiazAhmd:
## Notes for reviewer
This PR covers:
[2376](https://dsdmoj.atlassian.net/browse/CRM457-2376)

Screenshot 2025-01-22 at 08 45 17


app/presenters/app_store/v1/nsm/claim.rb
```diff
@@ -194,6 +195,10 @@ def local_record
           @local_record ||= ::Claim.find_or_create_by!(id:)
         end
 
+        def gdpr_documents_deleted?
+          respond_to?(:gdpr_documents_deleted) && gdpr_documents_deleted
```
ivanELEC, line 199: this logic will potentially change depending on [this conversation](https://mojdt.slack.com/archives/C05212WRK5J/p1737561667235629)

app/models/claim.rb
```diff
@@ -74,6 +74,10 @@ def additional_defendants
     defendants.where(main: false)
   end
 
+  def gdpr_documents_deleted?
+    false
+  end
```
ivanELEC, line 79: should there be a comment here to explain that this can default as false on the model because it would only ever not be false once the claim is in the app store (therefore we wouldn't be using this logic, but using the logic in the app store claim presenter)

app/presenters/app_store/v1/nsm/claim.rb
imtiazAhmd: resolved in the PR

app/models/claim.rb
imtiazAhmd: comment added 👍🏼

ivanELEC (approved):

PR #1398, opened 2025-01-21
CRM457-2374: Add service to handle draft GDPR claims

elken:
## Description of change

After 6 months of no updates, we have to remove all the supporting
evidence for a claim in order to comply with GDPR.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2374)


app/services/delete_document_service.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      evidence = claim.supporting_evidence.each do |file|
+        file_uploader.destroy(file.file_path) if File.exist?(file.file_path)
+        file.destroy
+      end
+
+      claim.supporting_evidence.delete(*evidence)
+
+      claim.gdpr_documents_deleted = claim.supporting_evidence.blank?
+      claim.save
```
elken, line 22: Not confident this should be here, if anyone else has similar objections I'll probably make this non-destructive

app/services/delete_document_service.rb
```diff
@@ -0,0 +1,31 @@
+# A service to delete all documents for a draft application, in order
+# to comply with GDPR.
+#
+# This only operates on drafts since the app store regulates the other
+# submissions. For each evidence file in the claim, we destroy the
+# local file if it exists and then delete the object itself. We use
```
robinsk1, line 6: are you sure we ever touch local file storage in prod env? surely thats just in local dev context thats ever the case

app/services/delete_document_service.rb
```diff
@@ -0,0 +1,31 @@
...
+# `delete` here to save a round-trip from doing `.reload`.
+class DeleteDocumentService
+  class << self
+    def call(claim_id)
+      claim = Claim.find(claim_id)
+      return unless claim && claim.state == 'draft'
+
+      evidence = claim.supporting_evidence.each do |file|
+        file_uploader.destroy(file.file_path) if File.exist?(file.file_path)
```
robinsk1, line 15: this is fine in local dev context but will that ever work in prod - s3 `file_uploader` is destroying a file only if it exists first in local file storage? - maybe theres something i'm missing about how we're doing file uploads in submit ie we are storing locally and on S3

app/services/delete_document_service.rb
elken: Yeah just a typo

app/services/delete_document_service.rb
elken: Good spot yeah, have updated to handle this at the file_uploader level

db/schema.rb
```diff
@@ -243,6 +244,7 @@
     t.datetime "originally_submitted_at"
     t.jsonb "viewed_steps", default: [], array: true
     t.index ["core_search_fields"], name: "index_prior_authority_applications_on_core_search_fields", using: :gin
+    t.index ["created_at", "service_type"], name: "idx_pas_service_type_created_at", where: "((service_type IS NOT NULL) AND ((service_type)::text <> ''::text))"
```
robinsk1, line 247: this index looks like its just appeared, maybe an artefact from your local db?

app/services/delete_document_service.rb
```diff
@@ -0,0 +1,31 @@
...
+    def call(claim_id)
+      claim = Claim.find(claim_id)
+      return unless claim && claim.state == 'draft'
+
+      evidence = claim.supporting_evidence.each do |file|
+        file_uploader.destroy(file.file_path) if file_uploader.exists?(file.file_path)
+        file.destroy
+      end
+
+      claim.supporting_evidence.delete(*evidence)
```
robinsk1, line 19: not too sure what this is doing

elken: The comment above explains it, removing all the evidence from the array using splat without having to reload from the db

app/services/delete_document_service.rb
```diff
@@ -0,0 +1,31 @@
...
+#
+# This only operates on drafts since the app store regulates the other
+# submissions. For each evidence file in the claim, we destroy the
+# file if it exists and then delete the object itself. We use `delete`
+# here to save a round-trip from doing `.reload`.
+class DeleteDocumentService
+  class << self
+    def call(claim_id)
+      claim = Claim.find(claim_id)
+      return unless claim && claim.state == 'draft'
```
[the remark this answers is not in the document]
elken: Probably makes sense still so we don't delete evidence by mistake before then

app/services/delete_document_service.rb
robinsk1: hmm - i guess you'd also only want to update this if 14-16 had been a success also?

elken: If it errors it'll throw an exception so it won't reach the save, my only thinking was the fact it "silently" mutates the record but it's such a minor thing. It's easy to overthink these tiny choices sometimes lol

robinsk1 (approved):

PR #1405, opened 2025-01-30
Fixes template rendering logic

imtiazAhmd:


elken (approved):
Approved under the assumption you've proven this works either locally or dev

PR #1408, opened 2025-02-07
CRM457-2430: Provider: Disbursement Sorting on Type Breaks

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2430)

- Protect translated_disbursement_type method from returning nil
- Add additional testing to cover previously broken scenario

### DIFF
--- a/app/models/concerns/disbursement_details.rb
+++ b/app/models/concerns/disbursement_details.rb
@@ -6,9 +6,8 @@ module DisbursementDetails
   end
 
   def translated_disbursement_type
+    return '' if other_type.nil? && disbursement_type.nil?
     if disbursement_type == DisbursementTypes::OTHER.to_s
-      return '' if other_type.nil?
-
       # rubocop:disable Performance/InefficientHashSearch
       # (False positive - Rubocop sees `.values` and wrongly infers OtherDisbursementTypes is a hash)
       known_other = OtherDisbursementTypes.values.include?(OtherDisbursementTypes.new(other_type))
--- a/app/models/concerns/disbursement_details.rb
+++ b/app/models/concerns/disbursement_details.rb
@@ -7,6 +7,8 @@ module DisbursementDetails
 
   def translated_disbursement_type
     if disbursement_type == DisbursementTypes::OTHER.to_s
+      return '' if other_type.nil?
+
       # rubocop:disable Performance/InefficientHashSearch
       # (False positive - Rubocop sees `.values` and wrongly infers OtherDisbursementTypes is a hash)
       known_other = OtherDisbursementTypes.values.include?(OtherDisbursementTypes.new(other_type))
--- a/app/models/concerns/disbursement_details.rb
+++ b/app/models/concerns/disbursement_details.rb
@@ -7,6 +7,7 @@ module DisbursementDetails
 
   def translated_disbursement_type
     return '' if other_type.nil? && disbursement_type.nil?
+
     if disbursement_type == DisbursementTypes::OTHER.to_s
       # rubocop:disable Performance/InefficientHashSearch
       # (False positive - Rubocop sees `.values` and wrongly infers OtherDisbursementTypes is a hash)
--- a/app/models/concerns/disbursement_details.rb
+++ b/app/models/concerns/disbursement_details.rb
@@ -6,8 +6,6 @@ module DisbursementDetails
   end
 
   def translated_disbursement_type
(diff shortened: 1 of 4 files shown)


app/models/concerns/disbursement_details.rb
```diff
@@ -6,6 +6,8 @@ module DisbursementDetails
   end
 
   def translated_disbursement_type
+    return '' if other_type.nil? && disbursement_type.nil?
```
elken, line 9: This should be an or right? If other_type is nil but disbursement_type isn't it'll still hit the error

ivanELEC: this is where it gets tricky (I think) if disbursement_type is populated from a disbursement in the list, other_type is going to be nil so that would lead is to return an empty string when the disbursement has been selected

ivanELEC: yeah other_type would be nil in that scenario so can't just use OR, may have to add some more conditions

elken: Oh you put it in the wrong place anyway, it needs to go inside the `if` condition as it only applies if it's other

ivanELEC: I think in general we are overcomplicating it, I'm just going to add else blank to the whole if else statement in this method

ivanELEC: when I say blank I mean ''

elken: It's not overcomplicating it, `OtherType.values` can't parse nil as a valid value so it has to early return to prevent an exception.

robinsk1: bumping up a unit test might be your friend here

ivanELEC: > It's not overcomplicating it, `OtherType.values` can't parse nil as a valid value so it has to early return to prevent an exception.

I think the misunderstanding here is the condition in which this error happens, if you select Other for the disbursement type but don't populate a specific disbursement at all like so, the error doesn't happen because disbursement_type in this instance is stored as a "other" and other_type is stored as blank string as we would want: 
image

The bug only occurs when you don't click on that radio button for other disbursement at all - in this instance disbursement_type is nil as nothing has been selected for it so it will never go down that branch: 
image

Agreed though - should unit test

elken: Yes and if `other_type` is nil, you get the below

image

It's a very simple change that as Robin points out is evidently not correctly handled in tests

ivanELEC: The reason I wanted to simplify is because this means unnecessarily adding the same line of code to 2 branches 
From going through the code/testing via trying the different combinations, I found that other_type can only be nil if disbursement_type is nil - so from what I know, the outcome you've mentioned here is only possible via changing db data (and not via user input through the form) unless I'm missing a scenario which I could be. 

I originally put the code where you suggested but found it didn't solve the problem I thought it did because in all the real life scenarios it can never reach that branch. That said, adding the code there also ensures we don't run into this problem if we change anything about the form (i.e. add a scenario in the form where other_type can be nil and not "" when disbursement_type is other) - so I guess it makes sense to add

elken: You don't have to add it to both at all, the only reason other_type is nil is if other is selected, otherwise we never have to consider it. 

The `if type == other` branch is the only place it needs to check

ivanELEC: sorry I said something innacurate in that message, it's not the same logic in two branches, just similar logic in 2 branches - but yeah your point about the XML convinces me even more that it's not a scenario that we should ignore just because there's no obvious way a user can trigger it

ivanELEC: *obvious current way

robinsk1: LGTM - does that change have any impact on the schema? probably doesn't but just floating that in case

elken: > LGTM - does that change have any impact on the schema? probably doesn't but just floating that in case

XML? No

JSON? We can probably handle this in jsonschema if it supports dependant data which XSD doesn't quite cover.

ivanELEC: Updated!

ivanELEC:
> The else check is redundant here, but it's no real harm

it's not redundant, it is a possible scenario (as described in the messages earlier)

PR #1411, opened 2025-02-11
CRM457-2388: NSM Provider: Bulk upload - incomplete work items/ disbursements

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2388)

## Changes

- Updated UI elements for incomplete work items/disbursements in the summaries
- Updated UI elements for work items/disbursements in the summaries when the claim is imported
- Update UI elements for the flash message on successful imports
- Added extra fields (import_date) for metrics
- Made presenter methods generic so that this can be applied to many list items in a claim

## Before Changes
image

## After Changes
image
image

## Notes for reviewer

- The import_date field is there purely for reporting reasons now that we are showing the imported work items/disbursement information in the banner after the import form 
- E2E test branch will fail as I need to update the E2E tests


app/controllers/concerns/incomplete_items_concern.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  private
+
+  def incomplete_items_summary
+    @incomplete_items_summary ||= Nsm::IncompleteItems.new(current_application, item_type,
+                                                           self)
+  end
+
+  def build_items_incomplete_flash
+    flash = if import_notification_needed?
```
elken, line 12: No need for this to be a variable

app/presenters/nsm/incomplete_items.rb
```diff
@@ -0,0 +1,78 @@
...
+          path_title(item),
+          path_url(item)
+        )
+      end.join(', ')
+    end
+
+    private
+
+    def items
+      if @type == :work_items
```
elken, line 45:
```ruby
@items ||= case @type
           when :work_items then @claim.work_items 
           when :disbursements then @claim.disbursements
           end
```

No reason you can't just set this in the constructor though

app/presenters/nsm/incomplete_items.rb
```diff
@@ -0,0 +1,78 @@
...
+      end
+      # :nocov:
+    end
+
+    def path_title(item)
+      t("#{path_key}.item", index: item.position)
+    end
+
+    def path_url(item)
+      if item.is_a? WorkItem
```
elken, line 62: https://api.rubyonrails.org/classes/ActionDispatch/Routing/PolymorphicRoutes.html

config/locales/en/nsm/steps.yml
```diff
@@ -152,6 +152,9 @@ en:
             duplicate: Duplicate
             delete: Delete
           unknown_date: Unknown date
+          items_imported:
+            one: You imported %{count} work item
```
elken, line 156: Does this also show correctly for 0?

app/controllers/concerns/incomplete_items_concern.rb
```diff
@@ -0,0 +1,37 @@
+module IncompleteItemsConcern
+  extend ActiveSupport::Concern
+
+  private
+
+  def incomplete_items_summary
+    @incomplete_items_summary ||= Nsm::IncompleteItems.new(current_application, item_type,
```
robinsk1, line 7: formatting

app/controllers/nsm/steps/work_items_controller.rb
```diff
@@ -28,6 +30,10 @@ def set_default_table_sort_options
         @sort_by = params.fetch(:sort_by, default)
         @sort_direction = params.fetch(:sort_direction, 'ascending')
       end
+
+      def item_type
```
robinsk1, line 34: you could maybe infer this from controller name

app/services/submit_to_app_store/nsm_payload_builder.rb
```diff
@@ -68,7 +68,7 @@ def direct_attributes
                                allowed_calls_uplift allowed_letters allowed_letters_uplift allowed_youth_court_fee
                                youth_court_fee_adjustment_comment calls_adjustment_comment letters_adjustment_comment
                                state core_search_fields submit_to_app_store_completed send_notification_email_completed
-                               originally_submitted_at])
+                               originally_submitted_at imported_work_items_viewed imported_disbursements_viewed import_date])
```
robinsk1, line 71: feels odd that that we have to do this and and that we're recording that a claim is imported as an attribute - shouldn't the claim itself be indifferent to how it got into the system? the user still has to pass through the form steps/validation which sort of negates the fact its imported as updates via that process could alter the claim  - can we not use cookie/session/flash data to track whatever we're trying to odo here? the claim being 'imported' seems very temporal and stops being imported the minute it is altered,saved, submitted? definitely feels like session storage

app/controllers/nsm/steps/work_items_controller.rb
elken: We discussed this on Wednesday, it's better to be explicit on what types are available. Could probably just be a constrcutor argument though

app/services/submit_to_app_store/nsm_payload_builder.rb
elken: We have to track the fact it's been imported to decide if we show the "you've import x work items" message, but I agree we don't need to worry about the other 2 fields.

app/presenters/nsm/incomplete_items.rb
ivanELEC: I was using case like this for it in a previous iteration but it didn't remedy the coverage issues, but yeah thinking about it probably better format wise

I thought since if we have the type in the constructor, then we infer the items from the type, it would reduce the risk of us having code where someone mismatches the type to the items they feed in. Or do you mean creating the instance variable within the initializer instead of a private method?

app/services/submit_to_app_store/nsm_payload_builder.rb
ivanELEC: This line of code is explicitly taking tese fields out of the payload, not adding them in (the payload adds all the top level fields EXCEPT these ones

ivanELEC: That said, I think it's worth talking about whether we should include import_date to the payload for metrics reasons so we could track how many people are using the feature

ivanELEC: I wanted to open that up as a discussion once we started tying things together

app/presenters/nsm/incomplete_items.rb
elken: Yeah just creating it in the constructor. You can have the fallback case just throw the error and then you're covered

config/locales/en/nsm/steps.yml
ivanELEC: Shows like this: 
image
which is to spec, but not sure if it's something we want to show Mary to ask what to show or should it show at all of none have been imported

app/controllers/concerns/incomplete_items_concern.rb
ivanELEC: gonna need more details on this, this formatting is a result of me running rubocop (I like to make sure things are indented better too but rubocop automatically moved it to this)

app/services/submit_to_app_store/nsm_payload_builder.rb
robinsk1: do we need to save `imported` as a flag though in the claim model? we cant use the session object for that?

app/controllers/concerns/incomplete_items_concern.rb
ivanELEC: it needs to be a variable because: 

- the function needs to return the flash hash
- the function needs to set the "viewed" flag to true only after the hash has been generated fgrom the if statements, since import_notification_needed? relies on that flag to check if the page has been viewed or not 

Therefore I need to pre-store the flash to ensure it follows the correct path, reset the value, then return it

app/controllers/nsm/steps/work_items_controller.rb
ivanELEC: see: https://github.com/ministryofjustice/laa-submit-crime-forms/pull/1411/files/a044ee06067d4b05d068ccb523a3501a0f69999a#r1959641019

app/presenters/nsm/incomplete_items.rb
ivanELEC: forgot to report back but not 100% sure polymorphic paths are possible with the namespacing situation we have - or at least it's not too clear from the documentation, playing/reading around now though

app/controllers/concerns/incomplete_items_concern.rb
ivanELEC: I think this might be better reviewed on a call

app/controllers/concerns/incomplete_items_concern.rb
elken: It's just ugly having self on its own like that I think is Robin's point, which it kinda is

ivanELEC: ah ok, fixing that

app/presenters/nsm/incomplete_items.rb
```diff
@@ -15,16 +15,14 @@ class IncompleteItems
...
-
       @claim = claim
       @type = type
       @controller = controller
       @items ||= case @type
                  when :work_items then @claim.work_items
                  when :disbursements then @claim.disbursements
                  else
-                   raise 'Cannot create items from this type'
+                   raise "Cannot create items from type: #{type}"
```
elken, line 25: Should wrap type in quotes in case it's null or empty space somehow

elken:
Once this is green I'm happy enough to approve this

robinsk1 (commented):
should we squash these commits down maybe? i'm not super-fussed on that either btw if its a major headache

ivanELEC:
> Once this is green I'm happy enough to approve this

I've got rspec passing and full test coverage. the e2e tests need updating but easier to merge and update the e2e tests after

ivanELEC:
> should we squash these commits down maybe? i'm not super-fussed on that either btw if its a major headache

they'll squash automatically when it's merged

app/services/nsm/importers/xml/import_service.rb
```diff
@@ -34,11 +34,16 @@ def import_xml
...
+          {
+            title: I18n.t('nsm.imports.message_title'),
+            content: message_content
+          }
+        end
+
+        def message_content
+          "#{I18n.t('nsm.imports.work_items_message', count: claim.work_items.count)}
+            #{I18n.t('nsm.imports.disbursements_message', count: claim.disbursements.count)}
+            </p>#{I18n.t('nsm.imports.message_end')}"
```
elken, line 46: Why is this hanging closing tag here?

spec/controllers/nsm/steps/disbursements_summary_controller_spec.rb
```diff
@@ -1,6 +1,27 @@
...
+        build(:disbursement, :valid, disbursement_date: nil),
+        build(:disbursement, :valid, disbursement_date: nil)
+      ]
+    end
+
+    before do
+      allow(controller).to receive(:current_application).and_return(claim)
+    end
+
+    it 'generates the correct error summary' do
```
elken, line 18: Should also test for the item link going to the correct page

app/services/nsm/importers/xml/import_service.rb
ivanELEC: having a single closing </p> is equivalent to <p></p>, it adds an empty newline basically

ivanELEC: oh god forgot markdown would format that

ivanELEC: but also kind of proves my point haha

ivanELEC: having a single closing "/<p>" is equivalent to "<p></p>", it adds an empty newline basically

ivanELEC: image

elken: That's a bug not a feature and produces invalid HTML. Use `<br>` or just a regular `\n`

spec/controllers/nsm/steps/disbursements_summary_controller_spec.rb
```diff
@@ -20,6 +20,17 @@
       expect(controller.instance_variable_get(:@items_incomplete_flash)[:default])
         .to include('2 disbursements are incomplete:')
     end
+
+    it 'creates the correct link for incomplete disbursement' do
+      incomplete_disbursements = claim.disbursements.reject(&:complete?).sort_by(&:position)
+      get :edit, params: { id: claim.id }
+      expect(controller.instance_variable_get(:@items_incomplete_flash)[:default])
+        .to include("<a class=\"govuk-link\" href=\"/non-standard-magistrates/applications/#{claim.id}" \
```
elken, line 28: You could make it a bit more readable by just checking for `application/#{claim.id}`

db/schema.rb
```diff
@@ -111,14 +111,15 @@
...
+    t.boolean "include_youth_court_fee"
     t.date "change_solicitor_date"
     t.string "case_outcome_other_info"
-    t.boolean "include_youth_court_fee"
     t.string "main_offence_type"
     t.boolean "allowed_youth_court_fee"
     t.string "youth_court_fee_adjustment_comment"
     t.jsonb "viewed_steps", default: [], array: true
     t.boolean "gdpr_documents_deleted"
+    t.datetime "import_date"
```
robinsk1, line 122: do you have to update schema to include this in payload?

spec/controllers/nsm/steps/disbursements_summary_controller_spec.rb
ivanELEC: I wanted to ensure the index of the items consistently paired with the right item id in this test

db/schema.rb
ivanELEC: yeah, thanks for reminding me

ivanELEC: https://github.com/ministryofjustice/laa-crime-forms-common/pull/41

elken (approved):

robinsk1 (changes requested):
we need to  include common form gem field change here, also need to update common form gem version number and bump gem version - ditto app-store and assess

elken (changes requested):
Blocked by QA changes

elken (approved):
QA resolved

robinsk1 (approved):

PR #1415, opened 2025-02-18
[CRM457-2373] bulk upload import a claim

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2373)

import a CRM7 claim from xml


config/import_nsm_claim.xsd
```diff
@@ -14,13 +14,6 @@
     </xs:restriction>
   </xs:simpleType>
 
-  <xs:simpleType name="stageReachedType">
```
elken, line 17: Why?

elken (commented):
This looks good, only other thing I would say is ensuring that the schema is versioned also and that we handle differently versioned xsd files

config/import_nsm_claim.xsd
elken: Ah yes that's true, good catch

app/services/nsm/importers/xml/v1/importer.rb
```diff
@@ -0,0 +1,74 @@
...
+
+          def call
+            create_work_items
+            create_defendants
+            create_disbursements
+            create_firm_office
+            create_solicitor
+            resolve_reasons_for_claim
+
+            claim.update(hash)
```
ivanELEC, line 21: can this/should this be in a transaction?

ivanELEC: was thinking of the import_date stuff and what I'd do if anythign failed, but applies to any of the data updates really

app/services/nsm/importers/xml/v1/mapper.rb
```diff
@@ -0,0 +1,21 @@
+module Nsm
+  module Importers
+    module Xml
+      module V1
+        class Mapper
```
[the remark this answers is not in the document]
ivanELEC: agreed

app/services/nsm/importers/xml/v1/importer.rb
elken: claim.update should be in a transaction yeah as that's doing the actual saving

robinsk1: `claim.update(hash)` is one atomic db action, putting that in a transaction will have no effect

elken: Oh given that then no probably not

ivanELEC: Ah I see that makes sense, not the best user experience but I acknowledge that it's what we agreed to start with

ivanELEC (approved):

PR #1422, opened 2025-02-26
Remove position from XML schema

elken:
## Description of change

Resolves a bug around adding new work items/disbursements with a position to an imported claim that had work items/disbursements in.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2454)


spec/services/nsm/importers/xml/v1/importer_spec.rb
```diff
@@ -71,13 +71,12 @@
...
   describe '#create_work_items' do
     it 'creates work items' do
-      expect(claim.work_items.find_by(position: 1)).to have_attributes(
+      expect(claim.work_items.first).to have_attributes(
         work_type: 'preparation',
         time_spent: 120,
         completed_on: Date.new(2024, 1, 15),
         fee_earner: 'JS',
         uplift: 20,
-        position: 1,
```
robinsk1, line 80: is position computed correctly here/what we do we expect position to be here? we would expect position to be 1 here i guess for the first work_item - if its nil it not being computed

elken: It's nil always except when we show it on the frontend

imtiazAhmd (approved):

spec/services/nsm/importers/xml/v1/importer_spec.rb
robinsk1: ok, yeah it being in the schema is odd - might be something to do with how we build test factories

PR #1430, opened 2025-03-07
Skip tests impacted by recent Chrome update

robinsk1:
## Description of change

A recent Chrome update has caused our system tests to fail for some unknown reason. We have tried looking at pinning to a specific version locally but nothing seems to resolve it so we're for now skipping these tests to review at a later date.

### DIFF
--- a/spec/system/nsm/disbursement_type/suggestion_autocomplete_spec.rb
+++ b/spec/system/nsm/disbursement_type/suggestion_autocomplete_spec.rb
@@ -1,6 +1,7 @@
 require 'system_helper'
 
-RSpec.describe 'Test suggestion autocomplete for court', :javascript, type: :system do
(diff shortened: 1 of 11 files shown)


spec/system/nsm/disbursement_type/suggestion_autocomplete_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'system_helper'
 
-RSpec.describe 'Test suggestion autocomplete for court', :javascript, type: :system do
+RSpec.describe 'Test suggestion autocomplete for court', :javascript, :skip, type: :system do
```
elken, line 3: (applies to all of them)

We don't want to skip all the tests, just the cases that directly interact with a JS component and thus fails

elken (approved):

PR #1433, opened 2025-03-10
Introduce a new XML errors summary

elken:
## Description of change

To provide a nicer user experience, the ugly errors we get back from
XML validation are shadowed behind a PDF that can be downloaded and
sent to CMS providers to fix and a general error message returned to
the user instead.

### Screen

image

### Error summary
image

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2445)


app/controllers/nsm/imports_controller.rb
```diff
@@ -4,27 +4,76 @@ class ImportsController < ApplicationController
...
-            render :new
+            session['xml_errors'] = CompressionService.compress(@validation_errors)
+
+            @form_object.errors.add(:file_upload, :validation_errors)
           end
         end
-      else
-        render :new
       end
+      render :new
```
robinsk1, line 27: is it possible to put that back in under an `else` - reads more declaratively than having it fall through?

app/controllers/nsm/imports_controller.rb
```diff
@@ -4,27 +4,76 @@ class ImportsController < ApplicationController
...
+      schema = Nokogiri::XML::Schema.new(schema_file)
+
+      schema.validate(xml_file)
+    end
+
+    def build_message(claim)
+      hash = Hash.from_xml(xml_file.to_s)['claim']
+
+      # TODO: CRM457-2473: Refactor this to handle versioning better
+      Nsm::Importers::Xml.const_get("v#{xml_file.version.to_i}".capitalize)::Importer.new(claim, hash).call
```
robinsk1, line 66: this method looks like its doing the importing and building the flash message, seems like two different things, maybe a naming thing or seperation of concerns

app/services/pdf_service.rb
```diff
@@ -8,12 +8,12 @@ class PdfService
       right: '1cm',
     },
     viewport: {
-      width: 2400,
```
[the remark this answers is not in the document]
elken: Leftover from me trying to work out why Grover kept timing out, good spot

app/controllers/nsm/imports_controller.rb
```diff
@@ -4,27 +4,76 @@ class ImportsController < ApplicationController
...
       if @form_object.valid?
         initialize_application do |claim|
-          success_message = Nsm::Importers::Xml::ImportService.call(claim, @form_object)
-          if success_message
-            redirect_to edit_nsm_steps_claim_type_path(claim.id), flash: { success: success_message }
+          @validation_errors = validate
+
+          if @validation_errors.empty?
+            redirect_to edit_nsm_steps_claim_type_path(claim.id), flash: { success: build_message(claim) }
+            return
```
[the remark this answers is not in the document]
elken: I did while I was testing yeah, otherwise it falls back to the `:new` view

app/controllers/nsm/imports_controller.rb
elken: Reads better like that and saves a redundant branch 😛

app/controllers/nsm/imports_controller.rb
elken: It is, I had to move it out of the main method as it was hitting the AbcSize but I've since reduced it so it can probably go back there now

app/services/pdf_service.rb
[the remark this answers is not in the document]
elken: I will also spend a little time seeing if I can make Grover time out a bit less, I have a couple of thoughts

app/controllers/nsm/imports_controller.rb
robinsk1: i think thats maybe why you need the explicit `return` on `:19`- its not really redundant as its explicitly informing that the action renders new if the `form_object` isn't valid, which is pretty clear - that to me reads better and aligns with the way we do other controller actions

app/services/pdf_service.rb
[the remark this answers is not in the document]
elken: Added a cache parameter that seems to help a bit, but still think we need to look at an alternative in the future. I'll add a dev meeting note.

app/controllers/nsm/imports_controller.rb
```diff
@@ -4,27 +4,77 @@ class ImportsController < ApplicationController
 
     def new
       @form_object = ImportForm.new
+      @validation_errors = []
+      session['xml_errors'] = nil
```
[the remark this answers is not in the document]
elken: It could and that's why we zlib compress them before we send them down. Setting it to nil like that just ensures that we don't send an old error report by mistake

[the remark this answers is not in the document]
elken: Pushed a change to use a "temp" file instead now (can't use actual `Tempfile` but it's close enough) and that seems to handle anything I can generate

app/views/shared/_flash_banner.html.erb
```diff
@@ -1,5 +1,6 @@
 <% flash.each do |type, msg| %>
   <% next unless msg.is_a?(String) || msg.is_a?(Hash) %>
+  <% msg = msg.is_a?(Hash) ? msg.transform_keys(&:to_sym) : msg %>
```
robinsk1, line 3: can `sanitize` take care of this in order to take this logic out of the view?

app/controllers/nsm/imports_controller.rb
```diff
@@ -4,27 +4,87 @@ class ImportsController < ApplicationController
...
       @form_object = ImportForm.new
+      @validation_errors = []
+
+      # Ensure we don't keep a leftover file
+      # Can't use Tempfile here as it expires too quickly
+      begin
+        errors_file_path.unlink
+      rescue StandardError
+        nil
+      end
```
robinsk1, line 15: dont need this `end`

app/controllers/nsm/imports_controller.rb
```diff
@@ -4,27 +4,87 @@ class ImportsController < ApplicationController
...
+
+            # TODO: CRM457-2473: Refactor this to handle versioning better
+            Nsm::Importers::Xml.const_get("v#{xml_file.version.to_i}".capitalize)::Importer.new(claim, hash).call
+
+            redirect_to edit_nsm_steps_claim_type_path(claim.id), flash: { success: build_message(claim) }
+            return
           else
-            render :new
+            errors_file_path.write(@validation_errors.to_json)
+            @form_object.errors.add(:file_upload, :validation_errors)
```
robinsk1, line 35: if you stick `ender :new` under here it sticks to conventional rails pattern and you can get rid of the dangling return on `:32`

robinsk1 (approved):

robinsk1 (dismissed):
squash before merge

app/controllers/nsm/imports_controller.rb
elken: Rubocop fixed it to that

app/views/shared/_flash_banner.html.erb
elken: That's just for HTML as far as I'm aware.

app/controllers/nsm/imports_controller.rb
elken: https://api.rubyonrails.org/classes/ActionController/Redirecting.html convention is to return after redirecting, I've adjusted it to `and return` instead

elken:
Tried to dismiss the request changes but it still replaced the approval

robinsk1 (approved):

PR #1446, opened 2025-03-17
Document XML schema

elken:
## Description of change
Add documentation to the XML schema and make sensible assumptions
about some of the redundant fields (e.g. fields rely on a yes/no
answer we can autofill based on the presence)

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2485)


imtiazAhmd (approved):
LGTM

PR #1458, opened 2025-03-25
CRM457-2473 Handle XML schema versioning

imtiazAhmd:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2473)


app/controllers/nsm/imports_controller.rb
```diff
@@ -71,10 +102,29 @@ def xml_file
       @xml_file ||= Nokogiri::XML::Document.parse(@form_object.file_upload.tempfile.read, &:noblanks)
     end
 
+    # Extract version from XML
+    def extract_version_from_xml
+      version_node = xml_file.at_xpath('//version')
+      return version_node.text.to_i if version_node&.text.present?
+
+      1 # Default to version 1 if not found
```
ivanELEC, line 110: not feeling comfortable with the fact that we are defaulting to version 1 if not found.  We should always be getting the version with the xml file, there is even logic in the validate method that adds an error if the version node is empty - meaning it shouldn't be possible to get to this point if this is the case (therefore, why add code for it?) - this could cause us to slip up when making future updates - if this is for testing purposes, I feel like there's a better way

app/controllers/nsm/imports_controller.rb
```diff
@@ -63,6 +58,42 @@ def self.model_class
...
+      hash = prepare_claim_hash
+      version = extract_version_from_xml
+
+      import_claim(claim, hash, version)
+
+      redirect_to edit_nsm_steps_claim_type_path(claim.id), flash: { success: build_message(claim) }
+    end
+
+    def prepare_claim_hash
+      hash = Hash.from_xml(xml_file.to_s)['claim']
```
ivanELEC, line 83: Probably cleaner to do this in 1 line by using the [except](https://apidock.com/rails/Hash/except) method: 
`Hash.from_xml(xml_file.to_s)['claim'].except(:version)`

app/controllers/nsm/imports_controller.rb
imtiazAhmd: Forgot to delete this, already added all the case scenarios check like , missing, invalid alongside valid and unsupported.

app/controllers/nsm/imports_controller.rb
```diff
@@ -23,22 +23,20 @@ def create
...
           else
-            errors_file_path.write(@validation_errors.to_json)
-            @form_object.errors.add(:file_upload, :validation_errors)
+            handle_validation_errors
           end
         end
       end
 
       render :new
+    rescue ActionController::ParameterMissing
```
elken, line 36: This isn't needed, it's already handled as part of the new `before_action`

app/controllers/nsm/imports_controller.rb
```diff
@@ -62,6 +60,38 @@ def self.model_class
...
+
+      import_claim(claim, hash, version)
+
+      redirect_to edit_nsm_steps_claim_type_path(claim.id), flash: { success: build_message(claim) }
+    end
+
+    # Removed prepare_claim_hash method as it's no longer needed
+
+    def import_claim(claim, hash, version)
+      Nsm::Importers::Xml.const_get("V#{version}".capitalize)::Importer.new(claim, hash).call
```
elken, line 87: No handling of "version not found" errors

app/controllers/nsm/imports_controller.rb
```diff
@@ -78,6 +108,17 @@ def xml_file
       @xml_file ||= Nokogiri::XML::Document.parse(@form_object.file_upload.tempfile.read, &:noblanks)
     end
 
+    # Extract version from XML
```
elken, line 111: `version` is already part of the XML file, included in the `<?xml` doctype tag and available in the `.version` method. The point of this ticket as I said was to just enforce that that's included and if it's not we throw an error before we try to validate (since we can't find the right version to validate without the right version) rather than introducing a new field.

This comment applies to all the added-field-related work

elken (changes requested):

app/controllers/nsm/imports_controller.rb
imtiazAhmd: Thanks, It was before I merged the main. updating now.

app/controllers/nsm/imports_controller.rb
imtiazAhmd: The XML declaration (<?xml version="1.0"?>) indicates the XML standard version, The XML Declaration provides the XML Parser with version of the XML specification to use encoding. UTF-8, UTF-16, ISO-8859-1 etc. This can't/ shouldn't be used for our versioning.

app/controllers/nsm/imports_controller.rb
imtiazAhmd: its already handled, `extract_version_from_xml` and in `validate` method. 
- Missing version element
- Invalid version value (0 or negative)
- Unsupported version (no schema file exists)

app/controllers/nsm/imports_controller.rb
elken: I see, makes sense then. Annoying that we have to add a field for this as it'll constitute _another_ schema push-out.

In that case, the comment against it should be expanded a bit to clarify the purpose of the field, maybe

> The version of the schema to compare an export against.

and an additional `xs:documentation` to clarify that we expect an integer, not to pass a SemVer number or anything (minOccurs/maxOccurs also isn't needed for an integer, if they try and submit a self-closing or empty tag, it throws an error as you can't have "nothing" for a number)

app/controllers/nsm/imports_controller.rb
elken: If you're confident there's no way that you can hit any bad paths here then fine

app/controllers/nsm/imports_controller.rb
```diff
@@ -23,22 +23,18 @@ def create
...
           else
-            errors_file_path.write(@validation_errors.to_json)
-            @form_object.errors.add(:file_upload, :validation_errors)
+            handle_validation_errors
           end
         end
       end
 
       render :new
+    rescue StandardError => e
```
elken, line 36: We already have a standard error page for this, there's no point calling something explicit here otherwise we'd have to for every action

elken (changes requested):

app/controllers/nsm/imports_controller.rb
imtiazAhmd: Agreed, deleted now.

elken (approved):
Looks fine now, there's a couple of comments you can probably remove as they don't add anything but otherwise fine

PR #1468, opened 2025-03-28
Show all service types when the box is focused even when empty

elken:
## Description of change

Allow the service type combo box in PA function more like a drop-down
box and hopefully reduce the amount of free-text entries.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2478)


ivanELEC:
I know this isn't part of the original ticket spec, but if we're going to do this, should the entries be alphabetic as well so that it's easy for users for find what they want?

ivanELEC (approved):

PR #1470, opened 2025-03-28
Delete evidence when a draft claim/application is deleted

elken:
## Description of change

Prevent orphaned evidence files caused by deleting draft claims/applications.

Also re-syncs the SimpleCov numbers so that coverage "works" locally.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2471)


app/models/claim.rb
```diff
@@ -1,6 +1,8 @@
 class Claim < ApplicationRecord
   include ClaimDetails
 
+  before_destroy :destroy_attachments
+
```
ivanELEC, line 5: I think you made the right call here by adding this as a hook instead of, for example a concern added to the controllers where draft deletions happen. One thing we will need to remember/think about is if we were to go full app store, this functionality would need to be carried over accordingly - but that's almost a moot point as I can't see a reason to migrate drafts to app store in the future

ivanELEC (approved):

app/models/claim.rb
elken: It'd be _far_ more effort than its worth, if it was ever a consideration I'd push back on it. We need to get local non-draft submissions migrated first before we can even consider it and that itself is a massive body of work.

PR #1488, opened 2025-04-23
Clear out old supporting_documents

elken:
When a supporting document has its file path updated, schedule a job
to clear out the old document after a day.


elken:
Marking as draft for now until associated app store PR goes up

elken:
Moved to draft and do not merge for reasons on Slack

PR #1489, opened 2025-04-23
CRM457-2517:  Record unsuccessful XML import attempts

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2517)

## Notes for reviewer

- Create FailedImport class associated to Providers 
- Store import error details in a FailedImport object
- Use FailedImport object to generate error PDF files
- If error doesn't generate error details (taken from Nokigiri error output), do not attempt to show download link for error pdf


app/services/app_store_client.rb
```diff
@@ -39,6 +39,24 @@ def search(payload)
...
 
+  def post_import_error(message)
+    url = "#{host}/v1/failed_imports"
+    response = self.class.post(url, **options(message))
+
+    case response.code
+    when 201
+      JSON.parse(response.body)
+    else
+      raise "Unexpected response from AppStore - status #{response.code} for '#{message[:id]}'"
```
elken, line 50: Something to be aware of for the future (no change needed here) is that we often swallow and ignore these "unexpected app store" errors, we should come up with a better plan.

app/views/nsm/imports/errors.html.erb
```diff
@@ -7,8 +7,9 @@
     <p><%= t('.explanation') %></p>
 
     <%# Ugly formatting because any whitespace in the block is passed through to the frontend %>
+    <% formatted_errors = JSON.parse(AppStoreClient.new.get_import_error(params[:error_id])['details']) %>
```
elken, line 10: This call should be in the controller, the view should recieve the data it needs

app/services/app_store_client.rb
elken: I'll add a note to the dev meeting board and resolve this when I have

app/views/nsm/imports/errors.html.erb
ivanELEC: yes you're right I don't know why I left it like this - I'll use an instance variable from the controller

elken (approved):

PR #1536, opened 2025-06-06
Use local file for glibc key

ivanELEC:


elken (approved):
The key can probably just go in a `etc/` folder in the repo, doesn't need to be top-level

PR #1541, opened 2025-06-23
Map new EntraID users and log potentially missing users

elken:
Attempt to lookup users by email first and use that to update their
old Portal account, otherwise behave as before.

We also log out any potential manual matches to handle later.

## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2669)

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -19,13 +19,21 @@ class Provider < ApplicationRecord
 
   class << self
     def from_omniauth(auth, office_codes)
-      find_or_initialize_by(auth_provider: auth.provider, uid: auth.uid).tap do |record|
-        record.update(
+      user = find_by(email: auth.info.email) ||
+             find_by(auth_provider: auth.provider, uid: auth.uid) ||
+             new
+
+      user.tap do |record|
+        record.assign_attributes(
           email: auth.info.email,
           description: auth.info.description,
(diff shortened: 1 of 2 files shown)


app/models/provider.rb
```diff
@@ -19,13 +19,23 @@ def multiple_offices?
...
+      Rails.logger.info("Potential manual mapping needed for #{auth.info.email} (#{auth.uid})") if user.new_record?
+
+      user.tap do |record|
+        record.assign_attributes(
           email: auth.info.email,
           description: auth.info.description,
           roles: auth.info.roles,
           office_codes: office_codes,
+          auth_provider: auth.provider,
+          uid: auth.uid
```
ivanELEC, line 35: Not sure what the data looks like coming out of Entra ID e.g. roles and accounts/office codes - so can't assess whether or not this is correct (even though I know it probably is)

Could we document what the response looks like somewhere like confluence?

ivanELEC (commented):
a few comments

app/models/provider.rb
elken: As per a previous thread, the data comes from Omniauth and is identical to the SAML response with some extra fields.

ivanELEC (approved):

PR #1565, opened 2025-07-25
Improve login UX across the app

elken:
This improves the general login UX encompassing multiple scenarios.

If we can, just attempt to silently authenticate a user if they have a
valid Entra session. Otherwise we include several fallbacks.

## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2703)


robinsk1 (commented):
Looks fine given all SSO limitations and possible scenarios here 😄 - can we include the steps to reproduce the fallback scenarios either on ticket or on PR description just to help with testing etc?  might be good to have a walk-through/demos of those as well if you get time

PR #1568, opened 2025-07-30
Turn PDA back on

elken:
We can now go back to using this

## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2708)


ivanELEC:
Before I approve, could we get Ewan or someone to write up a [decision record](https://dsdmoj.atlassian.net/wiki/spaces/CRM457/pages/4599152958/Decision+Log+to+be+updated+by+Mati) on the area of law criteria? I just want to make sure we're covered in the decision

app/services/active_office_code_service.rb
```diff
@@ -1,15 +1,25 @@
...
       office_codes.select { active?(_1) }
     end
 
     def active?(office_code)
-      always_inactive_office_codes.exclude?(office_code)
+      return true if always_active_office_codes.include?(office_code)
+      return false if always_inactive_office_codes.include?(office_code)
+      return contract_active?(office_code) if FeatureFlags.provider_api_login_check.enabled?
+
+      true
```
robinsk1, line 14: whats the scenario here that we need the fallback if it escapes the scope of the guard clauses?

robinsk1 (approved):

PR #1775, opened 2026-04-09
20260409 - Package Updates

ivanELEC:
- Update setup-ruby version in GHA pipelines
- Update rack-session from 2.1.1 to 2.1.2
- Updates from dependabot/bundler/bundler-df5b5f635b


RobBoothMOJ (approved):
LGTM

